PR #192, opened 2025-05-17
[Issue 191] Transformer Scaffolding

### LINKED ISSUE #191: [Py SDK] Transformer scaffolding
### Summary

Implement scaffolding for transforming arbitrary data into canonical Opportunity model instances.

### Acceptance criteria

- [ ] Python developers can leverage a transformer do something like the following:

```python
from common_grants_sdk.schemas.models import OpportunityTransformer

transformer = OpportunityTransformer(source_data={ ... })

opportunity = transformer.transform_opportunity()
```

- [ ] The transformer is extensible to enable various forms of mappings

DavidDudas-Intuitial:
### Summary

- Fixes #191 
- Time to review: 10 minutes

Creates scaffolding for custom data transformers. 

Grant platform data may come from many heterogeneous sources (e.g. XML APIs, proprietary CSV formats, legacy JSON, etc). This PR introduces an abstract base class designed for runtime transformations from arbitrary data structures to canonical data models as defined by the CommonGrants protocol. This class defines a standard interface for transforming raw input data into structured `OpportunityBase` instances. Developers may subclass the abstract base class to implement custom transformation logic for normalizing various source formats.

Features
- Enforces a clear contract for field-by-field transformation
- Automatically generates certain fields if not provided
- Provides a reusable and extensible template method pattern
- Built on Pydantic for type safety and validation

### DIFF
--- a/lib/python-sdk/DEVELOPMENT.md
+++ b/lib/python-sdk/DEVELOPMENT.md
@@ -7,51 +7,196 @@ This document provides instructions for common development tasks in the CommonGr
 - Python 3.11 or higher
 - Poetry for dependency management
 
-## Development Commands
-
-### Setup
-
-The commands described in this guide should be run from the `python-sdk` directory:
-```bash
-cd simpler-grants-protocol/lib/python-sdk
-```
+## Dependencies
 
 Install dependencies:
 ```bash
+cd simpler-grants-protocol/lib/python-sdk
 poetry install
 ```
 
-### Testing
+## Code Quality & Testing
+
+### Tests
 
 Run all tests:
 ```bash
-poetry run pytest
+cd simpler-grants-protocol/lib/python-sdk
+poetry run pytest tests/ -v
 ```
 
 Run tests with coverage report:
 ```bash
+cd simpler-grants-protocol/lib/python-sdk
 poetry run pytest --cov=common_grants --cov-report=term-missing
 ```
 
-### Code Quality
-
-#### Formatting
+### Formatting
 
 Format code with Black:
 ```bash
+cd simpler-grants-protocol/lib/python-sdk
 poetry run black .
 ```
 
-#### Linting
+### Linting
 
 Check code with Ruff:
 ```bash
+cd simpler-grants-protocol/lib/python-sdk
 poetry run ruff check .
 ```
 
-#### Type Checking
+### Type Checking
 
 Verify types with MyPy:
 ```bash
+cd simpler-grants-protocol/lib/python-sdk
 poetry run mypy .
+```
+
+## Usage Examples
+
+### Initialize Opportunity Model Instance
+
+```python
+from datetime import datetime, date, UTC
+from uuid import uuid4
+
+from common_grants.schemas.fields import Money, Event
+from common_grants.schemas.models.opp_base import OpportunityBase
+from common_grants.schemas.models.opp_funding import OppFunding
+from common_grants.schemas.models.opp_status import OppStatus, OppStatusOptions
+from common_grants.schemas.models.opp_timeline import OppTimeline
+
+# Create a new opportunity
+opportunity = OpportunityBase(
+    id=uuid4(),
+    title="Research Grant 2024",
+    description="Funding for innovative research projects",
+    status=OppStatus(
+        value=OppStatusOptions.OPEN,
+        description="This opportunity is currently accepting applications"
+    ),
+    created_at=datetime.now(UTC),
+    last_modified_at=datetime.now(UTC),
+    funding=OppFunding(
+        total_amount_available=Money(amount="100000.00", currency="USD"),
+        min_award_amount=Money(amount="10000.00", currency="USD"),
+        max_award_amount=Money(amount="50000.00", currency="USD"),
+        estimated_award_count=5
+    ),
+    key_dates=OppTimeline(
+        app_opens=Event(
+            name="Application Opens",
+            date=date(2024, 1, 1),
+            description="Applications open"
+        ),
+        app_deadline=Event(
+            name="Application Deadline",
+            date=date(2024, 3, 31),
+            description="Applications close"
+        )
+    )
+)
+```
+
+### Serialize / Deserialize
+
+```python
+# Create an opportunity instance
+opportunity = OpportunityBase(
+    ...
+)
+
+# Serialize to JSON
+json_data = opportunity.dump_json()
+
+# Deserialize from JSON
+loaded_opportunity = OpportunityBase.from_json(json_data)
+```
+
+### Transform Data Sources and Models
+
+When extending an existing system to adopt the CommonGrants protocol, a developer might need to transform existing data model implementations or data sources into canonical model instances. Such custom transformations can be easily implementing by leveraging the abstract base class `OpportunityTransformer`. 
+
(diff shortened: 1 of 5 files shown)


lib/python-sdk/DEVELOPMENT.md
```diff
@@ -7,51 +7,196 @@ This document provides instructions for common development tasks in the CommonGr
...
+
+# Deserialize from JSON
+loaded_opportunity = OpportunityBase.from_json(json_data)
+```
+
+### Transform Data Sources and Models
+
+When extending an existing system to adopt the CommonGrants protocol, a developer might need to transform existing data model implementations or data sources into canonical model instances. Such custom transformations can be easily implemented by leveraging the abstract base class `OpportunityTransformer`. 
+
+The `OpportunityTransformer` base class defines a standard interface for transforming raw input data (from third-party feeds, legacy formats, custom JSON, etc.) into structured `OpportunityBase` instances. As an abstract base class, `OpportunityTransformer` itself *does not contain any transformation logic*; custom transformation logic must be implemented by subclassing the abstract base class (see following example).
```
widal001, line 123: This doesn't align with the pattern that we're trying to establish here. Adopters shouldn't have to define a custom transformer in Python code.

The goal is to automate simple transformations using serializable mappings.

widal001 (changes requested):
Appreciate you taking a first stab at this, but this isn't quite the approach we want to take.

I implemented a basic pattern for automating transformations using a serializable mapping format in #193 sorry I didn't get a chance to do this before I went OOO to provide an example.

lib/python-sdk/DEVELOPMENT.md
DavidDudas-Intuitial: Let's discuss. IMO this is compatible. A transformer pattern and portable serializable mapping formats are not mutually exclusive and developers can benefit from one, the other, or both combined. 

> _"Adopters shouldn't have to define a custom transformer in Python code"_

Nor should they have to learn a complicated deeply-nested mapping format if they just want to instantiate an object from a known data source! :) IMO we should give them options.

The spirit of this PR is to provide a very simple scaffold for enabling the ADR stated goal (_"flexibly map existing data structures to the canonical models"_) within the accepted tradeoffs (_"requires a custom runtime to apply transformations in each new language or SDK"_) which can be extended to support any number of formats and transformers (such as the draft PR you opened today) OR used as-is without use of any mapping format.

widal001: While I agree these aren't _**incompatible**_ approaches, maybe we can talk tomorrow about the benefit that this transformer ABC is providing if we want to keep it as an option? 

Maybe you could outline the why someone might implement this ABC versus creating their own custom transformer function / class? Typically, the main use case for ABCs in Python is to support polymorphism, providing a consistent interface for downstream consumers of objects that implement this interface. Did you anticipate someone implementing multiple versions of this transformer? Or was there some specific downstream consumer you had in mind that would create the need for a consistent interface for the transformer class?

I _could_ see the value of providing a standard Python interface for transformations that would be consistent across CommonGrants models, like Opportunity (current) and Organization, Person, etc. (future), but it would probably make more sense to define an ABC that is generic (e.g. `ModelTransformer`) rather than specific to a given model.

I'll also take some time to provide feedback on some specific design choices in this transformer ABC, because even if we decide to proceed with this as an option, there are a few things I'd recommend tweaking.

lib/python-sdk/common_grants_sdk/schemas/models/opp_transformer.py
```diff
@@ -0,0 +1,98 @@
...
+        Extract and transform title data from source_data
+        into a canonical title string.
+        """
+
+        pass
+
+    @final
+    def _generate_id(self) -> UUID:
+        """Generate unique ID for the opportunity."""
+        return uuid4()
```
widal001, line 93: We might either want to replace this with a deterministic UUID method (e.g. v5) or at least provide an alternative internal function that supports deterministic UUID generation.

The current approach would generate two different UUIDs for the same input data.

lib/python-sdk/common_grants_sdk/schemas/models/opp_transformer.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+
+    source_data: Annotated[
+        dict[str, Any],
+        Field(
+            description="Arbitrary source data",
+        ),
+    ]
+
+    def transform_opportunity(self, id: UUID | None = None) -> OpportunityBase:
```
widal001, line 28: Related to my comments below on the `self._generate_id()` method, I think it would make more sense to have `id` be a required value that is either a uuid or a string that is used to generate a UUID so that successive runs of this function with the same input data will yield identical results.

widal001: Or if we're sticking with the pattern of having one instance of this transformer per record, expecting an ID value on initialization or generating and saving the UUID when the transformer is instantiated with input data.

lib/python-sdk/common_grants_sdk/schemas/models/opp_transformer.py
```diff
@@ -0,0 +1,98 @@
...
+from pydantic import Field
+
+from common_grants_sdk.schemas.base import CommonGrantsBaseModel
+from common_grants_sdk.schemas.models.opp_base import OpportunityBase
+from common_grants_sdk.schemas.models.opp_funding import OppFunding
+from common_grants_sdk.schemas.models.opp_status import OppStatus
+from common_grants_sdk.schemas.models.opp_timeline import OppTimeline
+
+
+class OpportunityTransformer(CommonGrantsBaseModel, ABC):
```
widal001, line 15: What's the motivation for inheriting from the `CommonGrantsBaseModel` and thus making this transformer class a pydantic schema?

Is there a reason someone implementing this function would call `transform_opportunity()` multiple times after instantiating a given transformer with a single input record?

Some alternative patterns that I think would make more sense:

- Treating a given transformer as a singleton class that accepts arbitrary input data to the `transform_opportunity()` method, so it can be used in a loop to transform a list of records without having to instantiate the class, then invoke `transform_opportunity()`
- Removing the inheritance from `CommonGrantsBaseModel` so it's not a pydantic schema, accepting input data and invoking `transform_opportunity()` as part of the `__init__()` method and storing the resulting model in a `model` so that folks could simply run `LegacyTransformer(data={...}).model` to get the resulting opportunity instance.

lib/python-sdk/common_grants_sdk/schemas/models/opp_transformer.py
```diff
@@ -0,0 +1,98 @@
...
+        ),
+    ]
+
+    def transform_opportunity(self, id: UUID | None = None) -> OpportunityBase:
+        """
+        Extract and transform opportunity data from source_data
+        into a canonical OpportunityBase model instance.
+        """
+
+        return OpportunityBase(
```
widal001, line 34: This works if folks only want to use the default `OpportunityBase` class, but if they want to extend the `OpportunityBase` class with custom fields (like outlined in [this tutorial](https://commongrants.org/guides/using-python/#extend-the-sdks-opportunitybase-model)) this approach won't work.

An alternative would be to make this a generic class to which can pass a sub-class of `OpportunityBase` when they are implementing this ABC.

PR #193, opened 2025-05-19
[Issue #185] Support mapping-based transformations

### LINKED ISSUE #185: [Py SDK] Support for mapping-based transformations
### Summary

Create a method for instantiating a CommonGrants model, e.g. `OpportunityBase` from an arbitrary object using a standardized mapping config.

The goal is to enable third-party developers to generate and publish a mappings for existing platforms that enable multiple developers to translate between those platforms and the CommonGrants standard without having to write custom code.

### Note

The exact mapping format is still be refined so we might want to experiment with a few options. See #184 for more information about the options being considered.

### Acceptance criteria

- [ ] Developers can automatically instantiate a CommonGrants model from arbitrary data using something like the following code:

```python
from common_grants.schemas import OpportunityBase

model = Organization.translate(
    mapping={
       "org.address.street1": {
          "field": "org_street_1"
      },
      "org.contact.fullName": {
        "concat": [
            { "field": "first_name" }, 
            { "literal": " " }, 
            { "field": "last_name" }
          ]
      }
    }
  },
  data={
    "organization_name": "Acme",
    "org_street_1: "123 Main Street",
    "contact": {
      "first_name": "John",
      "last_name": "Doe",
   },
  },
)
```

widal001:
### Summary

Adds support for mapping based transformations to the Python SDK.

- Fixes #185 
- Time to review: 10 minutes

- Adds `transform_from_mapping()` function that transforms arbitrary JSON from one format to another using a mapping format that serves as a kind of lightweight DSL.
- Adds a `validate_with_mapping()` to the `CommonGrantsBaseModel` to deserialize a dictionary after transforming it with a mapping.
- Adds a `dump_with_mapping()` to `CommonGrantsBaseModel` to serialize and transform a model to a python dictionary.
- Removes the extra `push` trigger from the `ci-python-sdk.yaml` to avoid duplicate runs of the CI checks on each PR.

1. Checkout the PR
2. Change directory into `python-sdk`: `cd lib/python-sdk/`


.github/workflows/ci-python-sdk.yml
```diff
@@ -1,14 +1,10 @@
...
 
 on:
-  push:
-    paths:
-      - 'lib/python-sdk/**'
-      - '.github/workflows/ci-python-sdk.yml'
   pull_request:
     paths:
-      - 'lib/python-sdk/**'
-      - '.github/workflows/ci-python-sdk.yml'
```
widal001, line 11: Removes the `push` trigger because it was causing duplicate runs of the CI checks on each PR. Not a big deal, but HHS admins recently asked us to try to reduce GitHub action usage wherever we can.

lib/python-sdk/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.analysis.typeCheckingMode": "basic"
+}
```
widal001, line 3: Enables Python type checking when using VSCode

lib/python-sdk/common_grants_sdk/schemas/base.py
```diff
@@ -18,13 +22,17 @@ def dump_json(self) -> str:
         """Convert model to JSON string (alias for model_dump_json for backward compatibility)."""
         return self.model_dump_json()
 
+    def dump_with_mapping(self, mapping: dict) -> dict:
+        """Convert model to dictionary with mapping."""
+        return transform_from_mapping(self.model_dump(mode="json"), mapping)
+
     @classmethod
-    def from_json(cls, json_str: str) -> "CommonGrantsBaseModel":
+    def from_json(cls, json_str: str) -> Self:
```
widal001, line 30: This improves the intellisense output, so that the class method displays the name of the sub-class being instantiated (if it inherits from this base model) instead of always displaying "CommonGrantsBaseModel".

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+    val = get_from_path(data, match_spec["field"])
+    return match_spec["case"].get(val, match_spec.get("default"))
+
+
+# Registry for handlers
+DEFAULT_HANDLERS: dict[str, handle_func] = {
+    "field": handle_field,
+    "const": handle_const,
+    "match": handle_match,
+}
```
widal001, line 86: Right now we just define a base set of reserved words and their transformation functions, but this pattern makes it really easy to extend the mapping with additional key words.

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def transform_from_mapping(
+    data: dict,
+    mapping: dict,
+    depth: int = 0,
+    max_depth: int = 500,
+    handlers: dict[str, handle_func] = DEFAULT_HANDLERS,
```
widal001, line 94: Allows library users to provide their own custom handlers if they'd like to extend the base set.

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+) -> dict:
+    """
+    Transforms a data dictionary according to a mapping specification.
+
+    The mapping can contain nested transformations and supports three types of transformations:
+    - field: Extracts a value from the data using a dot-notation path
+    - const: Returns a constant value
+    - match: Performs a case-based transformation based on a field value
```
DavidDudas-Intuitial, line 102: IMO the term `match` implies usage of a regex, and I didn't realize that's not what `handle_match()` does until I read this comment in `transform_from_mapping()`. Consider alternate nomenclature, e.g. `switch`.

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+        data: The source data dictionary to transform
+        mapping: A dictionary describing how to transform the data
+        depth: Current recursion depth (used internally)
+        max_depth: Maximum allowed recursion depth
+        handlers: A dictionary of handler functions to use for the transformations
+
+    Returns:
+        A new dictionary containing the transformed data according to the mapping
+    """
+    if depth > max_depth:
```
DavidDudas-Intuitial, line 114: What is the purpose of enforcing a max depth? I agree 500 is an absurd depth, but curious what the intent is. Fear of runaway recursion?

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+        raise ValueError("Maximum transformation depth exceeded.")
+
+    def transform_node(node: Any, depth: int) -> Any:
+        if depth > max_depth:
+            raise ValueError("Maximum transformation depth exceeded.")
+        if isinstance(node, dict):
```
DavidDudas-Intuitial, line 120: Consider obviating the `else` statement 5 lines down with an immediate transfer of control:
```
if not isinstance(node, dict):
    return node
```

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+            for k in node:
+                if k in handlers:
+                    return handlers[k](data, node)
+            return {k: transform_node(v, depth + 1) for k, v in node.items()}
```
DavidDudas-Intuitial, line 124: There's a lot going on in these 4 lines! Maybe add a brief comment to explain what's happening.

DavidDudas-Intuitial (commented):
Overall LGTM. 

Consider adding a README to explain usage of `transform_from_mapping()`.

lib/python-sdk/common_grants_sdk/utils/transformation.py
widal001: Good call, I added some comments and example to show what happens at each step in this commit: [refactor(py-sdk): Adds docs to transformation functions](https://github.com/HHS/simpler-grants-protocol/pull/193/commits/3b179f2a27ec324da425d56b1f36ccda93829689)

lib/python-sdk/common_grants_sdk/utils/transformation.py
widal001: Good call. I updated the reserved keyword here and in the handler function dictionary.

widal001: [refactor(py-sdk): Update handler functions](https://github.com/HHS/simpler-grants-protocol/pull/193/commits/62dc251b96ba91df3de64254f3f370ca189851f1)

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+    Handles a constant transformation by returning the specified constant value.
+
+    Args:
+        data: The source data dictionary (unused)
+        spec: A dictionary containing the constant value under the 'const' key
+
+    Returns:
+        The constant value specified in the spec
+    """
+    return spec["const"]
```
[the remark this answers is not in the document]
widal001: I refactored the access pattern a bit. The reason these were direct access instead of `get()` is that (previously) when this function was invoked we had already checked that the `spec` contains the key. But this is no longer relevant, since I updated the handler functions to accept the value of this key instead of passing the full dictionary with that key.

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -0,0 +1,128 @@
...
+    parts = path.split(".")
+    for part in parts:
+        if isinstance(data, dict) and part in data:
+            data = data[part]
+        else:
+            return default
+    return data
+
+
+def handle_field(data: dict, spec: dict) -> Any:
```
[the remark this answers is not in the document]
widal001: Good suggestion! I updated that here: [refactor(py-sdk): Update handler functions](https://github.com/HHS/simpler-grants-protocol/pull/193/commits/62dc251b96ba91df3de64254f3f370ca189851f1)

DavidDudas-Intuitial (approved):
Looks good. Nice work!

PR #197, opened 2025-05-22
Issue 146 update docs

### LINKED ISSUE #146: [Py SDK] Update docs
### Summary

Update the protocol reference docs to add a "Python" tab to [the "Formats" tab group](https://commongrants.org/protocol/fields/money/#formats) in each field, model, filter, etc. This tab should display the pydantic definition for that field.

We should also add a "Usage" section to the docs that includes tabs for both TypeSpec and Python and displays how to use each schema with the `@common-grants/core` and Python SDK respectively.

### Acceptance criteria

- [x] A "Python" tab has been added to each of the "Formats" tab groups with the implementation of a given schema in the Python SDK
- [ ] A "Usage" section has been added to schema in the protocol reference docs and it displays how to use that field in TypeSpec and in the Python SDK

DavidDudas-Intuitial:
### Summary

- Fixes #146 
- Time to review: 10 minutes

Added a new tab to the "Formats" tab groups for each field, filter, and model. Each new tab contains example python code illustrating how to implement the given entity.

### DIFF
--- a/website/src/content/docs/protocol/models/opp-base.mdx
+++ b/website/src/content/docs/protocol/models/opp-base.mdx
@@ -196,4 +196,12 @@ model OpportunityBase {
 ```
 
   </TabItem>
+
+  <TabItem label="Python">
+
+```python
+# TO DO: add content
+```
+
+  </TabItem>
 </Tabs>
--- a/website/src/content/docs/protocol/models/opp-base.mdx
+++ b/website/src/content/docs/protocol/models/opp-base.mdx
@@ -200,7 +200,68 @@ model OpportunityBase {
   <TabItem label="Python">
 
 ```python
-# TO DO: add content
+from datetime import date, datetime, time
+from uuid import UUID
+from common_grants_sdk.schemas.fields import Event, Money, CustomField, CustomFieldType
+from common_grants_sdk.schemas.models.opp_base import OpportunityBase
+from common_grants_sdk.schemas.models.opp_funding import OppFunding
+from common_grants_sdk.schemas.models.opp_status import OppStatus, OppStatusOptions
+from common_grants_sdk.schemas.models.opp_timeline import OppTimeline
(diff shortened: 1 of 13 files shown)


website/src/content/docs/protocol/models/opp-base.mdx
```diff
@@ -193,6 +193,77 @@ model OpportunityBase {
...
+The Python code for this model is:
+
+```python
+from datetime import date, datetime, time
+from uuid import UUID
+from common_grants_sdk.schemas.fields import Event, Money, CustomField, CustomFieldType
+from common_grants_sdk.schemas.models.opp_base import OpportunityBase
+from common_grants_sdk.schemas.models.opp_funding import OppFunding
+from common_grants_sdk.schemas.models.opp_status import OppStatus, OppStatusOptions
+from common_grants_sdk.schemas.models.opp_timeline import OppTimeline
```
widal001, line 211: Small nit, but part of the value of importing these models in the `common_grants_sdk/schemas/__init__.py` file is so that this can become a single `from common_grants_sdk.schemas import ...` or at least just two imports:
- `from common_grants_sdk.schemas.models  import...`
- `from common_grants_sdk.schemas.models  import...`

The benefit of that pattern (besides making the import lines less cluttered) is that consumers of the library API don't have to worry about how the classes are organized within the sub-packages, for example if we wind up putting `OpportunityBase` and `OppFunding` into a new sub-module for opportunities after adding models for `Organization`, `Person`, `Application` etc.

widal001 (approved):
LGTM! I left one small comment about the import patterns you use, but this looks perfect.

On the failing cspell check, you might be able to add a [python dictionary](https://www.npmjs.com/package/@cspell/dict-python) or otherwise just manually add `dataclass` and the other failing items to the `cspell.json` dictionary or ignore section

PR #208, opened 2025-06-03
[Issue 154] Introduce Changeset-based Versioning

### LINKED ISSUE #154: [Release CI/CD] Set up change sets in repo
### Summary

Set up change sets in the HHS/simpler-grants-protocol repo based on the learning from #153

### Acceptance criteria

- [ ] Change sets are configured in the repo and the first change set has been created

DavidDudas-Intuitial:
### Summary

- Fixes #154 
- Time to review: 10 minutes

- Support for semantic versioning using Changesets, extended with custom tooling for Python packages
- Independent versioning of monorepo packages, enabling granular release control
- Automatic changelog generation triggered by version bumps
- GitHub Actions workflows for versioning, tagging, and optional GitHub Releases
- Initial stubs for packages in `lib/cli` and `lib/core`
- Documentation explaining the versioning strategy and developer workflow


pnpm-workspace.yaml
widal001, line 1: Is there a reason we're using pnpm here when the rest of the project is using npm?

scripts/bump_python_version.py
widal001, line 1: Since we're already using `poetry` is there a reason we aren't using the [`poetry version`](https://python-poetry.org/docs/cli#version) command, since we'll also likely be using `poetry publish` to actually publish the package to PyPI?

If the _main_ value of this script is to parse the change sets to figure out the correct version to update (MAJOR, MINOR, PATCH) can we maybe add a little more documentation explaining that workflow?

scripts/bump_python_version.py
```diff
@@ -0,0 +1,107 @@
...
+
+    bump_priority = {"major": 3, "minor": 2, "patch": 1}
+    highest = max(bumps, key=lambda b: bump_priority[b])
+
+    current = get_current_version()
+    new_version = str(apply_bump(current, highest))
+
+    update_pyproject(new_version)
+    update_package_json(new_version)
+    update_changelog(new_version, highest)
```
widal001, line 104: Ahh I see that we're updating both the `pyproject.toml` and the `package.json`

scripts/bump_python_version.py
widal001: After diving into it further, I can see that it looks like it's doing 3 things:

- Parsing the change sets to get the latest updates to this package
- Using those updates to determine the new version number
- Updating the `pyproject.toml` _and_ the `package.json` with that version number

Maybe we can add that to the module-level docstrings just to help people quickly grok the purpose/flow of this file at a glance?

.github/workflows/ci-create-release-from-tag.yml
widal001, line 1: Just to confirm my understanding here, after merging this we'd modify our CD GitHub actions to watch for new releases before deploying, right?

Or would we update this GitHub action to trigger those other actions?

widal001 (approved):
I left a few questions just to clarify my understanding, 

Since you tested this in the [sample repo](https://github.com/DavidDudas-Intuitial/sgp-changeset-poc) and it seemed to work as expected I'll approve it!

.github/workflows/ci-package-version.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Tag Node packages (if any)
+        run: |
+          if pnpm changeset status | grep -q 'NO packages'; then
+            echo "No node packages to tag."
+          else
+            pnpm changeset tag
+          fi
+
+      - name: Push all tags
+        run: git push --tags
```
widal001, line 72: It looks like this line might be failing when there's a merge that doesn't include a new change set.

Can we confirm this works in the sandbox repo before we merge this in?

widal001: See this GitHub action run: https://github.com/DavidDudas-Intuitial/sgp-changeset-poc/actions/runs/15289752388/job/43006915668

widal001 (changes requested):
After looking more closely at the commit and GitHub action run history, it looks like the `ci-package-version.yml` workflow might be failing if a merged PR doesn't include a change set.

Before we merge this in, can we demonstrate on that sandbox repo that we can reliably merge in PRs that don't introduce a new tag and that this action doesn't fail in that scenario?

pnpm-workspace.yaml
DavidDudas-Intuitial: Yes, we're using a "fake" node package to trick changesets into versioning our python package. When using npm + changesets, I could not get npm to work nicely with the fake package, whereas pnpm + changesets handles the fake package without issue.

scripts/bump_python_version.py
DavidDudas-Intuitial: Good callout. I changed to workflow to use `poetry version`.

.github/workflows/ci-package-version.yml
DavidDudas-Intuitial: Good call! I added conditional logic to the workflow in the POC repo, to avoid versioning/tagging/etc. if there are no changesets (see https://github.com/DavidDudas-Intuitial/sgp-changeset-poc/actions/runs/15456289325/job/43509066917). I'll add similar logic to this PR in my next commit.

Screenshot 2025-06-04 at 6 30 12 PM

DavidDudas-Intuitial:
@widal001 please review again

.github/workflows/ci-package-version.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Install dependencies
+        run: poetry install --no-root
+        working-directory: lib/python-sdk
+
+      - name: Install tomlq
+        run: |
+          pip install tomlkit
+          curl -sSL https://raw.githubusercontent.com/sdispater/tomlkit/main/scripts/tomlq.py -o /usr/local/bin/tomlq
+          chmod +x /usr/local/bin/tomlq
```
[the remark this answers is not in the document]
DavidDudas-Intuitial: Fixed

PR #212, opened 2025-06-10
[Issue 156] Update FastAPI template to pass check spec

### LINKED ISSUE #156: [Py SDK] Make FastAPI template pass `check spec`
### Summary

Set up CI checks for the FastAPI template that automatically run `cg check spec` and update the template to pass that check.

### Acceptance criteria

- [x] A CI check runs on every PR for the FastAPI template
- [x] This CI check triggers the `cg check spec` command for the FastAPI template
- [x] The template passes that check

DavidDudas-Intuitial:
### Summary

- Fixes #156 
- Time to review: 10 minutes

Multiple updates to FastAPI template to make it compliant with base spec.

BEFORE:

before

AFTER:

after

### DIFF
--- a/templates/fast-api/src/common_grants/scripts/generate_openapi.py
+++ b/templates/fast-api/src/common_grants/scripts/generate_openapi.py
@@ -35,49 +35,35 @@ def convert_to_openapi_v3(schema: dict[str, Any]) -> dict[str, Any]:
     schema["openapi"] = OPENAPI_V3
 
     # Move schemas from components to definitions for OpenAPI 3.0.0 compatibility
-    if "components" in schema and "schemas" in schema["components"]:
-        schema["definitions"] = schema["components"]["schemas"]
-        del schema["components"]["schemas"]
-
-    # Remove any remaining components section if empty
-    if "components" in schema and not schema["components"]:
-        del schema["components"]
-
-    return rewrite_schema_refs(schema)
-
-
-def rewrite_schema_refs(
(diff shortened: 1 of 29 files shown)


DavidDudas-Intuitial:
Still need to fix linter errors

widal001:
WOOOHOOO!!! This is super exciting to see

templates/fast-api/src/common_grants/scripts/generate_openapi.py
widal001, line 1: I see what you mean here. After doing some googling around this issue, I think it has to do with how [Pydantic generates JSON schema](https://community.sanicframework.org/t/problem-with-generated-openapi-schema-via-pydantic/1223). [FastAPI does have an `openapi_version`](https://fastapi.tiangolo.com/reference/fastapi/#fastapi.FastAPI.openapi_version) but it doesn't seem to have any [meaningful effect](https://github.com/fastapi/fastapi/discussions/10748) on the output format.

This is a smart workaround for now! But in a coming sprint we should investigate whether the root of the issue is how the schema is being generated or how it's being parsed, which will help us figure out whether a long-term solution is to handle this in the template (as you're doing here) or in the `@common-grants/cli` package.

widal001 (approved):
Thanks for fixing all of these issues! The hack to get the OpenAPI spec working is a decent workaround for now, but we should investigate further.

FWIW, I think that this might be a FastAPI / Pydantic bug because I've run into a similar error when were trying to upload a generated OpenAPI spec to ChatGPT, which expects OpenAPI 3.1

PR #242, opened 2025-07-01
[Issue 234] Extend CLI check spec to validate OpenAPI 3.1 specs

### LINKED ISSUE #234: [CLI] Fix how `check spec` parses OpenAPI version 3.1
### Summary

Currently we have a workaround in the Fast API template to fix a parsing error when OpenAPI specs with version 3.1 are passed to `check spec` we should fix how these get parsed.

### Acceptance criteria

- [ ] `check spec` can correctly parse the OpenAPI spec auto-generated by FastAPI

DavidDudas-Intuitial:
### Summary

- Fixes #234 
- Time to review: 5 minutes

The existing `check spec` command in the CLI lib could validate OpenAPI 3.0 spec files but could _not_ validate OpenAPI 3.1 spec files. This PR adds that capability. Also, this PR removes from FastAPI template the now-obsolete code to convert spec files, because the conversion logic has been moved to CLI `check spec`.

### DIFF
--- a/templates/fast-api/src/common_grants/scripts/generate_openapi.py
+++ b/templates/fast-api/src/common_grants/scripts/generate_openapi.py
@@ -13,6 +13,7 @@ from fastapi.openapi.utils import get_openapi
 
 from common_grants.routes import opportunity_router
 
+
 def get_openapi_schema() -> dict[str, Any]:
     """
     Generate OpenAPI schema.
--- a/templates/fast-api/src/common_grants/scripts/generate_openapi.py
+++ b/templates/fast-api/src/common_grants/scripts/generate_openapi.py
@@ -3,109 +3,19 @@
 """
 Script to generate OpenAPI specifications for the CommonGrants API.
 
-This module provides functionality to generate OpenAPI specifications in different versions,
-with support for converting between OpenAPI 3.0.0 and FastAPI's default format.
+This module provides functionality to generate OpenAPI specifications.
 """
 
-import argparse
-from typing import Any, Union, cast
+from typing import Any
 
 from fastapi import FastAPI
 from fastapi.openapi.utils import get_openapi
 
 from common_grants.routes import opportunity_router
 
-# OpenAPI version constants
-OPENAPI_V3 = "3.0.0"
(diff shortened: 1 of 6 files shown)


DavidDudas-Intuitial:
The `CI - Template: FastAPI Run Checks` job will continue to fail until the new version of CLI is published.

lib/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@common-grants/cli",
-  "version": "0.1.0-alpha.12",
+  "version": "0.1.0",
```
widal001, line 3:
```suggestion
  "version": "0.0.0",
```

Shouldn't this be 0.0.0 based on how we did the first release of the core library?

templates/fast-api/src/common_grants/scripts/generate_openapi.py
widal001, line 1: Beautiful!! Look how clean this is now 🎉

.changeset/loose-cycles-see.md
```diff
@@ -0,0 +1,5 @@
+---
+"@common-grants/cli": patch
+---
+
+Extend check spec to support OpenAPI v3.1
```
widal001, line 5: Is there a reason we're making this a patch versus a minor release and setting the current version to v0.0.0?

Doesn't it make sense to make the first release v0.1.0 instead of v0.1.1?

widal001 (approved):
Woohoo! The fix is working for me (see screenshots below).

I'm approving, but I did have a question about why we're making this a patch release on top of v0.1.0, which would publish v0.1.1 as the first release, instead of a minor release on top of v0.0.0 which would publish v0.1.0

lib/cli/package.json
DavidDudas-Intuitial: Yes, that was just an oversight. I'll fix it. Good catch!

.changeset/loose-cycles-see.md
DavidDudas-Intuitial: Just an oversight. Fixed in most recent commit.

Issue #233, opened 2025-06-25
[CLI] Update how CLI is linked to core library


widal001:
### Summary

Update how CLI is linked to the core library so they can either be versioned independently or they are versioned together.

### Acceptance criteria

The CLI and Core libraries can either be versioned independently, without any dependencies on one another, _or_ CLI is automatically bumped when the core library is bumped.

DavidDudas-Intuitial:
@widal001 I believe our current configuration already supports the desired outcome for this task.

In [.changeset/config.json](https://github.com/HHS/simpler-grants-protocol/blob/9c00861fdac774cb6e17afa1901f08ede53cf575/.changeset/config.json#L9C41-L10C1) we set the `updateInternalDependencies` value to  `patch`. 

The effect is that whenever Core package version is bumped at any level (patch or higher), the following happens automatically:

1. CLI's dependency on Core is updated to match the new Core version 
2. CLI version gets a patch bump

This ensures that CLI remains compatible with Core. Also, CLI can still be versioned independently.

DavidDudas-Intuitial:
Now that CLI is in the changeset workspace, we can leverage this capability

DavidDudas-Intuitial:
<img width="867" alt="Image" src="https://github.com/user-attachments/assets/aab3542c-2514-4427-8e36-edf13400def8" />

DavidDudas-Intuitial:
Hmm, this did not behave as I expected (see screenshot). 

The config option `updateInternalDependencies` is set, and CLI has a dependency on Core, but when Core version is bumped by changesets I do not see the CLI version bumped.

DavidDudas-Intuitial:
Changeset [documentation](https://www.npmjs.com/package/@changesets/cli) says `changeset version` "Updates the versions for all packages described in changesets since last release along with any dependents inside the repo that are **out of range**". I think the "out of range" gives a clue why CLI version didn't get bumped along with Core: because the dependency is "^0.1.0" (not the carat ^) which is not out of range.

DavidDudas-Intuitial:
Related: https://github.com/changesets/changesets/discussions/920

DavidDudas-Intuitial:
Confirmed: removing the carat prefix from the dependency causes the dependency to be updated by `changeset version`.

<img width="863" alt="Image" src="https://github.com/user-attachments/assets/99fa029e-8ceb-49fb-bffc-e8029736bc78" />

DavidDudas-Intuitial:
[...]

Issue #167, opened 2025-04-10
[Core] Add models and routes for apply


widal001:
### Summary

Add a new set of routes to the `@common-grants/core` library to support the apply workflow

### Acceptance criteria

- [ ] The new routes are available in a pre-release of the `@common-grants/core` library

PR #246, opened 2025-07-03
Bump library TypeSpec versions

widal001:
### Summary

Bumps the version of the TypeSpec libraries included in the `lib/` packages to a stable release. This partially serves to test #233 and also unblocks some new TypeSpec functionality I need around versioning for #167 

- Bumps to v1.0 of for TypeSpec libraries used by `lib/core/` and `lib/cli`
- Fixes issues with core library after upgrading to v1.0 for TypeSpec modules
- Removes local package-lock.json for `lib/core/` and `lib/cli/` now that they are managed by the root-level package-lock.json
- Removes `prepublishOnly` script for CLI -- this was an outdated check for automatic pre-releases

This will be a test of linking our libraries together and releasing them together.

#### Validating that compiled outputputs matching

I did a recursive diff of the OpenAPI spec and JSON Schema outputs of the new (unpublished) core library and the currently published v0.1.0 library to check that they match. You can see the difference in compiler versions being use, but there are no differences between the outputs:

Screenshot 2025-07-03 at 6 58 07 PM
Screenshot 2025-07-03 at 6 58 15 PM

#### Testing change set bumps

Locally I also ran `pnpm changeset version` to confirm this will patch bump the version for CLI, and it seems to be working as expected.

Technically, we'd want to patch bump the CLI directly anyway because I also updated `@typespec/compiler` to v1.1.0, but the note I added about pinning the version of `@typespec/json-schema` was specific to Core.

Screenshot 2025-07-03 at 7 21 33 PM


lib/core/lib/core/fields/event.tsp
```diff
@@ -2,7 +2,7 @@ import "../types.tsp";
 
 namespace CommonGrants.Fields;
 
-using CommonGrants.Types;
```
widal001, line 5: The v1.0 version of the compiler doesn't support referencing absolute namespace paths, so I had to remove any shared part of the namespace. Otherwise the TypeSpec compiler would have raised an error about `CommonGrants.CommonGrants.Types`

lib/cli/scripts/prepublish-checks.sh
widal001, line 1: This was a stale script used to make sure we didn't accidentally publish the same pre-release version twice. Now version bumps are handled by changesets.

.github/workflows/ci-lib-cli-publish.yml
```diff
@@ -32,11 +32,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      # Checks that the version is not already published and
-      # that there are no uncommitted changes
-      - name: Run prepublish checks
-        run: npm run prepublishOnly
-
```
widal001, line 39: Removes this prepublish checks step now that version bumping is handled by changesets. We removed the same check in the Core library CI previously.

lib/core/package.json
```diff
@@ -51,11 +51,11 @@
...
   "peerDependencies": {
-    "@typespec/compiler": "^0.66.0",
-    "@typespec/http": "^0.66.0",
-    "@typespec/json-schema": "^0.66.0",
-    "@typespec/openapi3": "^0.66.0",
-    "@typespec/rest": "^0.66.0"
+    "@typespec/compiler": "^1.1.0",
+    "@typespec/http": "^1.1.0",
+    "@typespec/json-schema": "1.0.0",
+    "@typespec/openapi3": "1.0.0",
```
widal001, line 57: We're pinning these to v1.0.0 because there is a bug with the `@typespec/json-schema` emitter in v1.1.0 see: https://github.com/microsoft/typespec/issues/7828

Issue #235, opened 2025-06-25
[Template] Update FastAPI template to match v0.1.0 release


widal001:
### Summary

Update the FastAPI template to match the v0.1.0 release of the core library.

### Acceptance criteria

- [ ] FastAPI template passes `check spec` using the v0.1.0 version of the CLI and core library

DavidDudas-Intuitial:
PR https://github.com/HHS/simpler-grants-protocol/pull/250

PR #250, opened 2025-07-12
[Issue #214] Update PySDK and FastAPI template to support Core v0.1.0 release

### LINKED ISSUE #214: [Py SDK] Update Python SDK to match v0.1.0 release
### Summary

Update Python SDK and FastAPI template based on changes made to the core models in #162 and update the FastAPI template to pass `check spec`.

### Acceptance criteria

- [ ] The FastAPI template CI for `cg check spec` is using the latest CommonGrants OpenAPI spec
- [ ] The FastAPI template CI for `cg check spec` is passing without any errors

DavidDudas-Intuitial:
### Summary

- Fixes #214 and #235 
- Time to review: 10 minutes

This PR updates Python SDK and FastAPI template to adopt changes introduced in Core v0.1.0 release.

Python SDK
* Update fields and models
* Update tests

FastAPI template
* Update routes, filters, models, and Opportunity service
* Update tests
* Add new make target, `make check-spec-devmode`, to validate OpenAPI spec with local build of Core rather than current distribution of Core


DavidDudas-Intuitial:
Depends on https://github.com/HHS/simpler-grants-protocol/pull/252

widal001:
> Depends on #252

While the fix you introduced in #252 is the desired behavior for CLI checks, did you discover that bug because there's an optional field you're _not_ implementing in the SDK?

DavidDudas-Intuitial:
> did you discover that bug because there's an optional field you're not implementing in the SDK?

I found it after changing type on an existing implementation (e.g. key_dates: OppTimeline) to an optional (key_dates: Optional[OppTimeline]) which caused check-spec validation to fail.

DavidDudas-Intuitial:
(apologies for the edit of you comment, I mistakenly hit edit instead of quote-reply)

widal001:
> I found it after changing type on an existing implementation (e.g. `key_dates: OppTimeline`) to an optional (e.g. `key_dates: Optional[OppTimeline]`) which caused check-spec validation to fail.

Oh interesting, that actually highlights a different potential bug. Changing the optionality of a matching field shouldn't have raised a missing field error at all. It should have hit [this step](https://github.com/HHS/simpler-grants-protocol/blob/1e6da80437555cf2894c04bbdf377aa2806297af/lib/cli/src/commands/check/utils/check-schema-compatibility.ts#L126-L133) instead of [this step](https://github.com/HHS/simpler-grants-protocol/blob/1e6da80437555cf2894c04bbdf377aa2806297af/lib/cli/src/commands/check/utils/check-schema-compatibility.ts#L135-L144)

templates/fast-api/src/common_grants/services/utils.py
```diff
@@ -1,17 +1,19 @@
...
 
 from common_grants.schemas.models import (
+    OppFilters,
     OpportunityBase,
     OppStatusOptions,
 )
 from common_grants.schemas.pagination import PaginatedItems, PaginationInfo
 
-NAMESPACE = UUID("58315de5-1411-4c17-a394-561f1a47376f")
+NAMESPACE = UUID("6ba7b810-9dad-11d1-80b4-00c04fd430c8")
```
widal001, line 16: Any reason in particular we're changing the namespace?

templates/fast-api/src/common_grants/services/utils.py
```diff
@@ -29,62 +31,101 @@ def paginate(items: list, page: int, page_size: int) -> PaginatedItems[Opportuni
     )
 
 
+def build_applied_filters(filters: OppFilters) -> dict[str, Any]:
+    """
+    Build a dictionary of only the filters that were actually provided.
```
widal001, line 36: I like this! I wonder if this is a candidate for the SDK when we consider moving other shared transformation logic into it.

widal001 (approved):
Looks good!

I left a few comments, mostly with questions or suggestions about our use of pydantic built-in features, but otherwise seems solid enough to :shipit:

templates/fast-api/src/common_grants/services/utils.py
DavidDudas-Intuitial: Oops, that is not intentional. I'll fix it

templates/fast-api/src/common_grants/schemas/models/opp_search_request.py
```diff
@@ -1,15 +1,36 @@
...
+        examples=["Pre-school education"],
+    )
+    filters: Optional[OppFilters] = Field(
+        default_factory=OppFilters,
+        description="Filters to apply to the opportunity search",
+    )
+    sorting: OppSorting = Field(
+        default_factory=create_default_sorting,
+        description="The sort order to apply to the results",
+    )
```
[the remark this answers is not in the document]
DavidDudas-Intuitial: Thanks for catching this! This change should not have been committed.

templates/fast-api/src/common_grants/services/utils.py
DavidDudas-Intuitial: You are correct! Thank you for the suggestion

PR #252, opened 2025-07-12
Update CLI check-spec logic for optional properties

DavidDudas-Intuitial:
### Summary

- Fixes #[issue number]
- Time to review: 1 minutes

The `check-spec` command currently throws an error if optional fields are missing. I assume this is a bug! This PR changes the CLI `check-spec` command behavior to NOT throw errors if optional properties are missing.

### DIFF
--- a/lib/cli/src/commands/check/utils/check-schema-compatibility.ts
+++ b/lib/cli/src/commands/check/utils/check-schema-compatibility.ts
@@ -132,11 +132,15 @@ function checkObjectCompatibility(
     errors.addErrors(propErrors.getAllErrors());
   }
 
-  // Step 3: Handle missing props
+  // Step 3: Handle missing props - only flag as errors if they are required
+  const requiredProps = baseSchema.required || [];
   for (const propName of propsByStatus.missing) {
-    const propLoc = `${location}.${propName}`;
-    const error = missingFieldError(propLoc, propName, { ...ctx, baseSchema });
-    errors.addError(error);
+    // Only flag as missing if the property is actually required
+    if (requiredProps.includes(propName)) {
+      const propLoc = `${location}.${propName}`;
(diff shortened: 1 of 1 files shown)


DavidDudas-Intuitial:
@widal001 PTAL

lib/cli/src/commands/check/utils/check-schema-compatibility.ts
```diff
@@ -132,11 +132,15 @@ function checkObjectCompatibility(
...
   for (const propName of propsByStatus.missing) {
-    const propLoc = `${location}.${propName}`;
-    const error = missingFieldError(propLoc, propName, { ...ctx, baseSchema });
-    errors.addError(error);
+    // Only flag as missing if the property is actually required
+    if (requiredProps.includes(propName)) {
+      const propLoc = `${location}.${propName}`;
+      const error = missingFieldError(propLoc, propName, { ...ctx, baseSchema });
+      errors.addError(error);
+    }
```
widal001, line 143: Good catch! In the future we'll probably want to make this configurable with something like `--strict` or have optional fields get flagged as warnings.

But that can be scoped for a future improvement.

lib/cli/src/__tests__/commands/check/utils/check-schema-compatibility.test.ts
```diff
@@ -112,6 +112,36 @@ describe("Schema Compatibility Checks", () => {
     );
   });
 
+  // ############################################################
+  // Optional properties
+  // ############################################################
+
+  it("should ignore missing properties if they are optional in base schema", () => {
```
widal001, line 119: I added this unit test for the new check

widal001 (approved):
Good catch! Left a note about future options for configuring a `--strict` check that does flag missing optional fields, but that's a future improvement we can implement later.

PR #286, opened 2025-07-31
[Issue #284] Fix CLI bug with circular refs

### LINKED ISSUE #284: [Closeout] CLI - `check spec` fails with circular reference
### Summary

Trying to run `check spec` on an OpenAPI spec with a circular reference causes the CLI to crash with the following error:

```
Validation error: Maximum call stack size exceeded
```

Image

### Reproducibility

I tried multiple times and saw the bug every time

### How to reproduce the bug

1. Checkout the latest version of the main branch
2. Run `npm run dev -- check spec lib/openapi.yaml` from the root of `lib/cli`
3. You should get the error.
4. Comment out line 1707 in `lib/openapi.yaml` with the circular ref.
5. Re-run `npm run dev -- check spec lib/openapi.yaml` from the root of `lib/cli`
6. The command should pass

### Browser

Not applicable

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

widal001:
### Summary

Fix a bug in the CLI that broke `check spec` when OpenAPI specs contained circular references.

- Fixes #284 
- Time to review: 5 minutes

- Ignores circular references in `SwaggerParser.dereference()` to prevent the stack overflow
- Checks the compatibility of `additionalProperties` schemas between base and implementation specs
- Introduces a temporary fix to handle how `allOf` schemas are resolved
- Cleans up how base and implementation schema are parsed in `check spec`
- Adds support for parsing JSON as well as YAML files

#### Test instructions

1. Checkout `main`
2. Run `npm run dev -- check spec lib/openapi.yaml` from the root of `lib/core`
3. You should see something like this error:
    ```
    npm run dev -- check spec lib/openapi.yaml

    > @common-grants/cli@0.1.5 dev
    > ts-node src/index.ts check spec lib/openapi.yaml
    
    Validation error: Maximum call stack size exceeded
    npm error Lifecycle script `dev` failed with error:
    npm error code 1
    npm error path /Users/billy.daly/Code/agile-six/protocol/simpler-grants-protocol/lib/cli
    npm error workspace @common-grants/cli@0.1.5
    npm error location /Users/billy.daly/Code/agile-six/protocol/simpler-grants-protocol/lib/cli
    npm error command failed
    npm error command sh -c ts-node src/index.ts check spec lib/openapi.yaml
    ```
4. Now checkout `issue-284-fix-circular-ref-bug`
5. Re-run `npm run dev -- check spec lib/openapi.yaml`
6. It should pass with something like:
    ```
    npm run dev -- check spec lib/openapi.yaml
    
    > @common-grants/cli@0.1.5 dev
    > ts-node src/index.ts check spec lib/openapi.yaml
    
    Spec is valid and compliant with base spec
    ```

#### Notes

- I marked this as a PATCH release and I'd like to merge it so we have a stable version of v0.1 published before we release a new minor version with #283 
- This fix does cut some corners in how we reconcile schemas with `anyOf` because validating compatibility between two different schemas with potentially different sets of `anyOf` schemas is a bit complex from a design standpoint.

Screenshot 2025-07-30 at 9 38 22 PM


lib/cli/src/__tests__/commands/check/utils/check-schema-compatibility.test.ts
```diff
@@ -236,6 +236,32 @@ describe("Schema Compatibility Checks", () => {
     );
   });
 
+  it("should check the compatibility of additionalProperties schemas", () => {
```
widal001, line 239: The previous version of `check-schema-compatibility.ts` also missed checking for compatibility of the `additionalProperties` schemas if both the base spec and implementation spec allowed additional properties

lib/cli/src/commands/check/utils/check-matching-routes.ts
```diff
@@ -171,16 +171,15 @@ function checkQueryParameters(
 
     // Check parameter schema compatibility if schemas exist
     if (baseParam.schema && implParam.schema) {
+      // Flatten schemas before compatibility check
+      const flattenedBaseSchema = deepFlattenAllOf(baseParam.schema as OpenAPIV3.SchemaObject);
+      const flattenedImplSchema = deepFlattenAllOf(implParam.schema as OpenAPIV3.SchemaObject);
+
```
widal001, line 177: While we were able to remove duplicate calls to `deepFlattenAllOf()` in most places, this was actually a place where we _weren't_ flattening query params but needed to -- though the likelihood of having a nested `allOf` in a query param is quite low.

lib/cli/src/commands/check/utils/flatten-schemas.ts
```diff
@@ -1,38 +1,79 @@
...
+  if (!schema.anyOf || schema.anyOf.length === 0) {
+    return schema;
+  }
+
+  // For now, select the first option that's not a reference
+  for (let i = 0; i < schema.anyOf.length; i++) {
+    const option = schema.anyOf[i];
+    if (typeof option === "object" && option !== null && !("$ref" in option)) {
+      // Recursively process the selected option
+      const processedOption = deepFlattenAllOf(option as OpenAPIV3.SchemaObject);
```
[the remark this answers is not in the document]
widal001: We can't just match by index since the schemas included in the `anyOf` array may be in different orders, so we'll have to think about what it means for two schemas to be partial matches and how to report that in the `check spec` output.

lib/cli/src/commands/check/check-service.ts
```diff
@@ -27,25 +27,15 @@ export class DefaultCheckService {
    *   4) Checking that matching routes are compatible
    */
   async checkSpec(specPath: string, options: CheckSpecCommandOptions): Promise<void> {
-    // Read the spec file as raw YAML/JSON first
-    const specContent = fs.readFileSync(specPath, "utf8");
-    const rawSpec = yaml.load(specContent) as OpenAPISchema;
-
-    // Convert OpenAPI v3.1 to v3.0 if needed
-    const convertedSpec = convertOpenApiToV3(rawSpec) as OpenAPIV3.Document;
```
widal001, line 35: Moves this conversion logic to a shared function since CLI users can pass in their own base spec

DavidDudas-Intuitial (approved):

PR #292, opened 2025-08-07
[FastAPI template] Bugfix to avoid namespace collisions

DavidDudas-Intuitial:
### Summary

- Fixes #[issue number]
- Time to review: 1 minute

The service layer in FastAPI template defines a namespace, and that namespace gets duplicated into APIs that are bootstrapped from the template (e.g. the California API example and the upcoming Pennsylvania API example). This PR simply changes the namespace in FastAPI template to a value that obviously must be changed in bootstrapped APIs, to avoid potential for collisions.

### DIFF
--- a/templates/fast-api/src/common_grants/services/utils.py
+++ b/templates/fast-api/src/common_grants/services/utils.py
@@ -12,8 +12,8 @@ from common_grants.schemas import (
     OppStatusOptions,
 )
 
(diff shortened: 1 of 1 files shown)


DavidDudas-Intuitial:
See https://github.com/HHS/simpler-grants-protocol/pull/294

Issue #171, opened 2025-04-10
[Forms PoC] Update form mappings (as needed)


widal001:
### Summary

Update the form mappings for the forms in the form library PoC to match the application model in v0.2.0.

### Acceptance criteria

- [x] Form mappings match the v0.2.0 models
- [x] We can validate this programmatically

DavidDudas-Intuitial:
PR: https://github.com/HHS/simpler-grants-protocol/pull/296

DavidDudas-Intuitial:
<img width="376" height="843" alt="Image" src="https://github.com/user-attachments/assets/d3555664-0354-4e5b-96d0-fd24e0a0075f" />

PR #293, opened 2025-08-07
[Issue #171] Validate form library mappings

widal001:
### Summary

- Relates to #171 
- Time to review: 4 minutes

- Updates the CommonGrants JSON schema emitters to explicitly disallow additional properties
- Adds `npm run validate:schemas` to validate the form library schema mappings against the CommonGrants data models
- Adds a few dependencies to support this validation

#### Instructions to review

1. Checkout the PR locally
2. Run `npm run validate:schemas` from the root of the `website` directory

#### Notes 
Since setting up automatic validation was a bit more work than I anticipated, I want to merge in the validation code first, then I'll fix the form mappings to pass the validation in a separate PR.

Here's the PR preview: https://cg-pr-293.billy-daly.workers.dev/

You should see this output when you run `npm run validate:schemas` locally:

Screenshot 2025-08-06 at 10 11 09 PM

### DIFF
--- a/website/public/schemas/yaml/Address.yaml
+++ b/website/public/schemas/yaml/Address.yaml
@@ -35,6 +35,8 @@ required:
   - stateOrProvince
   - country
   - postalCode
+unevaluatedProperties:
+  not: {}
 examples:
   - street1: 123 Main St
     city: Anytown
--- a/website/src/lib/transformation.ts
+++ b/website/src/lib/transformation.ts
@@ -1,5 +1,4 @@
-import type { JsonValue, FormData, TransformOutput } from "./types";
-import { schemas } from "./schemas";
+import type { JsonValue } from "./types";
 
 /**
  * This module provides a utility function for transforming data using a mapping.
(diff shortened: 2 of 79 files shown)


website/public/schemas/yaml/Address.yaml
```diff
@@ -35,6 +35,8 @@ required:
   - stateOrProvince
   - country
   - postalCode
+unevaluatedProperties:
+  not: {}
 examples:
```
widal001, line 40: This explicitly disallows additional properties, which helps us catch mapping errors when extra fields are added (rather than silently allowing them)

website/src/lib/transformation.ts
```diff
@@ -208,42 +207,3 @@ export function transformWithMapping(
...
-
-/**
- * Maps JSON data from one form schema to another using the common data format as an intermediary.
- *
- * @param data - The source form data
- * @param sourceId - The ID of the source form schema
- * @param targetId - The ID of the target form schema
- * @returns A TransformOutput object containing the transformation results
- */
-export function mapJson(
```
widal001, line 220: Moves this to `playground/index.tsx` since it depends on the Vite glob function which only runs on build. This allows us to use the `transformWithMapping()` function in this file in the local script `npm run validate:schemas`

website/tspconfig.yaml
```diff
@@ -4,3 +4,5 @@ emit:
 options:
   "@typespec/openapi3":
     "omit-unreachable-types": false
+  "@typespec/json-schema":
+    "seal-object-schemas": true
```
widal001, line 8: Explicitly disallows additional properties

PR #365, opened 2025-10-08
[Issue #258] WEBSITE: Generate code examples in schema documentation

### LINKED ISSUE #258: [Website] Auto-generate JSON example in formats section
### Summary

Optionally auto-generate the JSON example for a schema from the JSON schema provided to keep the example in sync with the JSON schema.

### Acceptance criteria

- [ ] The JSON example in `SchemaFormatTabs` component can be auto-generated from a JSON schema file
- [ ] (Stretch) The auto-generated example defaults to using the example of from the `$ref` if a root level example is not already provided

DavidDudas-Intuitial:
### Summary

- Fixes #258 
- Time to review: 5 minutes

This PR introduces the capability to generated JSON code examples from YAML schema files, and updates multiple schemas to use the new capability:

* Extends `SchemaFormatTabs` component with new `generateCodeFromSchema()` function which generates example code for the component's "Example" tab from a YAML schema file
* Resolves nested references within schema files, including support for `allOf`, `anyOf`, and `oneOf`
* Leverages schema example code if available and validates example code against schema properties
* Finds gaps in schema example code and autofills them with generated code
* Removes hardcoded examples code from 20 schema files and replaces it with function call

Note: this PR also commits files that are automatically generated/updated by `npm run build` (does this require a bump to website version?)

```bash
$ poetry install
$ npm run build
$ npm run dev
```
Then visit http://localhost:4321/


DavidDudas-Intuitial:
@widal001 Not sure if you're aware of this: running `npm run build` causes `FormBase.yaml` to be deleted and `Form.yaml` to be generated. I committed the changes in this PR but then the build failed in CI, so I reverted them.

website/public/openapi/openapi.0.1.0.yaml
```diff
@@ -13,9 +13,7 @@ tags:
   - name: Competitions
     description: Endpoints related to competitions, which are distinct application processes for the same funding opportunity
   - name: Applications
-    description: Endpoints related to submitting applications for a given competition
-  - name: Application Reviews
-    description: Endpoints related to reviewing applications for a given competition
```
widal001, line 18: I think these changes got added because you might not have re-run `npm install` to pick up the latest `@common-grants/core` library (the pre-release)

The same is true for all other changes in the `/website/public/` folder

website/src/components/SchemaFormatTabs.astro
```diff
@@ -131,6 +151,241 @@ function extractCode(props: CodeBlockConfig): string {
...
+      // Just filename reference
+      filePath = join(repoRoot, 'website/public/schemas/yaml', ref);
+    }
+    
+    const content = readFileSync(filePath, "utf-8");
+    return yaml.load(content, { schema: yaml.CORE_SCHEMA }) as JsonSchema;
+  } catch {
+    // If we can't load the referenced schema, return null
+    return null;
+  }
```
widal001, line 386: We're already collecting and resolving all of these schemas in [validation.ts](https://github.com/HHS/simpler-grants-protocol/blob/c2dab45b9bbad9576b6160bba2802bad18151537/website/src/lib/validation.ts#L13-L15)

If you create an `ajv` instance using `createAjvWithSchemas()` (or export that instance) then you can replace this with:

```ts
  // Use AJV to resolve external references
  try {
    const resolvedSchema = ajv.getSchema(ref);
    return (resolvedSchema?.schema as JsonSchema) || null;
  } catch {
    return null;
  }
```

The other bonus of using the existing ajv instance is that we can actually validate the generated code against the JSON schema to make sure it's valid

website/src/components/SchemaFormatTabs.astro
```diff
@@ -131,6 +151,241 @@ function extractCode(props: CodeBlockConfig): string {
...
+        }
+      }
+      return obj;
+    }
+
+    case 'array':
+      if (schema.items) {
+        const itemExample = generateJsonFromSchemaObject(schema.items as JsonSchema, visitedRefs, rootSchema);
+        return [itemExample];
+      }
```
widal001, line 311: As I was reviewing this, I noticed that this file doesn't seem to be adhering to the `prettier` standards that other TypeScript files in the website project are (e.g. single quotes, line wrapping, etc.) and realized it was because we don't have the `prettier-plugin-astro` added as a dev dependency yet.

Can you add this and update `.prettierrc` to the following:

```
{
  "trailingComma": "all",
  "tabWidth": 2,
  "plugins": ["prettier-plugin-astro"],
  "overrides": [
    {
      "files": "*.astro",
      "options": {
        "parser": "astro"
      }
    }
  ]
}
```

Then these formatting discrepancies should get resolved.

website/src/content/docs/protocol/models/form.mdx
```diff
@@ -5,65 +5,8 @@ sidebar:
...
-          "formType": {
-            "name": "Form Type",
-            "fieldType": "string",
-            "value": "application",
-            "description": "The type of form"
-          }
-        }
-      }
+    file:
+      path: "website/public/schemas/yaml/FormBase.yaml"
```
widal001, line 9: Soooo much cleaner ❤️

widal001 (changes requested):
Great work! Love that we're able to auto-generate the JSON and it makes the docs look sooo much cleaner.

A couple of important changes to make:

- Make sure you've installed the latest dependencies, specifically the latest `@common-grants/core`. I'm pretty sure that's the reason you were detecting changes to the yaml files when running `npm build`. Right now your PR is reverting some changes that we want to preserve from the pre-release.
- In addition, could you add the Astro plugin for prettier to dev dependencies and re-run `npm run check:format` locally to fix some of the formatting issues in `SchemaFormatTabs.astro`
- Less critical, but I'd also recommend using the `createAjvWithSchemas()` and using that for ref resolution, then also validating the generated JSON against the top-level schema using ajv.

PR #373, opened 2025-10-11
[Issue #257] WEBSITE: Create and integrate SchemaTable component

### LINKED ISSUE #257: [Website] Create `SchemaTable` component
### Summary

Create a `SchemaTable` component that auto-generates the [table section](https://commongrants.org/protocol/models/opportunity/#table) for a given schema using its JSON schema.

### Acceptance criteria

- [ ] Maintainers can insert a table for a schema into the associated MDX file using something like the following code:
     ```mdx
     <SchemaTable schema={frontmatter.opportunityBase.table} />
     ```
- [ ] The "Type" column in this table correctly links to the associated section in the docs

DavidDudas-Intuitial:
### Summary

- Fixes #257 
- Time to review: 10 minutes

This PR introduces a `SchemaTable` component and a set of supporting utils. The architecture ensures that the necessary-but-expensive operations of crawling YAML files, analyzing schema structures, and computing type formatting happen once at build-time, while runtime operations are limited to fast JSON file reads and simple table generation.

**Core Components**
- `SchemaTable` - Main Astro component that renders schema data as HTML tables
- `SchemaLoader` - Class for loading and parsing YAML schema files
- `SchemaDocMapper` - Class for mapping schema names to their corresponding documentation paths
- `SchemaMetadataLoader` - Class for loading and querying the pre-computed schema metadata index
- `TableGenerator` - Class for generating structured table rows from JSON schemas
- `TypeFormatter` - Class for converting JSON Schema types to displayable HTML with clickable links

**Build Time Scripts**

YAML parsing, schema analysis, and complex type resolution all happen once during build via the following scripts:
- `generate-schema-docs.ts` - Crawls YAML schema files and generates `schema-docs-mapping.json` to map schema names to doc paths
- `generate-schema-metadata.ts` - Processes all schemas and generates an index of schema metadata in `schema-metadata.json`
- `generate-type-formatting.ts` - Pre-computes type formatting results and generates `type-formatting-cache.json` to enable type display formatting

*BEFORE*
before

*AFTER*
after

### DIFF
--- a/website/src/components/SchemaTable.astro
+++ b/website/src/components/SchemaTable.astro
@@ -11,7 +11,7 @@ const { filePath, code } = Astro.props;
  * Props for the SchemaTable component.
  */
 interface Props {
-  /** Path to the schema file relative to the repository root, or inline schema code */
+  /** Path to the schema file relative to the repository root */
   filePath?: string;
   /** Inline schema code (YAML string) */
   code?: string;
@@ -23,10 +23,10 @@ const getSchemaDocPath = (schemaName: string): string | undefined => {
 };
 
 /**
- * Main component logic with proper error handling
+ * Main component logic with error handling
  */
 
-// Load and parse the schema with error handling
+// Load and parse the schema
 let schemaData: JsonSchema;
 try {
   if (filePath) {
@@ -41,7 +41,7 @@ try {
   throw error;
 }
 
-// Determine table type and generate rows with error handling
+// Determine table type and generate rows
 const isEnumSchema = SchemaLoader.isEnumSchema(schemaData);
 const isObjectSchema = SchemaLoader.isObjectSchema(schemaData);
 


website/public/schema-docs-mapping.json
widal001: If we're dynamically generating this on every build, why commit this to the repo and why put it in the `public/` folder which serves hosted content that site visitors access?

It seems more appropriate to store this in a .gitignored local folder

website/public/type-formatting-cache.json
widal001: Same comment as the other two auto-generated files

widal001 (commented):
I started going through and adding some specific comments, but I'd actually like us to take a step back and think about our broader approach here and the maintenance tradeoff associated with this new component.

We're adding ~5k lines to the codebase to save ~300 in the docs themselves. Granted those 300 lines require more regular upkeep than the 5k, but it might be worth thinking about this problem differently.

We're trying to solve a few related but distinct problems:

1. Checking that the table properties match the JSON schema
2. Automatically updating the table as the JSON schema changes
3. Automatically linking to related sections of the docs based on type references

You'd have a better sense but it seems like there's an asymmetry between how important each problem is to solve and how much code we're adding to solve it.

In other words, I'm wondering if could get a similar amount of value from a linter that just checks each table has the correct fields. And avoids the complexity of trying to validate the links (since we have a separate build-time step for link checking) or auto-update the table.

### Brainstorming alternative patterns

On a related note, I'm wondering if it makes sense to offload some of the ref resolution and `allOf` merging work to well-established third-party libraries like `json-schema-ref-parser` and `json-schema-merge-allof`.

I haven't checked this code, so there are probably adjustments that we'd need to make, but here's an example script for generating a list of rows from an `ajv` schema using some of those tools:

#### Getting an `ajv` instance from a bundled JSON schema

TypeSpec gives us the option of emitting a bundled JSON schema.

We could use that bundled schema and `json-schema-merge-allOf` to get a simpler schema registry and avoid a lot of the file system walking and merging with something like this code:

```ts
// ajv-merged-instance.ts
import Ajv, { Options as AjvOptions } from "ajv";
import mergeAllOf from "json-schema-merge-allof";

/**
 * Takes a *bundled* JSON Schema object, merges allOf (deep), and returns an Ajv instance
 * with the merged schema registered. `$ref`s remain as pointers (since input is bundled).
 */
export function createAjvWithMergedAllOf(
  bundledSchema: any,
  {
    schemaId,
    ajvOptions,
  }: {
    schemaId?: string;          // fallback if $id is missing
    ajvOptions?: AjvOptions;    // pass-through Ajv options
  } = {}
): Ajv {
  const merged = mergeAllOf(bundledSchema, { deep: true });

  const ajv = new Ajv({
    strict: false,               // relax if your schemas aren’t fully strict
    ...ajvOptions,
  });

  const id = merged.$id || schemaId || "MergedRoot";
  ajv.addSchema(merged, id);

  return ajv;
}
```

### Generate a list of rows for a given schema

If we drop the complexity of generating the doc section links, we should at least be able to generate the other columns for each row with something like this.

Note: We'd still have to tweak the `type` formatting and output a little, but I think this gets us like 80% of the way to having the rows we'd need for a linter.

```ts
// rows-top-level.ts
export type Row = {
  name: string;        // top-level property key
  type: string;        // e.g., "string", "enum<string>", "const<number>", "array<UserId>", "UserId"
  description?: string;
  required: boolean;
};

function shortenRef(ref: string): string {
  const idx = ref.indexOf("#");
  const fragment = idx >= 0 ? ref.slice(idx + 1) : ref; // may start with "/"
  const path = fragment.replace(/^\/+/, "").split("/");
  const last = path[path.length - 1];
  return last || ref;
}

function normalizeType(t: unknown): string | undefined {
  if (!t) return undefined;
  return Array.isArray(t) ? t.join(" | ") : String(t);
}

/** No inference from enum/const values — only declared "type" or $ref. */
function innerTypeLabel(node: any): string {
  if (!node) return "unknown";

  if (node.$ref) return shortenRef(String(node.$ref));

  const declared = normalizeType(node.type);

  if (node.enum) return declared ? `enum<${declared}>` : "enum";
  if (Object.prototype.hasOwnProperty.call(node, "const")) {
    return declared ? `const<${declared}>` : "const";
  }

  if (declared) return declared;

  if (node.items) return "array";
  if (node.properties) return "object";

  return "unknown";
}

function wrapContainerTypes(node: any, base: string): string {
  if (node?.type === "array" || node?.items) {
    const item = innerTypeLabel(node.items);
    return `array<${item}>`;
  }
  return base;
}

/** Assumes `schema` is already: (a) bundled, (b) allOf-merged. */
export function schemaTopLevelRows(schema: any): Row[] {
  const rows: Row[] = [];
  const props = schema?.properties ?? {};
  const requiredSet = new Set<string>(schema?.required ?? []);

  for (const [propName, propSchema] of Object.entries<any>(props)) {
    const base = innerTypeLabel(propSchema);
    const typeLabel = wrapContainerTypes(propSchema, base);

    rows.push({
      name: String(propName),
      type: typeLabel,
      description: propSchema?.description,
      required: requiredSet.has(String(propName)),
    });
  }
  return rows;
}
```

website/public/schema-docs-mapping.json
DavidDudas-Intuitial: Tnanks for the suggestion. I've removed the generated files from the repo and from the public folder.

website/public/schema-metadata.json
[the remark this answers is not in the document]
DavidDudas-Intuitial: Done

website/src/lib/schema/paths.ts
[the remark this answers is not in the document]
DavidDudas-Intuitial: Good suggestion. Done.

DavidDudas-Intuitial:
@widal001 Thanks for the PR comments and the chat today. After removing generated cache files from the branch, this PR adds +1377 net-new lines of code, which I think is reasonable to solve the maintenance and reliability problems you identified. In addition to solving those problems, please also note the new `SchemaTable` component has excellent performance with an **average render time of 0.11ms**, as demonstrated in the following snapshot of load times for the 6 instances on the Opportunity model page:
```
[SchemaTable-4v0fldb3a] Total SchemaTable render time: 0.23ms
[SchemaTable-yasisic5x] Total SchemaTable render time: 0.10ms
[SchemaTable-5bystn1eq] Total SchemaTable render time: 0.06ms
[SchemaTable-2d62x9xhd] Total SchemaTable render time: 0.15ms
[SchemaTable-zaj0z7hlx] Total SchemaTable render time: 0.06ms
[SchemaTable-1n2ospiz1] Total SchemaTable render time: 0.07ms
AVERAGE: 0.11 ms
```
By way of comparison, the existing `SchemaFormatTabs` component has an **average render time of 20.66ms**, as demonstrated in the following snapshot of load times for the 6 instances on the Opportunity model page:
```
[SchemaFormatTabs-mf24q3amn] Total SchemaFormatTabs render time: 38.77ms
[SchemaFormatTabs-t87aynclb] Total SchemaFormatTabs render time: 28.77ms
[SchemaFormatTabs-2k3mf46y6] Total SchemaFormatTabs render time: 13.86ms
[SchemaFormatTabs-wjc6yi0rb] Total SchemaFormatTabs render time: 13.11ms
[SchemaFormatTabs-sf3k26zga] Total SchemaFormatTabs render time: 14.21ms
[SchemaFormatTabs-vy2zjal7i] Total SchemaFormatTabs render time: 14.26ms
AVERAGE: 20.66 ms
```
In other words: the new `SchemaTable` component is 188x faster to render than the similar `SchemaFormatTabs` component, due to the pre-build/pre-cache architecture instead of runtime file I/O and code generation, and therefore I am bullish on this approach! :)

Nonetheless, I like your idea of simplifying ref-resolution and allOf-merging logic, but I would consider that a nice-to-have future improvement rather than a blocker for this PR.

website/cache/README.txt
widal001: Appreciate this note

widal001 (approved):
LGTM!

Thanks for the conversation yesterday, and as you highlighted in your last comment, the additional lines of code for the build-time optimizations and the reduced overhead seems like a worthwhile tradeoff.

I can add an item to the improvement backlog about simplifying some of the ref resolution and merging behavior, but that's an optimization for a future day.

Let's get this merged in and benefit from the sweet, sweet improvements of faster doc updates.

PR #405, opened 2025-11-07
Update automated-codejson-generator to use main branch

DinneK:
This patch moves the workflow for automatically creating codeJSON from @v1.2.0 which has breaking changes to @main, which should be a permanent fix.


widal001 (approved):
Thanks for this fix @DinneK

PR #421, opened 2025-11-18
[Issue #419] Implement base methods and automated pagination handling

### LINKED ISSUE #419: [Py client] Add ability to get all results from opportunity.list()
### Summary

Add automatic pagination iteration to opportunity.list() method, such that callers can get all records in one method call. 

### Acceptance criteria

Callers to opportunity.list() can optionally get all records with one method call (as opposed to calling method multiple times to iterate through pages).

DavidDudas-Intuitial:
### Summary

- Fixes #419 
- Time to review: 10 minutes

Extends PySDK Client:
- encapsulates common methods (get, list, etc) for CommonGrants resources
- handles manual or automatic pagination
- enforces max batch size limit 
- adds tests


lib/python-sdk/common_grants_sdk/client/base.py
```diff
@@ -0,0 +1,185 @@
...
+            client: Client instance for making API requests
+            path: Base path for the resource (e.g., "/common-grants/opportunities")
+        """
+        self.client = client
+        self.path = path
+
+    def list(
+        self,
+        page: int | None = None,
+        page_size: int | None = None,
```
widal001: We should probably add an optional `total` or a `limit` param (or something like a `max_requests`) so this method doesn't accidentally burn through an SDK user's rate limits if they're listing resources that have thousands of items.

Like if we tested this out on SGG production API there's a good chance it could make like 40+ round trips and return like 4000+ records without someone even realizing it.

widal001 (commented):
Excited that this begins to abstract common tasks like pagination to shared methods but see my comments below about:

- Our choice of inheritance (`BaseResource`) vs composition (methods on `Client`)
- How we'd re-use pagination logic across `search()` and `list()` when `BaseResource._list_all()` doesn't support request payloads.
- Why we're returning tuples from `BaseResource.list()` and `BaseResource.get()` when we already have generic response models that wrap the raw ingredients for those tuples in a more type safe class.

lib/python-sdk/common_grants_sdk/client/base.py
```diff
@@ -0,0 +1,185 @@
...
+    ) -> tuple[DefaultResponse, list[dict], PaginatedResultsInfo]:
+        """Fetch a set of items from an endpoint.
```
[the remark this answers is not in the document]
DavidDudas-Intuitial: Good callout. I was looking for a single model that contained all components but didn't find it. Now I see that `Paginated` does indeed exist and is defined in `responses/success.py` (rather than `pagination.py`).

DavidDudas-Intuitial:
@widal001 @jcrichlake Thanks for the discussion today which cleared up questions about general direction. I spent some more time on this PR so it now includes all of the following:
- composition pattern for common methods
- getter methods return Pagination instances instead of tuples
- max items limit when fetching all items
- better method names to avoid need for overrides

lib/python-sdk/common_grants_sdk/client/base.py
DavidDudas-Intuitial: Done, but added as a property of Config, rather than a param to list()

PR #430, opened 2025-11-21
Issue 386 opportunity search

### LINKED ISSUE #386: [Py client] Implement `Client.opportunities.search()`
### Summary

Expand the `Client` to also support a `Client.opportunities.search()` method which makes calls to the `POST /common-grants/opportunities/search/` endpoint. The first iteration can support only default fields and filters.

### Acceptance criteria

- [ ] SDK users can search opportunities with the following code:

```python
from commongrants.client import Client, Auth
from commongrants.schemas.pydantic import Opportunity

client = Client(
    base_url="https://api.example.org",
    auth=Auth.bearer("YOUR_TOKEN"),
    timeout=10,         # seconds
)

# List opportunities with transparent pagination and typed results
# If `total=` exceeds max batch size (100) then paginate the results
for opp in client.opportunities.search(q="Education", status="open", paginate=True):
    print(opp.id, opp.title)
```

jcrichlake:
### Summary

- Fixes #386 
- Time to review: 20 minutes

Added a generic search endpoint to the client as well as a opportunity specific search endpoint

I set up the opportunity search to take the opportunity search request object since that seemed like a better developer experience for users of the SDK. 

You can run the example code for search_opportunity.py to verify the return value of the search

### DIFF
--- a/lib/python-sdk/common_grants_sdk/client/opportunity.py
+++ b/lib/python-sdk/common_grants_sdk/client/opportunity.py
@@ -155,5 +155,14 @@ class Opportunity:
         response_data = paginated_response.model_dump(by_alias=True)
         response_data["items"] = items
 
+        response_data["sortInfo"] = {
+            "sortBy": "",
+            "sortOrder": "",
+            "customSortBy": None,
+            "errros": [],
+        }
+
+        response_data["filterInfo"] = {"filters": {}, "errors": []}
+
         # Hydrate OpportunitiesListResponse from response data
         return OpportunitiesSearchResponse.model_validate(response_data)
--- a/lib/python-sdk/common_grants_sdk/schemas/pydantic/responses/success.py
+++ b/lib/python-sdk/common_grants_sdk/schemas/pydantic/responses/success.py
@@ -39,7 +39,7 @@ class Paginated(Success, Generic[ItemsT]):
         alias="paginationInfo",
     )
 
-    model_config = {"populate_by_name": True, "extra": "allow"}
+    model_config = {"populate_by_name": True}
 
 
 class Sorted(Paginated[ItemsT], Generic[ItemsT]):
--- a/lib/python-sdk/common_grants_sdk/client/pagination.py
+++ b/lib/python-sdk/common_grants_sdk/client/pagination.py
@@ -103,15 +103,6 @@ def pagination(single_page_func: Callable) -> Callable:
         items = items[: self.config.list_items_limit]
 
         # Build aggregated response
-        # Preserve any extra fields from the latest response (like sortInfo, filterInfo)
-        extra_fields = {}
-        if latest_response:
-            response_dict = latest_response.model_dump(by_alias=True)
-            # Extract fields that aren't part of the base Paginated model
-            for key in response_dict:
-                if key not in {"status", "message", "items", "paginationInfo"}:
-                    extra_fields[key] = response_dict[key]
-
         return Paginated[ItemsT](
             status=latest_response.status if latest_response else 200,
             message=latest_response.message if latest_response else "Success",
@@ -122,7 +113,6 @@ def pagination(single_page_func: Callable) -> Callable:
                 totalItems=len(items),
                 totalPages=1,
             ),
-            **extra_fields,
         )
 
     return wrapper


lib/python-sdk/common_grants_sdk/client/opportunity.py
```diff
@@ -96,3 +97,57 @@ def get(self, opp_id: str | UUID) -> OpportunityBase:
 
         # Return the OpportunityBase from the response
         return opportunity_response.data
+
+    def search(
+        self,
+        search: OpportunitySearchRequest,
+        page: int | None = None,
+        page_size: int | None = None,
+    ) -> OpportunitiesListResponse:
```
DavidDudas-Intuitial: Shouldn't this return `OpportunitiesSearchResponse`?

lib/python-sdk/common_grants_sdk/client/client.py
```diff
@@ -226,6 +226,135 @@ def list_some_items(
...
+        """
+
+        if page:
+            return self.search_some_items(path, request_data, page, page_size)
+        else:
+            return self.search_all_items(path, request_data, page_size)
+
+    def search_all_items(
+        self, path: str, request_data: dict[str, Any], page_size: int | None = None
+    ) -> Paginated[ItemsT]:
```
DavidDudas-Intuitial: I'm not sure how I feel about `search_all_items` being a near clone of `list_all_items`, and `search_some_items` being a near clone of `list_some_items`, with the only diff being the getter called by each to fetch data. 
How did you reason about this and arrive at this approach?

jcrichlake: I had a couple of reasons why I settled on this approach, 

1. I didn't want to have both GET and POST calls taking the same route through the code in order to minimize the blast radius of potential future bugs.

2.  I didn't want to muddy up the method signatures with the combined approach by having to potentially pass info for both POST and GET through the various levels of calls

3. I also wasn't sure how divergent these paths could become in the future.  I could see a world where we spend time trying to combine the paths only to have to split them back out again 

I'm open to refactoring this if we decide combining them is better.

lib/python-sdk/common_grants_sdk/client/opportunity.py
```diff
@@ -96,3 +97,57 @@ def get(self, opp_id: str | UUID) -> OpportunityBase:
...
+        """Search for opportunties by a query string
+
+        Args:
+            query: the string to search for
+            status: the status of the opportunities to return
+            paginate: whether to paginate the responses or not
```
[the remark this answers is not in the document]
jcrichlake: Fixed

lib/python-sdk/common_grants_sdk/client/client.py
```diff
@@ -214,9 +178,9 @@ def list_some_items(
...
 
         try:
-            # TODO: if params is not None then post instead of get
-            # see: https://github.com/HHS/simpler-grants-protocol/issues/386
-            api_response = self.get(path, params={"page": page, "pageSize": page_size})
+            api_response = self.post(
+                path, data=request_data, params={"page": page, "pageSize": page_size}
+            )
             api_response.raise_for_status()
             result_dict = Paginated[dict].model_validate(api_response.json())
```
DavidDudas-Intuitial: @jcrichlake Is this where you wanted the `Success` schema (base class of `Paginated`) to support extra properties, so the hydrated model includes all values from the api response? If so, that is probably not a good enough reason to change the schema, unless there is indeed a gap between the protocol and the schema (e.g. protocol allows extra properties but schema disallows). cc: @widal001

lib/python-sdk/common_grants_sdk/schemas/pydantic/responses/success.py
```diff
@@ -39,7 +39,7 @@ class Paginated(Success, Generic[ItemsT]):
         alias="paginationInfo",
     )
 
-    model_config = {"populate_by_name": True}
+    model_config = {"populate_by_name": True, "extra": "allow"}
```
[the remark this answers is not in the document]
DavidDudas-Intuitial: For an alternate, safer option, consider extending or adding to [client/response.py](https://github.com/HHS/simpler-grants-protocol/blob/main/lib/python-sdk/common_grants_sdk/client/response.py)

PR #447, opened 2025-12-15
[Issue #441] Check for missing releases based on tag

### LINKED ISSUE #441: [CI/CD] Set up automated check for tags without release
### Summary

Set up a GitHub action to check if there's a tag without a corresponding release.

### Acceptance criteria

- [ ] Unreleased tag linter runs at least once per day
- [ ] It fails if there are any tags without a corresponding release
- [ ] Confirm who gets a GitHub notification when the action fails

jcrichlake:
### Summary

- Fixes #441 
- Time to review: [10] minutes

Adding GitHub workflow that runs every daily to compare tags against releases in the repo and to fail if any tags or releases are missing an accompanying object.

I have it set to run daily at 6PM UTC as well as a manual trigger to quickly validate any fixes that we made instead of having to wait 24 hours.  I also limited the tag and release retrieval to be for the last seven days so that the v0.2.3 doesn't cause this to fail forever.


.github/workflows/missing-release-check.yml
```diff
@@ -0,0 +1,101 @@
...
+    - cron: '0 18 * * *'
+  workflow_dispatch: # Allows manual trigger
+
+jobs:
+  check-tags-releases:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
widal001: I think [`checkout@v6`](https://github.com/actions/checkout) is the latest version of this action.

.github/workflows/missing-release-check.yml
```diff
@@ -0,0 +1,101 @@
...
+        with:
+          fetch-depth: 0  # Fetch all tags and history
+      
+      - name: Check tags against releases
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          #!/bin/bash
+          set -e
```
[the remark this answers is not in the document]
widal001: You can put it in [this directory](https://github.com/HHS/simpler-grants-protocol/tree/main/.github/scripts)

.github/scripts/check-for-unreleased-tags.sh
```diff
@@ -0,0 +1,81 @@
...
+set -e
+
+echo "=== Checking Git tags against GitHub releases (last 7 days) ==="
+echo ""
+
+# Calculate date 7 days ago
+SEVEN_DAYS_AGO=$(date -u -d '7 days ago' '+%Y-%m-%d %H:%M:%S')
+SEVEN_DAYS_AGO_UNIX=$(date -u -d '7 days ago' '+%s')
+echo "Checking tags and releases created since: $SEVEN_DAYS_AGO UTC"
+echo ""
```
widal001: Thanks for pulling this out to its own script! But it 7 days is still a hard-coded value.

Per my earlier review, can we make this a configurable value and set the default in the CI workflow to 30 days?

widal001 (commented):
Working locally now! But there's still a mismatch between the script and how it's invoked in the GitHub action and the example provided in the script docstring.

Image

widal001 (approved):
I made [another change](https://github.com/HHS/simpler-grants-protocol/pull/447/commits/2df5be94752577d04d8f82f0f806bddace8c4378) to the workflow file to:

- Address one of my outstanding comments about the mismatch between the script and how it was being invoked in the GitHub action workflow
- Set up the workflow to run on PRs so we can test changes to it before merging
- Updated the workflow dispatch input to expect a number instead of a string
- Simplified the script invocation a bit to leverage some of GitHub's `||` operator for PR and scheduled runs

I think we should be good to merge now!

PR #485, opened 2026-02-02
465 Add get_custom_fields_value function

### LINKED ISSUE #465: [Py SDK] Implement `get_custom_field_value()`
### Summary

Add a typed accessor method on models that allows SDK users to retrieve and validate a custom field value by key without relying on dynamically generated attribute typing.

### Example

```python
from pydantic import BaseModel
from common_grants_sdk.schemas import OpportunityBase

class LegacyIdValue(BaseModel):
    system: str
    id: int

opp = OpportunityBase.model_validate({
    "id": "573525f2-8e15-4405-83fb-e6523511d893",
    "title": "Foo bar",
    "customFields": {
        "legacyId": {"type": "object", "value": {"system": "legacy", "id": 123}},
    },
})

legacy = opp.get_custom_field_value("legacyId", LegacyIdValue)
# legacy is Optional[LegacyIdValue]
print(legacy.id)
```

### Acceptance criteria

- [ ] Implement `get_custom_field_value(key: str, value_type: type[T]) -> Optional[T]` with generics
- [ ] Works for `value_type` that is a Pydantic BaseModel subclass and for primitive types (`int`, `str`, etc.)
- [ ] Returns `None` if the key is not present
- [ ] Raises a clear validation error if the value is present but invalid for `value_type`
- [ ] Works regardless of whether custom fields were registered via `with_custom_fields()`

jcrichlake:
### Summary

- Fixes #465 
- Time to review: 6 minutes

This PR adds a utility function named `get_custom_field_value` that can be added to any pydantic model via a wrapper function for retrieving the custom_fields.  As a demonstration of this functionality a wrapper function was add to the `OpportunityBase` Pydantic model.

#### Testing Instructions
1. Check out branch
2. cd to `lib/python-sdk/examples`
3. execute `poetry run python ./get_custom_fields.py` to see the print statements that are retrieving custom fields form the object. 

#### Additional Context
Part of the requirements of this PR included retrieving custom fields that were not already registered to a given Pydantic model via the `with_custom_fields` method.  This is why the example python script includes fields added without the registration call. 

There is also a set of commented lines in the example that can be uncommented to get an example of what the error message displays as in the terminal.

Example script
image

Output (with output for correctly retrieved fields followed by an example error message)
image

### DIFF
--- a/lib/python-sdk/tests/utils/test_custom_fields.py
+++ b/lib/python-sdk/tests/utils/test_custom_fields.py
@@ -1,7 +1,6 @@
 """Tests for custom fields functionality"""
 
 from datetime import datetime
-from pickle import GET
 from uuid import uuid4
 import pytest
 


lib/python-sdk/tests/utils/test_custom_fields.py
```diff
@@ -300,3 +300,189 @@ class ComplexSchema(BaseModel):
     opp = Opportunity.model_validate(complex_input_data)
 
     assert opp.custom_fields.complex_input_data.value.other_ids == [1, 2, 3, 4]
+
+
+####
+# Begin tests for custom field getter functions
+##
```
widal001: Small but non-blocking request: Can we re-organize this test file in a few ways:

- Replace the fixtures with constants, works better with intellisense, and consolidates them up to the top of the file.
- Re-use the same base record and then simply spread the custom fields into the other constants. Something like this should work
  ```python
  BASE_OPP = {
        "id": uuid4(),
        "title": "Foo bar",
        "status": OppStatus(value=OppStatusOptions.OPEN),
        "description": "Example opportunity",
        "createdAt": datetime.fromisoformat("2024-01-01T00:00:00+00:00"),
        "lastModifiedAt": datetime.fromisoformat("2024-01-01T00:00:00+00:00"),
  }
  ARRAY_INPUT = {
    **BASE_OPP
    "customFields": {
            "legacyId": {
                "name": "legacyId",
                "fieldType": "integer",
                "value": 12345,
            },
            "metadata": {
                "name": "metadata",
                "fieldType": "array",
                "value": [9, 8, 7, 6],
            },
            "ignoredForNow": {"type": "string", "value": "noop"},
        },
  }
  ```
- Group the tests for `with_custom_fields()` into their own test class (e.g. `TestWithCustomFields`) and do the same for `get_custom_field_value()`

PR #535, opened 2026-02-26
[Issue #499] Add Custom Fields to Opportunity Client

### LINKED ISSUE #499: [Py SDK] Client Support Custom fields
### Summary

The Opportunity Client in py-sdk needs expanded to support parsing and returning the registered custom fields if they are included in the API call.

We should update each of the endpoints to take in an optional opportunity base object that would enable sdk users to call the rest endpoints with the custom fields they wish to retrieve out of the response. 

### Notes
Below is a potential example of what this would look like

``` python
# Update method signature
    def search(
        self,
        search: str,
        status: List[OppStatusOptions],
        page: int | None = None,
        page_size: int | None = None,
       opportunity: OpportunityBase | None = None
    ) -> OpportunitiesSearchResponse:


        # Call client method to get paginated response
        paginated_response: Paginated[ItemsT] = self.client.search(  # type: ignore[valid-type]
            f"{self.path}/search",
            request_data.model_dump(by_alias=True, exclude_unset=True),
            page=page,
            page_size=page_size,
        )
    

    # Update
      if opportunity is None:
        opportunity = OpportunityBase

       opportunity.model_validate_json(json.dumps(item))

  

        # Convert paginated_response to dict and replace items with hydrated models
        response_data = paginated_response.model_dump(by_alias=True)
        response_data["items"] = items

        response_data["sortInfo"] = {
            "sortBy": "",
            "sortOrder": "",
            "customSortBy": None,
            "errros": [],
        }

        response_data["filterInfo"] = {"filters": {}, "errors": []}

        # Hydrate OpportunitiesListResponse from response data
        return OpportunitiesSearchResponse.model_validate(response_data)

```

### Acceptance criteria

- [ ] The 3 Opportunity Client ReST endpoints are updated with an optional OpportunityBase object
- [ ] The endpoints fallback to a default OpportunityBase object 
- [ ] The Opportunity Client examples are updated with examples that register, pass in, and retrieve the custom field values

jcrichlake:
### Summary

- Fixes #499 
- Time to review: 10 minutes

- Updates each of the methods in `Client.opportunities` to accept an optional `OpportunityBase` parameter that can be used to parse the API responses that include custom fields. 
- Changes the expected input for `with_custom_fields()` from `list[CustomFieldSpec]` to `dict[str, CustomFieldSpec]` to more closely match what is actually being added to the Opportunity. 
- A small change to the mock server in the `ts-sdk` to not return a page size of 0 if no results are found 
- Fixed a small bug for the base client where we were passing a param as data and not as json as the `httpx` library expected.  

1. Checkout the branch locally: `git checkout 499-client-custom-fields`
2. Change directory to the typescript sdk to `cd lib/ts-sdk`
3. Run `pnpm example:server` to spin up the local mock server
4. In a different terminal window cd to `lib/python-sdk/examples`
5. Update `search_opportunities.py` to use port 8000 instead of 8080
6. Run `poetry run python search_opportunities.py education`

image

image

image

### DIFF
--- a/code.json
+++ b/code.json
@@ -23,9 +23,9 @@
   "disclaimerURL": "",
   "disclaimerText": "",
   "vcs": "git",
-  "laborHours": 13411,
+  "laborHours": 13809,
   "reuseFrequency": {
-    "forks": 6,
+    "forks": 8,
     "clones": 0
   },
   "platforms": [],
@@ -40,8 +40,8 @@
     "JavaScript",
     "CSS",
     "Shell",
-    "Roff",
-    "Makefile"
+    "Makefile",
+    "Roff"
   ],
   "maintenance": "contract",
   "contractNumber": [],
@@ -51,8 +51,8 @@
   "partners": [],
   "date": {
     "created": "2024-12-19T21:14:35Z",
-    "lastModified": "2026-01-27T15:00:45Z",
-    "metadataLastUpdated": "2026-02-01T00:03:36.854Z"
+    "lastModified": "2026-02-24T15:32:14Z",
+    "metadataLastUpdated": "2026-03-01T00:03:50.345Z"
   },
   "tags": [
     "api",
--- a/examples/ca-opportunity-example/src/common_grants/routes/opportunities.py
+++ b/examples/ca-opportunity-example/src/common_grants/routes/opportunities.py
@@ -86,6 +86,7 @@ async def search_opportunities(
     opportunity_service = OpportunityService()
 
     return await opportunity_service.search_opportunities(
+        search=request.search,
         filters=request.filters,
         sorting=request.sorting,
         pagination=request.pagination,
--- a/examples/ca-opportunity-example/src/common_grants/services/opportunity.py
+++ b/examples/ca-opportunity-example/src/common_grants/services/opportunity.py
@@ -3,6 +3,7 @@
 from uuid import UUID
 
 from common_grants_sdk.schemas.pydantic import (
+    ArrayOperator,
     FilterInfo,
     OppFilters,
     OpportunitiesListResponse,
@@ -100,6 +101,23 @@ class OpportunityService:
                 if search_lower in opp.title.lower()
                 or (opp.description and search_lower in opp.description.lower())
             ]
+
+        # Apply status filter
+        if filters.status is not None and filters.status.value:
+            allowed_statuses = set(filters.status.value)
+            match filters.status.operator:
+                case ArrayOperator.IN:
+                    filtered_opportunities = [
+                        opp
+                        for opp in filtered_opportunities
+                        if opp.status.value in allowed_statuses
+                    ]
+                case ArrayOperator.NOT_IN:
+                    filtered_opportunities = [
+                        opp
+                        for opp in filtered_opportunities
+                        if opp.status.value not in allowed_statuses
+                    ]
 
         # Create PaginatedResultsInfo object
         start_idx = (pagination.page - 1) * pagination.page_size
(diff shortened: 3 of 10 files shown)


lib/python-sdk/common_grants_sdk/client/opportunity.py
```diff
@@ -37,10 +37,29 @@ def path(self) -> str:
...
+        if opp_base is None:
+            return OpportunityBase
+        else:
+            return opp_base
+
     def list(
         self,
         page: int | None = None,
         page_size: int | None = None,
+        opp_base: Type[OpportunityBase] | None = None,
```
widal001: I'm pretty sure you can avoid creating and using this static method if you change the function signature to this:

```python
opp_base: Type[OpportunityBase] = OpportunityBase
```

That way you don't have to check if it's `None` since either folks provide an `opp_base` directly or it defaults to `OpportunityBase`

widal001: Also we should probably align the name of this optional param across the Python and TypeScript SDKs.

Currently it's called `schema` in the [TS SDK](https://github.com/HHS/simpler-grants-protocol/blob/1ebfcf9499a1b50ed22a8ad57043af75b485d8ed/lib/ts-sdk/src/client/opportunities.ts#L104). 

I'm open to making an associated change in the TS SDK, but I chose `schema` for a few reasons:

- SDK users are most likely to use this field when they are creating a _new_ schema using `withCustomFields()` whose name _shouldn't_ be `OpportunityBase`
- We _could_ call it `oppSchema` (`opp_schema` in Python) to indicate that the schema passed needs to be a sub-type of `OpportunityBase`
- But I anticipate doing something similar for future client resource methods like `Client.awards.list()` and I was hoping to provide the same optional attribute across all methods that support this pattern rather than having different ones for each resource.

But I don't feel suuuper strongly about the param name we decide on as long as we're consistent across SDKs

lib/python-sdk/tests/client/test_opportunities.py
```diff
@@ -249,13 +262,23 @@ def test_list_opportunities_success(
...
-        response = client.opportunity.list(page=1)
+        fields = {
+            "legacyId": CustomFieldSpec(field_type=CustomFieldType.INTEGER, value=int),
+        }
+
+        opp_base = OpportunityBase.with_custom_fields(
+            custom_fields=fields, model_name="Opportuntiy"
+        )
+
+        response = client.opportunity.list(page=1, opp_base=opp_base)
```
widal001: One other major discrepancy I'm just now realizing between the Python and the TypeScript SDK is that the python SDK uses the singular `Client.opportunity.list()` where as TypeScript uses `Client.opportunities.list()`

It looks like plural resource names are more common and align with the API paths themselves (e.g. `GET /common-grants/opportunities/`) so can we change them in the Python SDK to make them consistent with TypeScript?

We can do that in a separate PR but it should happen before we publish the v0.5.0 of the Python SDK.

Image

widal001 (commented):
Thanks for working on this! I outlined a few key changes I'd like to get into this PR and one big change for a follow-up PR to make sure the TypeScript and Python SDKs are aligned.

**This PR**

- Adding `name` and `description` to `CustomFieldSpec` and using them to set the default value for those attributes in the `CustomField` that gets created by `with_custom_fields()`
- Updating the name of the `opp_base` param. If we want to set it to something other than `schema` I can make a corresponding change in the TypeScript SDK.
- Fixing how we're validating that the new param was actually used to parse values in the unit tests.

**Future PR:** Updating `Client.opportunity` --> `Client.opportunities` to match:
- The TypeScript SDK
- The API methods themselves (e.g. `GET /common-grants/opportunities/`)
- Other popular SDKs like [the Twilio SDK](https://github.com/twilio/twilio-python?tab=readme-ov-file#use-the-list-method)

widal001 (approved):
Looks good! Thanks for making these changes. I noted a set of small formatting issues with the examples README, but otherwise this PR looks good to merge.

Feel free to tackle those formatting issues in the PR that changes `Client.opportunity` --> `Client.opportunities`

PR #668, opened 2026-04-09
[Issue #649] Adding Plugins to Common Grants

### LINKED ISSUE #649: [Plugin] List plugins on CommonGrants.org
### Summary

Add the published plugins to CommonGrants.org to complete this final metric on our enhanced deliverable.

### Notes

We have two potential options:

1. Adding a section to the custom field details page that lists which package(s) that field is implemented in.
2. Updating the "Using python" and "Using TypeScript" guides to reference these packages.

Personally I would lead toward the first, but we can chat about it further!

### Acceptance criteria

- [ ] The Python and TypeScript published plugins are listed on the CommonGrants.org website.

jcrichlake:
### Summary

- Fixes #649 
- Time to review: 10 minutes

This PR adds a plugin catalog page to the CommonGrants website and introduces bidirectional cross-linking between plugins and the custom fields they implement.

  - **Plugin catalog page** — new `/plugins` index page listing all plugins from `content/plugins/index.json`, with filtering by language and system
  - **Plugin detail page** — new `/plugins/[slug]` page rendered via `PluginDetails.astro`, showing plugin metadata and the custom fields it exposes as `CustomFieldCard` components
  linking to each field's detail page
  - **`CustomFieldDetails.astro`** —Added a "Used By Plugins" section that filters `plugins/index.json` for plugins containing the
  current field and renders each as a `PluginCatalogCard`
  - **`PluginDetails.astro`** — Added a `CustomFieldCard` grid (mirrors the pattern above) 
  
  

The cross-linking works as follows:
  - On a custom field detail page (e.g. `/custom-fields/agency`), a "Used By Plugins" section appears only when at least one plugin includes that field. Each card links to
  `/plugins/[slug]`.
  - On a plugin detail page (e.g. `/plugins/ts-cg-grants-gov`), the custom fields section renders `CustomFieldCard` components, each linking to `/custom-fields/[id]`.

  **Verification:**
  1. Navigate to `/custom-fields/agency` — confirm a "Used By Plugins" section appears with cards for both `ts-cg-grants-gov` and `py-cg-grants-gov`
  2. Click a plugin card — confirm it routes to the correct `/plugins/[slug]` detail page
  3. On the plugin detail page — confirm custom field cards appear and each links to `/custom-fields/[id]`
  4. Navigate to `/custom-fields/eligibilityCriteria` (not used by either plugin) — confirm no "Used By Plugins" section appears
  5. Navigate to `/plugins` — confirm the plugin catalog index renders correctly


SnowboardTechie (approved):
LGTM, played around with local website plugin views look sharp and the filtering works well even if there isn't much to filter yet 😄

website/src/content/plugins/index.json
```diff
@@ -0,0 +1,50 @@
+{
+  "ts-cg-grants-gov": {
+    "label": "TypeScript Common-Grants Plugin",
+    "description": "A TypeScript adapter that maps Grants.gov opportunity data to the CommonGrants protocol.",
+    "system": "Grants.gov",
+    "url": "https://github.com/common-grants/ts-cg-grants-gov",
+    "language": "TypeScript",
+    "version": "0.1.0",
```
widal001: Rather than requiring maintainers to manually update the version, and specify the language etc., can we include the npm/PyPI package url and then update the loader to fetch:
  - The language
  - The repo URL
  - The version

An alternative might be do it in the reverse direction, i.e., infer these things from the GitHub repo.

Eventually we could even ask maintainers of these plugins to manage this metadata in their own repo e.g. in a standard `cg.metadata.json` (or something similar) file, so they can update stuff directly in place.

website/src/pages/plugins/[slug].astro
```diff
@@ -0,0 +1,36 @@
...
+---
+
+<StarlightPage
+  frontmatter={{
+    title: plugin.label,
+    description: plugin.description,
+    tableOfContents: false,
+  }}
+>
+  <PluginDetails plugin={plugin} />
```
Comment (widal001) [inline website/src/pages/plugins/[slug].astro:36]: Can we add a back arrow here like we do for [question bank](https://github.com/HHS/simpler-grants-protocol/blob/1f526c994da0b0d1efc1ce529fe5fe0676fb0618/website/src/pages/question-bank/%5Bslug%5D.astro#L39)?

Comment (widal001) [reply website/src/pages/plugins/[slug].astro:36]: It would be great if we could do the same for the custom fields page, even though I know that wasn't touched in this PR.

website/src/content/plugins/index.json
```diff
@@ -0,0 +1,50 @@
+{
+  "ts-cg-grants-gov": {
+    "label": "TypeScript Common-Grants Plugin",
```
widal001: Can we change this to match the name of the actual plugin? Or add a section that has that name?

So `@common-grants/cg-grants-gov` for the TypeScript plugin?

website/src/components/plugins/PluginDetails.astro
```diff
@@ -0,0 +1,173 @@
...
+
+  .repo-link:hover {
+    text-decoration: underline;
+  }
+
+  .fields-grid {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 1rem;
+  }
```
widal001: Same issues with this grid as the ones we saw with the plugin grid on the custom field detail page.

Image

website/src/components/plugins/PluginDetails.astro
```diff
@@ -0,0 +1,173 @@
...
+        <dt>Repository</dt>
+        <dd>
+          <a
+            href={plugin.url}
+            target="_blank"
+            rel="noopener noreferrer"
+            class="repo-link"
+          >
+            {plugin.url}
+            <span aria-hidden="true">↗</span>
```
widal001, line 54: Can we shorten the text of this link to just be `github.com/<org>/<repo>` or maybe even just `<org>/<repo>`

So either:
- `github.com/common-grants/ts-cg-grants-gov`
- `common-grants/ts-cg-grants-gov`

And also can we add an aria/screen reader only label that says "GitHub repo for this plugin" since screen readers will likely just spell out the name if it's all snake case.

website/src/pages/plugins/index.astro
widal001, line 1: We'll also want to link this page from [the navbar](https://github.com/HHS/simpler-grants-protocol/blob/1f526c994da0b0d1efc1ce529fe5fe0676fb0618/website/src/components/starlight-overrides/Header.astro#L37-L46) and [the sidebar](https://github.com/HHS/simpler-grants-protocol/blob/1f526c994da0b0d1efc1ce529fe5fe0676fb0618/website/astro.config.mjs#L94-L108), probably under custom fields in the "Protocol" section

website/src/components/plugins/PluginDetails.astro
jcrichlake: I'm not sure what you're referring to here. I checked the `.fields-grid` properties and they appear to match.  Is there a comment you meant to add that maybe didn't go through?

website/src/content/plugins/index.json
widal001: Sorry my comment meant to suggest that we make the `label` match the package name, or add a new attribute for package name, rather than changing the slug.

The reason we don't want to make the slug match the package name is that there might be name collisions between Python and TypeScript plugins, especially for third party TypeScript plugins that won't be prefixed with `@common-grants/` since that's reserved for packages published to our npm org.

widal001: Also using `@` and `/` symbols in the url can break the links (which is what happens with this package)

Image

website/src/content/plugins/index.json
```diff
@@ -0,0 +1,47 @@
+{
+  "@common-grants/cg-grants-gov": {
+    "label": "TypeScript Common-Grants Plugin",
+    "description": "A TypeScript adapter that maps Grants.gov opportunity data to the CommonGrants protocol.",
+    "system": "Grants.gov",
+    "packageUrl": "https://www.npmjs.com/package/@common-grants/cg-grants-gov",
```
widal001, line 7: Can we actually add this to the details page? This could be the href for the field that has the package name:

```jsx
<a href={packageUrl}> {packageName} </a>
```

website/src/content/plugins/index.json
```diff
@@ -0,0 +1,47 @@
...
+    },
+    "createdAt": "2026-01-20",
+    "lastModifiedAt": "2026-04-09"
+  },
+  "cg-grants-gov": {
+    "label": "Python Common-Grants Plugin",
+    "description": "A Python adapter that maps Grants.gov opportunity data to the CommonGrants protocol.",
+    "system": "Grants.gov",
+    "packageUrl": "https://pypi.org/project/cg-grants-gov/",
+    "repoUrl": "https://github.com/common-grants/py-cg-grants-gov",
```
widal001, line 29: PyPI allows you to [set a repo URL](https://docs.pypi.org/project_metadata/) for your package like npm does, but this might to be updated on the Python package itself.

website/src/content/plugins/index.json
```diff
@@ -0,0 +1,47 @@
...
+        "attachments",
+        "federalFundingSource",
+        "contactInfo",
+        "additionalInfo",
+        "fiscalYear",
+        "costSharing"
+      ]
+    },
+    "createdAt": "2026-01-20",
+    "lastModifiedAt": "2026-04-09"
```
widal001: Not a big deal, but since we don't seem to be presenting these in the details page and we can likely get the last published date from the package registry, might suggest leaving these out to avoid adding overhead for what needs to be updated and to avoid them becoming inaccurate.

website/src/content/plugins/index.json
jcrichlake: It does need to be on the package itself, I can log an issue on the plugin for that

widal001 (approved):
Woohoo!! This is a really exciting addition 🚀

PR #698, opened 2026-04-09
fix(ci): handle missing FIDER_TOKEN in coplanning sync dry-run

SnowboardTechie:
## Summary
- Skip the GitHub-to-Fider dry-run step when `FIDER_API_TOKEN` is empty
- Leave the workflow_dispatch real-run path unchanged

## Verification
- YAML parsed successfully
- Workflow diff only changes the dry-run step guard

### DIFF
--- a/.github/workflows/coplanning-sync-gh-to-fider.yml
+++ b/.github/workflows/coplanning-sync-gh-to-fider.yml
@@ -33,7 +33,7 @@ jobs:
           python-version: "3.13"
 
       - name: Run script (dry run)
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && env.FIDER_API_TOKEN != ''
         run: |
           python run.py \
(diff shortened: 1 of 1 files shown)


.github/workflows/coplanning-sync-gh-to-fider.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           python-version: "3.13"
 
       - name: Run script (dry run)
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && env.FIDER_API_TOKEN != ''
```
widal001, line 36: If this token ever gets accidentally deleted this would cause us to only run this in dry-run mode. And make it look like it's silently passing without doing anything.

If this change is related to a dependabot bump then we should follow the same pattern we did for other ones and skip the job if it's a dependabot PR.

PR #711, opened 2026-04-13
[Stack 2/4] feat(website): Add lib/forms loader with composition + x-overrides

widal001:
### Summary

Sets up a new forms pipeline that mirrors `src/content/question-bank/` and `src/content/custom-fields/`.

- Part 2 of 4 in a stacked diff
- Time to review: 15 minutes

> [!NOTE]
> The ultimate goal is to switch from hand-maintained [JSON schema forms](https://github.com/HHS/simpler-grants-protocol/tree/main/website/src/content/forms) to ones generated from the question bank. But this needs to be done in steps, because some of the forms (like the SF424 and its variants) are quite complex.

### Changes proposed

- Adds a new `Forms` namespace at `website/src/specs/forms/` that compiles through the existing `pnpm typespec:extensions` pipeline alongside `question-bank` and `custom-fields`
- Adds a placeholder `FormCanary` model + `website/src/content/forms/typespec-index.json` so the loader has a real schema to test against (the canary is removed in #713 once the first real is added)
- Added `website/src/lib/forms/` (`types.ts`, `loader.ts`, `overrides.ts`, `compose.ts`, `index.ts`) following the same pattern as `lib/question-bank/` and `lib/custom-fields/`
- The loader composes a form-level `x-ui-schema` and `x-mapping-*` from each form's nested QB property references, then merges per-form `x-overrides` to allow us to change the name or label of individual form fields without re-defining the entire UI schema or mapping.
- Adds a contributor-facing README at `website/src/specs/forms/README.md` documenting the:
  - spec authoring conventions
  - the field-level vs model-level override pattern
  - the backtick rule for dotted keys
  - the composition behavior (how inherited UI schemas and mappings are combined with overrides)
- Adds 47 unit + integration tests covering loader, overrides, and composition

### Context for reviewers

**Review instructions:**
1. Checkout the PR locally
2. Change directory to `website/` 
3. Run `pnpm typespec` and see that `.extensions/FormsCanary.yaml` was created
4. Run `pnpm test` and all tests should pass
5. Run `pnpm dev` and the current `/forms/` page should remain unchanged.
6. Browse the pages at: https://cg-pr-711.billy-daly.workers.dev/forms/

**Implementation notes:**
- Field-level overrides are the preferred authoring shape: declare `@extension("x-overrides", #{ uiSchema: #{ name: #{ label: "..." } } })` on the property whose composed type is being patched. The loader aggregates these alongside any model-level `x-overrides`, prepending the property name to each path.
- Composition only fires when the form spec declares no explicit form-level `x-ui-schema` / `x-mapping-*`. Forms that declare their own (like the canary) opt out and use the explicit value verbatim.
- Strict failure mode: a typoed override path throws at load time rather than silently no-op'ing.
- The loader follows the established `lib/question-bank/` and `lib/custom-fields/` conventions (module-scope cache, `extractSchemaData` style, `dereferenceSchema`) so anyone familiar with those will read this without surprise.

### Additional information

Here's a screenshot of the canary form JSON schema generated from this TypeSpec:

Screenshot 2026-04-14 at 8 28 15 AM

Screenshot 2026-04-14 at 8 27 03 AM

### Stack

> This PR is part of a stacked diff. Please review only the changes in this PR, not the cumulative diff.

| # | Branch | Targets | PR | Status |
|---|--------|---------|----|--------|
| 1 | `stack/593-generate-forms-from-question-bank/01-audit-cleanup` | `593-generate-forms-from-question-bank` | #710 | ⏳ Open |
| 2 | `stack/593-generate-forms-from-question-bank/02-forms-loader` | PR 1 | #711 | 🎯 Current |
| 3 | `stack/593-generate-forms-from-question-bank/03-forms-new-pages` | PR 2 | #712 | ⏳ Queued |
| 4 | `stack/593-generate-forms-from-question-bank/04-key-contact-form` | PR 3 | #713 | ⏳ Queued |

**Review order**: Start from PR 1 and work down.
**Merge strategy**: Squash and merge each PR in order. After merging, rebase the next PR onto the parent feature branch (`593-generate-forms-from-question-bank`) and update its base accordingly. Once all four merge into the parent, an umbrella PR will merge `593-generate-forms-from-question-bank` into `main`.

### DIFF
--- a/website/src/lib/forms/compose.ts
+++ b/website/src/lib/forms/compose.ts
@@ -10,16 +10,7 @@
  * composition (like the canary) keep working unchanged.
  */
 
-// =============================================================================
-// SHARED TYPES
-// =============================================================================
-
-/** A node in a JSON-Forms UI schema (Control or Layout). */
-type UiNode = Record<string, unknown> & {
-  type?: string;
-  scope?: string;
-  elements?: UiNode[];
-};
+import type { UiNode } from "./types";
 
 // =============================================================================
 // UI SCHEMA COMPOSITION
--- a/website/src/lib/forms/compose.ts
+++ b/website/src/lib/forms/compose.ts
@@ -102,9 +102,24 @@ function deepMergeInto(
  * Synthesizes a form-level `x-mapping-from-cg` by nesting each property's
  * own `x-mapping-from-cg` under its property name.
  *
- * The form-side keys in the child mapping are relative to the child's
- * structure; nesting them under `<propName>` makes them relative to the
- * parent form's structure. The CG-side leaf values stay unchanged.
+ * Each child mapping's keys describe paths relative to that child's
+ * structure. Nesting them under the property name makes them relative to
+ * the parent form.
+ *
+ * The field entry values are CommonGrants data-model paths (e.g.
+ * `organizations.primary.name`) and stay unchanged, since they always
+ * reference the same place in the CG model regardless of where the
+ * question sits in the form.
+ *
+ * @example
+ * // Given a form with `org: QuestionOrgName` whose x-mapping-from-cg is:
+ * //   { name: { field: "organizations.primary.name" } }
+ * //
+ * // composeMappingFromCg produces:
+ * //   { org: { name: { field: "organizations.primary.name" } } }
+ * //
+ * // The form-side key changed (name -> org.name) but the CG-side
+ * // value ("organizations.primary.name") stayed the same.
  */
 export function composeMappingFromCg(
   properties: Record<string, unknown>,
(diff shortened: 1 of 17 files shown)


website/src/lib/forms/compose.ts
```diff
@@ -0,0 +1,192 @@
...
+ */
+function deepMergeInto(
+  target: Record<string, unknown>,
+  source: Record<string, unknown>,
+): void {
+  for (const [key, value] of Object.entries(source)) {
+    const existing = target[key];
+    if (
+      typeof value === "object" &&
+      value !== null &&
```
jcrichlake, line 84: Would it be better to re-arrange these checks to have the null checks for value and existing  be first? I'm thinking having the most likely to short circuit clauses first so that we run less checks would be better

website/src/lib/forms/compose.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+}
+
+/**
+ * Synthesizes a form-level `x-mapping-from-cg` by nesting each property's
+ * own `x-mapping-from-cg` under its property name.
+ *
+ * The form-side keys in the child mapping are relative to the child's
+ * structure; nesting them under `<propName>` makes them relative to the
```
jcrichlake: This doc string is hard to follow, could you add some concrete examples? Also what is meant by CG-side leaf values?

website/src/lib/forms/overrides.ts
```diff
@@ -0,0 +1,141 @@
...
+  const segments = dottedPath.split(".");
+  return "#/" + segments.map((s) => `properties/${s}`).join("/");
+}
+
+// =============================================================================
+// UI SCHEMA OVERRIDE MERGE
+// =============================================================================
+
+/** A node in a JSON-Forms UI schema (Control or Layout). */
+type UiNode = Record<string, unknown> & {
```
jcrichlake: Isn't this a duplicate type? Can we extract it out so it's only in one place?

jcrichlake:
Overall this looks really good!  I was wondering though if we could potentially meet up to discuss some of the questions I have on this. @bryan-thompsoncodes  you're welcome to join as well!  My two main areas of concern are that I'm struggling a bit to get my head around the big picture pieces of this new piece of functionality (eg. we have loader vs overrides vs compose) and I'm wondering about the note that says this mirrors custom fields and the question bank modules. Could we perhaps refactor / generalize these 3 different libraries into a core piece that all 3 extend instead of having a lot of similar code in three places? 

@widal001 for awareness

website/src/lib/forms/overrides.ts
widal001: Good call! I moved it to `forms/types.ts` in this commit: [refactor(website): Move UiNode to shared types.ts](https://github.com/HHS/simpler-grants-protocol/pull/711/commits/2d3ca21d237a68dbeb6ff328ee6ceec64eb21061)

website/src/lib/forms/compose.ts
widal001: I updated these docstrings and added a section to the README in this commit: [docs(website): Improves docs for form loading](https://github.com/HHS/simpler-grants-protocol/pull/711/commits/1987bd1eb25028bc7f70d51f60efb2983bf89f5d)

Let me know if that helps!

website/src/lib/forms/compose.ts
widal001: Sure! I reordered them in this commit: [refactor(website): Reorder checks in compose.ts](https://github.com/HHS/simpler-grants-protocol/pull/711/commits/60f610625e4f70ecc0903082f5b7714f062d9fa1)

jcrichlake (approved):

jcrichlake:
Thanks for adding those clarifications! I especially like what you added to the readme!  I think that will be extremely helpful in the future if/when we revisit this code!

jcrichlake (approved):

SnowboardTechie (approved):

PR #729, opened 2026-04-15
fix: Replace pnpm audit with bulk advisory endpoint script

SnowboardTechie:
### Summary

- Fixes CI failures caused by npm retiring their legacy audit endpoint
- Time to review: 10 minutes

### Changes proposed

Added a new script (`.github/scripts/audit-deps.js`) that calls npm's replacement bulk advisory endpoint directly, bypassing pnpm's broken `pnpm audit` command. All CI workflow audit steps now use this script instead of `pnpm audit`.

**Files changed:**
- New: `.github/scripts/audit-deps.js` — standalone Node.js script that queries the bulk advisory endpoint
- 8 CI workflow files updated to call the new script instead of `pnpm audit`
- 5 package.json files reverted (no longer need `--ignore-registry-errors`)

### Context for reviewers

npm retired their legacy `/v1/security/audits` endpoint on 2026-04-15, returning HTTP 410. pnpm (all v10.x and v11 rc) still uses the old endpoint, breaking `pnpm audit` globally.

This is tracked upstream:
- **Issue:** https://github.com/pnpm/pnpm/issues/11265

Rather than using `--ignore-registry-errors` (which silently skips the audit), this PR adopts a community workaround from the upstream issue ([comment](https://github.com/pnpm/pnpm/issues/11265#issuecomment-4253931245)) that calls the new bulk advisory endpoint directly. This keeps real vulnerability detection active.

The script audits all dependencies (prod + dev) and supports a severity threshold argument matching the existing per-workflow audit levels.

Each workflow file has a comment linking to the upstream issue and documenting what to revert once pnpm ships native support (expected in pnpm v11).

### Additional information

**Dependabot PRs unblocked by this change:**
- #718 — chore(deps-dev): bump the tooling group (typescript-eslint)
- #719 — chore(deps): bump the website-framework group

### DIFF
--- a/.github/scripts/audit-deps.js
+++ b/.github/scripts/audit-deps.js
@@ -135,7 +135,15 @@ const req = https.request(
         process.exit(1);
       }
 
-      const advisories = JSON.parse(data);
+      let advisories;
+      try {
+        advisories = JSON.parse(data);
+      } catch {
+        console.error(
+          'Failed to parse registry response:\n' + data.slice(0, 200)
+        );
+        process.exit(1);
+      }
       const affected = Object.keys(advisories);
 
       if (affected.length === 0) {
--- a/.github/scripts/audit-deps.js
+++ b/.github/scripts/audit-deps.js
@@ -1,0 +1,200 @@
+#!/usr/bin/env node
+
+// Workaround for pnpm audit 410 error (pnpm/pnpm#11265).
+//
+// npm retired the legacy audit endpoints on 2026-04-15, returning
+// 410 Gone. pnpm <=10.x (and 11.0.0-rc.0) still uses the old
+// endpoints. This script calls the replacement bulk advisory
+// endpoint directly.
+//
+// Usage:
+//   node audit-deps.js [options]
+//
+// Options:
+//   --level <severity>    Minimum severity to fail on (low|moderate|high|critical). Default: low
+//   --filter <pattern>    pnpm workspace filter (e.g. @common-grants/cli)
+//   --ignore-ghsa <id>    GHSA ID to ignore (can be repeated)
+//
+// Requires pnpm on PATH. Remove this script once pnpm ships
+// native support for the new endpoint.
+
+'use strict';
+
+const { execSync } = require('node:child_process');
+const https = require('node:https');
+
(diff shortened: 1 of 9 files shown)


.github/scripts/audit-deps.js
```diff
@@ -0,0 +1,208 @@
...
+        console.error('Registry returned ' + res.statusCode + ': ' + data);
+        process.exit(1);
+      }
+
+      let advisories;
+      try {
+        advisories = JSON.parse(data);
+      } catch {
+        console.error(
+          'Failed to parse registry response:\n' + data.slice(0, 200)
```
jcrichlake, line 143: Do we want just 200 characters from the bad data or should we want all of the data that comes back to assist with debugging?

jcrichlake:
Thanks for being on top of this workaround! Overall it looks good and is easy to follow! I suggested one small change and I have one overall question that you don't have to adopt if you don't think it's worth it. We fairly recently went through the various project level build files (IE package.json or Makefile and updated them to make sure we had the commands that would be used by the CI pipelines readily available to devs so they could validate that they solved any errors occurring in the pipeline locally without having to guess and check.  Do you think this would be worth it to add or not? I'm on the fence since hopefully this is just a temporary measure.

.github/scripts/audit-deps.js
SnowboardTechie: Good question! The truncation is intentional here, this branch only fires when the registry returns 200 but non-JSON (e.g. a CDN error page). I figure 200 characters is enough to identify what comes back without risking an arbitrarily large malformed payload flooding CI logs. If someone needs the full response, they could curl the endpoint directly.

If that doesn't sit right with you though I'm not strong attached to this choice so I could go for a higher limit?

SnowboardTechie:
@jcrichlake I think in this case I'd prefer to minimize what needs to be removed when pnpm 11 comes around. Since it's a temporary workaround, I'd rather keep the new script scoped to CI and not widen the revert surface. But I appreciate the callout, 🤞🏻 hopefully this doesn't stick around long enough for me to regret not including those commands

jcrichlake (approved):

PR #737, opened 2026-04-16
[Issue #714] Fix AJV Error

### LINKED ISSUE #714: [Form library] Fix shared AJV multi-ref failure for composed forms
### Summary

Resolve the shared `extensionsAjv` initialization failure that blocks any TypeSpec form from composing multiple question-bank types whose transitive `$ref`s share a downstream schema (e.g. `QuestionAddress` reachable via `org` + `contact` + `aor` in sf424-mandatory).

When such a form's compiled YAML lives in `.extension-schemas/`, AJV's `_addSchema` → `getSchemaRefs` walks the `$ref` graph at module-load time and throws `"resolves to more than one schema"`. The throw fires from any page importing `validation.ts` (directly or transitively via `example-generator.ts` / `extensionsAjv`), which is most of the website. Tracked from the form-library stack PRs #710, #711, #712, #713.

### Context

- The form-library stack ships a forms-local per-call AJV pattern (`generateFormExample`) on branch `593-form-example-ajv-fix` (commit `dbd78bd`). It bypasses the shared AJV for example generation but does not address the module-load init failure.
- The `key-contact` form (the only one in PR #713) avoids the issue because it only references `QuestionAddress` once. Any further form that composes two or more `*-Details` composites (or any composite with a nested address / phone / etc.) will trip it.

### Implementation options

Three approaches, each with different blast radius. Pick the smallest that unblocks the rest of the form library:

1. **Per-call AJV inside the shared `generateSchemaExample`** (`website/src/lib/schema/example-generator.ts`). Fully replaces the use of `extensionsAjv` for example generation; the `validator` parameter becomes vestigial. Forward-compatible for QB and docs pages. Does not directly fix the init-time throw; needs to be paired with one of the next two.
2. **Move forms to a separate output directory** (e.g. `.form-schemas/`) and a new `pnpm typespec:forms` script. Forms never enter `.extension-schemas/`, so `extensionsAjv` is unaffected. Touches `tspconfig`, `package.json`, `Paths.ts`, `.gitignore`, the forms loader's path resolution, and the `extensions/index.tsp` import list.
3. **Guarded `addSchema` loop in `validation.ts`** that catches per-schema failures and skips with a warning. One-line change but allows future regressions to silently land in the warnings.

The forms-local pattern (`generateFormExample`) on branch `593-form-example-ajv-fix` is the per-call half of option 1.

### Acceptance criteria

- [ ] A TypeSpec form that composes at least two question-bank composites with shared downstream `$ref`s renders successfully on `/forms-new/<slug>`
- [ ] `pnpm build` and `pnpm test` succeed with such a form present in `.extension-schemas/` (or wherever forms are emitted after the fix)
- [ ] A regression test asserts that the loader produces a valid example for a multi-composite form

jcrichlake:
### Summary

  - Fixes #714
  - Time to review: 15 minutes

  ### Changes proposed

  **Step 1 — Per-call AJV in `generateSchemaExample`**
  - `generateSchemaExample` now creates a fresh AJV instance per-call from the schema's own directory, replacing the shared `extensionsAjv` passed as a parameter
  - `options.validator` parameter removed; `QuestionDetails.astro` and `FormDetails.astro` updated accordingly
  - `validateExample` now warns instead of throws when a schema isn't registered, so builds degrade gracefully

  **Step 2 — Forms to a separate output directory**
  - Forms now compile via `pnpm typespec:forms` into `.form-schemas/` instead of `.extension-schemas/`, so `extensionsAjv` never sees form schemas and can't throw at init time
  - Added `FORM_SCHEMAS_DIR` and `FORM_SCHEMAS_PATH_PREFIX` to `Paths.ts`
  - Added `formSchemaFilePath()` helper to catalog loader; forms loader updated to use it
  - Removed forms import from `extensions/index.tsp`
  - `.form-schemas/` added to `.gitignore` and `typespec:clear-output`

  **`sf424-mandatory` regression form**
  - New TypeSpec form composing `QuestionOrgDetails` + `QuestionPocDetails` + `QuestionAorDetails` — all three share `QuestionAddress` downstream, which is the exact multi-composite
  case that triggered the AJV failure
  - Registered as slug `sf424` → renders at `/forms-new/sf424`
  - Label: "SF-424 Mandatory"

  **`FormDetails.astro` — JsonForms `$id` fix**
  - After full `$ref` inlining, `$id: QuestionAddress.yaml` appears once per composite in the dereferenced schema tree. JsonForms internally calls AJV during SSR, which throws "resolves
   to more than one schema" when it encounters duplicate `$id`s
  - Schema is passed through a `JSON.stringify` replacer that strips all `$id` fields before being handed to `JsonFormRenderer`

  ### Context for reviewers

  **Root cause:** Two distinct AJV errors, both triggered by multi-composite forms sharing `QuestionAddress.yaml`:
  1. **Init-time throw** — `extensionsAjv` was initialized at module level with all schemas including forms. Loading `SF424Mandatory` caused AJV to throw before any page rendered. Fixed
   by step 2 (forms leave `.extension-schemas/`) paired with step 1 (no shared validator instance).
  2. **SSR throw in JsonForms** — After `$ref` inlining, duplicate `$id` values in the dereferenced schema cause AJV (used internally by JsonForms) to throw during Astro's static site
  generation. Fixed by stripping `$id` fields in `FormDetails.astro` before passing to the renderer.

  **Verification:** 
  
  1. Compile form schemas and verify sf424-mandatory lands in .form-schemas/
  cd website
  `pnpm typespec`
  `ls .form-schemas/SF424Mandatory.yaml`

  2. Full build (typespec + generate + astro build) — must exit 0
  `pnpm build`

  3. Run the full test suite — must exit 0
  `pnpm test`

  4. Run just the forms loader tests to confirm the regression test passes
  `pnpm test __tests__/lib/forms/loader.spec.ts`

  5. Spot-check the rendered page (starts dev server; navigate to /forms-new/sf424-mandatory)
  `pnpm dev`

Navigate to `localhost:XXX/forms-new` and visit the SF424 form page

  ### Additional information

  New test coverage in `__tests__/lib/forms/loader.spec.ts`:

  sf424 form (multi-composite with shared refs)
    ✓ loads without error (verifies no AJV shared-ref init failure)
    ✓ exposes org, contact, and aor properties
    ✓ composes a VerticalLayout with three top-level group elements

  loadAllFormItems
    ✓ returns a map keyed by form id  ← now also asserts sf424


website/src/lib/schema/example-generator.ts
```diff
@@ -40,12 +44,15 @@ export async function generateSchemaExample(
...
   example: string,
   schemaName: string,
-  ajvInstance: Ajv2020 = ajv,
+  ajvInstance: Ajv2020,
 ): void {
   // Validate the generated example against the source schema
   const validator = ajvInstance.getSchema(schemaName);
   if (!validator) {
-    throw new Error(`Schema ${schemaName} not found for ${schemaName}`);
+    console.warn(
```
SnowboardTechie: Two concerns with switching this from throw to console.warn:

This early return fires when a schema's $id doesn't match its filename, addSchema registers under $id, but the lookup here uses `path.basename(filePath)`. If those ever drift apart (e.g. a change to how TypeSpec emits $id), we'd lose example validation silently rather than getting a clear signal.

Combined with the missing example-generation test for sf424, there'd be no signal — runtime or CI to catch a future regression in example validation for multi-composite forms.
Would lean toward restoring the throw so drift surfaces immediately.

website/__tests__/lib/forms/loader.spec.ts
```diff
@@ -104,6 +106,40 @@ describe("forms loader", () => {
       });
     });
 
+    // =============================================================================
+    // sf424 form (multi-composite with shared refs)
+    // =============================================================================
+
+    describe("sf424 form (multi-composite with shared refs)", () => {
```
SnowboardTechie, line 115: Could we add a test that calls `generateSchemaExample` against the SF424Mandatory schema and asserts the result parses + validates without errors?

The current tests cover loading and structural composition, but example generation is the specific code path the shared-AJV bug was breaking and it's currently uncovered for the multi-composite case. This pairs with the comment I left on example-generator.ts: if `validateExample` keeps the silent-warn branch, this test would be the only safeguard against an example-generation regression slipping through unnoticed.

website/src/components/forms-new/FormDetails.astro
```diff
@@ -42,6 +39,17 @@ const hasOverrides =
...
   Object.keys(item.overrides.mappingToCg ?? {}).length > 0;
+
+// Strip all $id fields before passing to JsonForms. After full $ref
+// inlining, the same $id (e.g. QuestionAddress.yaml) can appear multiple
+// times in the schema tree — once per composite that references it. AJV
+// (used internally by JsonForms) throws "resolves to more than one schema"
+// when it encounters duplicate $id values.
+const rawSchemaForRenderer = JSON.parse(
+  JSON.stringify(item.rawSchema, (key, value) =>
+    key === "$id" && typeof value === "string" ? undefined : value,
```
SnowboardTechie: Nit: would be nice to pull the $id-stripping logic out into a small helper (e.g. stripIds(schema)) and add a unit test for it. As an inline expression here it's not really testable, so a future refactor of this block could quietly undo the fix.

SnowboardTechie (commented):
Nice work tracking down two distinct root causes here, the init-time failure and the SSR-side $id duplication. The sf424 regression form is a well-chosen test case.

A few things worth a look before approving, left comments inline. 

One extra ask not in the diff: now that nothing imports `extensionsAjv` as a validator, it'd be worth removing the export from `validation.ts` so the module-load trigger isn't lingering around.

SnowboardTechie (approved):
Thanks for the quick turnaround on the feedback! The `stripIds` extraction and its tests came out especially nice, and the new `example-generation` test is a great addition aswell.

widal001:
Hey @jcrichlake, thanks for pulling this fix together so quickly.

As I was reviewing your fixes, I realized something and decided to dig a little deeper on the root cause of this bug and why it seemed to only affect the forms and not the other schemas in which a single schema is also referenced multiple times. Here are a couple of notes based on my findings:

#### Root cause
                                                                                                      
First of all, great job diagnosing the root cause. Despite what my original bug ticket suggested, the duplicate `$id` after calling `dereferenceSchema()` was the core issue, and your `stripIds()` utility is the right fix. Thanks for finding and fixing that!
                                                                                                                                                                                      
#### Where my ticket led you astray

The issue I drafted pointed toward a few changes that turned out to be unnecessary:

1. I _thought_ that the shared `extensionsAjv` instance was the source of the issue because it threw an error when we added SF424 Mandatory that included multiple references to `Address.yml`. But the real source (as mentioned above) was the duplicate `$id` not the fact that the same schema was referenced multiple times.
2. I suggested per-call AJV and directory-split approaches. Those weren't necessary for the bug fix either, the throw happens purely when a dereferenced tree with duplicate $ids is handed to any AJV, regardless of which instance holds which schemas.
                                                                                                                                                                                                                                      
#### Why the other changes aren't worth the cost
                                                                                            
- Per-call AJV in `generateSchemaExample()`: `validateExample` only does `ajv.getSchema(filename)` on the file-level schema (which has `$ref` strings, not inlined subtrees), so the shared `extensionsAjv` handles it fine. Moving to per-call means reloading 166+ YAML files from disk on every `generateSchemaExample` call, which increases build times linearly with every new model, question, custom field, and form that we add.
- `.form-schemas/` directory split: Isn't needed now that we're stripping the duplicate `$id` and it adds a second typespec emit pipeline, which slows down the build.
                                                                                                                                                                                      
#### Minimum viable fix (#739):

Just to confirm that limiting our fix to `stripIds()` was enough I to fully resolve the issue, I created #739 which:

- Sets up a basic test to recreate the issue
- Moves the `stripIds()` call from `FormDetails.astro` into `dereferenceSchema()` (where that fix is needed)
- Re-adds the SF424 Mandatory form to confirm that this fix works with the more complex forms

Three commits, ~90 lines including the regression test, and I cherry-picked the changes from your commits on `stripIds()` and `sf424-mandatory.tsp`.
 
Happy to close #739 if you'd rather fold the changes into this PR. I mainly wanted a clean diff to show the minimal fix on its own. What do you think?

jcrichlake:
Closing this as @widal001 dug more and realized we just need a part of this PR.  He has a PR open for that and that is what we will merge in

Issue #730, opened 2026-04-15
SPIKE: Defining Plugin Enhancements Interface and Requirements


jcrichlake:
### Summary

As we continue to build upon the work done as part of the plugin/custom_fields framework we need to make some high level decisions around how we shape and leverage adapters. 

#### Key Decision Points
1. Will adapters be the only pathway to access sub-objects or will direct access to something like plugins be allowed?
2. Determine unified language so there isn't confusion around plugin vs adapter vs custom fields etc.
3. Define the high level Adapter interface i.e. the adapter object/class and the fields/methods we want to include at the highest level.

### Acceptance criteria

- [x] A draft ADR that captures the key decision points is written up and committed to the repository (It will be marked as complete as part of either (#657 or #646 )
- [x] The draft ADR and decision points are reviewed and agreed upon by the engineering team, either async or via a zoom call. 
- [x] Any updates or potential changes as a result of the ADR are made to the two follow on spikes (#657 and #646)z
- [x] Any implementation issues are written and assigned to the appropriate epic if needed
- [x] The draft ADR is added to the `common-grants` website

SnowboardTechie:
@jcrichlake from my side the ACs all read as met, but you opened this one so I'll leave the close call yours.

AC3 (updates back to the follow-on spikes):

[#657](https://github.com/HHS/simpler-grants-protocol/issues/657) is closed with the [ADR-outcome comment](https://github.com/HHS/simpler-grants-protocol/issues/657#issuecomment-4300097601) mapping each spike question to the ADR decision.
[#646](https://github.com/HHS/simpler-grants-protocol/issues/646) carries a Status note deferring custom-filter design until the transformation sub-issues close, and points at ADR-0022 for the plumbing.
AC4 (implementation issues filed under the epic):

Sub-issues of [#656](https://github.com/HHS/simpler-grants-protocol/issues/656):

[#756](https://github.com/HHS/simpler-grants-protocol/issues/756), [#757](https://github.com/HHS/simpler-grants-protocol/issues/757): TS and Py definePlugin() schemas extension
[#744](https://github.com/HHS/simpler-grants-protocol/issues/744): plugin-output and mapping-definition validation
[#745](https://github.com/HHS/simpler-grants-protocol/issues/745): port transformWithMapping and add the handler-registry API
[#746](https://github.com/HHS/simpler-grants-protocol/issues/746): migrate the PA / CA example plugins and publish docs
[...]

Issue #657, opened 2026-04-01
SPIKE: Plugin transformation contract and SDK integration strategy


SnowboardTechie:
## Summary

Research the minimal plugin transformation contract, its TypeScript and Python API shape, and how protocol-defined mappings and runtime utilities should integrate with that contract.

## Context

[ADR 0017](https://commongrants.org/governance/adr/0017-mapping-format) defines a JSON mapping format for declarative transformations. Runtime implementations already exist:

- **Python SDK**: `transform_from_mapping()` with `field` and `switch` handlers
- **Website**: `transformWithMapping()` with the same handler set

However, these utilities exist in isolation — there's no formal plugin contract that defines what a transformation interface should look like. The [PA](https://github.com/HHS/simpler-grants-protocol/tree/main/examples/pa-opportunity-example) and [CA](https://github.com/HHS/simpler-grants-protocol/tree/main/examples/ca-opportunity-example) example APIs use imperative class-based transformers (`OpportunityTransformer`) that don't use the mapping utilities at all.

This spike should define the transform contract first, then determine how mapping utilities fit behind that contract.

### Design constraints

The repo's public API style leans structural and simple:
- TypeScript plugin system uses structural typing
- Python uses simple dataclasses and config objects
- Avoid inheritance-heavy abstract base class hierarchies

The transform contract should feel like a small interface or plain object contract, not an elaborate class hierarchy.

## Questions to answer

1. **Minimal contract shape**: What is the minimal transform interface? A single function signature? A config object with transform methods?

2. **Directionality**: Should the contract be directional?
   - `transformToCommonGrants(source: S) -> T` (required)
   - `transformFromCommonGrants(cg: T) -> S` (optional, for round-trip scenarios)

3. **Structural vs class-based**: Should the contract be:
   - A structural interface (TS) / Protocol (Python) that any conforming object satisfies?
   - A factory function pattern?
   - A class-based approach (like the existing `OpportunityTransformer` in examples)?

4. **Mapping integration**: How does the JSON mapping format fit into the contract?
   - Required (all plugins must use mappings)?
   - Optional utility (one way to implement the contract)?
   - Separate concern (mapping runtime is independent of adapter contract)?

5. **Porting the website utility**: Can `website/src/lib/transformation.ts` be ported directly to `@common-grants/sdk`, or should it be adapted to fit behind the contract interface?

6.  **Validation boundaries**: What should the SDK validate, and at what layer?
[...]

Issue #646, opened 2026-03-31
SPIKE: Custom filter design and SDK integration strategy


SnowboardTechie:
## Summary

Research and recommend the approach for custom filter support across the Python and TypeScript SDKs. This spike answers the open design questions below and produces a refined set of sub-issues for the [custom filters epic (#645)](https://github.com/HHS/simpler-grants-protocol/issues/645).

## Status (2026-04-23)

**Deferred.** Do not pick up until transformation sub-issues [#744](https://github.com/HHS/simpler-grants-protocol/issues/744), [#745](https://github.com/HHS/simpler-grants-protocol/issues/745), [#746](https://github.com/HHS/simpler-grants-protocol/issues/746), [#756](https://github.com/HHS/simpler-grants-protocol/issues/756), [#757](https://github.com/HHS/simpler-grants-protocol/issues/757) close and Q3 (placement) is settled — filter-shape decisions made before then will churn. [ADR-0022 (Plugin Framework)](https://github.com/HHS/simpler-grants-protocol/blob/main/website/src/content/docs/governance/adr/0022-plugin-framework.mdx) merged in [#740](https://github.com/HHS/simpler-grants-protocol/pull/740) and already lands the plumbing custom filters will ride on.

## Protocol-level context

[ADR-0012](https://commongrants.org/governance/adr/0012-filtering/#custom-filters) defines custom filters at the protocol level: they go under a `customFilters` property in the request body and conform to the standard `Filter` schema (operation + value). The protocol already defines [base filter types](https://commongrants.org/protocol/filters/) (string, numeric, date, money).

The [Custom Fields Technical Spec](https://docs.google.com/document/d/1hT3d4g53nm0pSAxBWygNLyBVyPgHhpkmQtInlhGhl8U/edit?tab=t.0) established runtime composition and build-time generation patterns for custom fields. Custom filters may or may not follow that pattern — that's part of what this spike decides.

## What ADR-0022 already provides

Custom filters ride on the **plugin framework** (ADR-0022), not a separate adapter framework (the earlier "adapter framework" concept tracked in [#560](https://github.com/HHS/simpler-grants-protocol/issues/560) was subsumed by ADR-0022).

Fixed by ADR-0022:

- `"customFilters"` is a value in the `PluginCapability` enum.
- `CustomFilterSpec` interface: `{ filterType: "string" | "number" | "integer" | "boolean"; description?: string }`.
[...]

PR #740, opened 2026-04-20
[Issue #730] Plugin framework ADR

jcrichlake:
Summary

  - Adds ADR-0022 documenting the decision on how to structure the Adapter object in the expanded plugin framework
  - The ADR covers naming ("plugin" for registry catalog vs. "Adapter" for the SDK runtime object), top-level grouping strategy, and how custom fields and data mappings coexist
  - Three options were evaluated; the chosen approach uses functional top-level grouping (meta, client, schemas, extensions) with per-object grouping inside schemas

  Context

  The existing plugin framework supports custom field declarations. This ADR governs how to expand it to also support bidirectional data mappings (as specified in ADR-0017), while
  keeping the interface stable, DI-friendly, and consistent across both the Python/Pydantic and TypeScript/Zod SDKs.

  Decision

  Option 3: Functional top-level with per-object schema grouping was selected because it:

  - Keeps the top-level Adapter surface short and closed — new protocol objects only add a key under schemas, not at the root
  - Keeps client singular (one auth config, one base URL, one rate limiter per source system)
  - Enables clean dependency injection along functional lines (Client, Schemas, Extensions as coherent units)
  - Allows mergeExtensions() / merge_extensions() to operate on flat, serializable data
  - Preserves real coupling between native schema, CommonGrants schema, and bidirectional transforms inside schemas.<ObjectName>
  - Maintains backward compatibility — existing custom-fields-only plugins remain valid

  Test plan

  - Verify ADR renders correctly on the docs site
  - Confirm all internal links (e.g. to ADR-0017) resolve correctly


website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+
+_How should the Adapter object be structured to support both custom field declarations and data mappings, while enabling clean dependency injection and remaining stable as the protocol's object list grows?_
+
+### Questions
+
+- Should the framework still be called a "plugin", or does adding mappings warrant a new name?
+- Should the top-level Adapter structure group by feature (`meta`, `client`, `schemas`, `extensions`) or by object (`Opportunity`, `Application`, …)?
+- Should client configuration (auth, transport, rate-limiting) sit alongside per-object schemas, or be lifted to the top level as a system-level concern?
+- Should custom fields and mappings be coupled in the same package, or allowed independently?
```
SnowboardTechie: I don't see this getting an explicit answer in the Decision. The shape implies "independent" since both keys are optional, so maybe a one-line callout would close it?

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+```
+
+**TypeScript interface (Zod-based):**
+
+```ts
+interface AdapterMeta {
+  name: string;
+  version: string;
+  sourceSystem: string;
+  capabilities?: string[];
```
SnowboardTechie: I'm curious what you had in mind for `capabilities`. Enum, registry, or free-form for now? A pointer would help me plan for it in #743.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+//   - interpreting the declarative mappings (ADR-0017) into toCommon / fromCommon functions
+//   - native defaults to Record<string, unknown> (extensions is JSON-safe; runtime
+//     Zod schemas cannot be included)
+//
+// Client config (auth, base URL, rate-limiting) is not JSON-safe and must be
+// provided separately at runtime — it is never part of extensions.
+function defineAdapter(extensions: AdapterExtensions, client: Client): Adapter;
+
+// Combine multiple extension objects (e.g. from separate packages)
+function mergeExtensions(...extensions: AdapterExtensions[]): AdapterExtensions;
```
SnowboardTechie: I'm wondering about the merge semantics here. Deep or shallow, and what happens if two packages set the same `schemas.Opportunity.customFields.programArea`? I think it's worth a short note under Decision #5, since the multi-package composition story below depends on it.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
+---
+title: Adapter framework
+description: ADR documenting the decision to use an Adapter pattern with functional top-level grouping and per-object schema grouping for the expanded plugin framework that supports both custom fields and data mappings across the Python/Pydantic and TypeScript/Zod SDKs.
+---
+
+The existing plugin framework allows adopters to publish npm or PyPI packages that declare custom fields for CommonGrants schema objects. We want to expand this framework to also support bidirectional data mappings — declarative transforms that the SDK compiles into type-safe `toCommon`/`fromCommon` functions at runtime (see [ADR-0017](https://commongrants.org/governance/adr/0017-mapping-format)). We also want to organize this framework to support other potential future features with minimal rework.
```
widal001: Small nit: We should use a root-relative link here so that way it resolves correctly locally and in the PR preview rather than always pushing users to the hosted content.

You might need to check that exact relative path syntax, but something like that should work.

```suggestion
The existing plugin framework allows adopters to publish npm or PyPI packages that declare custom fields for CommonGrants schema objects. We want to expand this framework to also support bidirectional data mappings — declarative transforms that the SDK compiles into type-safe `toCommon`/`fromCommon` functions at runtime (see [ADR-0017](governance/adr/0017-mapping-format)). We also want to organize this framework to support other potential future features with minimal rework.
```

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+- Auth, transport, and rate-limiting are system-level concerns that belong to a single client, not distributed across per-object branches.
+- The top-level Adapter surface should be short and stable — adding new protocol objects should not expand the top-level key set.
+
+## Decision
+
+We decided to:
+
+1. **Keep "plugin" as the registry term** for published npm/PyPI packages displayed in the website catalog. No change to `PluginSourceEntry` or `src/content/plugins/index.json`.
+
+2. **Introduce "Adapter" as the SDK term** for the runtime object that wires together meta, client, schemas, and extensions for a single source system.
```
widal001: Personally I don't _love_ the idea of having adapters _and_  plugins, especially since these concepts and interfaces are sooo close to one another.

I'd suggest either:

- Updating `Plugin` and `definePlugin()` to support the additional interface elements we're describing in this ADR
- Renaming these to `Adapter` and `defineAdapter()` as part of this rework and then updating the plugins page accordingly.

widal001: If we made the various parts of the top-level interface optional, then you could still publish a plugin/adapter that only defines custom fields and doesn't include transformation logic, client, etc.

Or package authors could start by publishing a plugin/adapter that implements only part of this interface and then slowly expands the package to provide the others over time.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+adapter.extensions  // serializable input to defineAdapter(); used by mergeExtensions()
+adapter.schemas.<ObjectName> = { native, common, toCommon, fromCommon }
+```
+
+**TypeScript interface (Zod-based):**
+
+```ts
+interface AdapterMeta {
+  name: string;
+  version: string;
```
widal001: Question: Do we want to make `version` required? Or is there a way to populate this from the library itself?

I would hate for a package maintainer to have update this in two places, especially if (like us) they have automated versioning set up.

website/src/content/docs/governance/adr/0022-adapter-framework.md
widal001: What do we think would go in capabilities? If this is a list of supported operations can we make it an enum array?

Like:

- `customFields`
- `customFilters`
- `client`
- etc.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+  sourceSystem: string;
+  capabilities?: string[];
+}
+
+interface CustomFieldDef {
+  name?: string; // optional; dict key is used as the display name fallback
+  fieldType: string;
+  description?: string;
+}
```
widal001: We should just re-use `CustomFieldSpec`

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+  native: ZodType<TNative>;
+  common: ZodType<TCommon>;
+  toCommon: (native: TNative) => TCommon;
+  fromCommon: (common: TCommon) => TNative;
+}
+
+// Per-object config shape inside extensions.schemas — mirrors Python's AdapterExtensionsSchema
+interface AdapterExtensionsObjectConfig {
+  customFields?: Record<string, CustomFieldDef>;
+  mappings?: Record<string, unknown>; // see ADR-0017 for mapping format
```
widal001: How do we anticipate this will be used? And by whom?

If it's just for the transformation functions, would it make more sense to simply provide a utility function that can help an adapter author generate `toCommon()` and `fromCommon()` transformation functions from a mapping, and keep the mapping internal to those functions?

If we do want to support mappings here, we'll need to support both a `toCommon` and a `fromCommon` mapping as well.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+}
+
+// Serializable input to defineAdapter() — safe to store as JSON
+interface AdapterExtensions {
+  meta?: Partial<AdapterMeta>;
+  schemas?: Record<string, AdapterExtensionsObjectConfig>;
```
widal001: At least on the TypeScript side, rather than supporting _any_ string in `Record<string, AdapterExtensionsObjectConfig>` we'll want to re-use or follow the pattern of the existing [`SchemaExtensions` type](https://github.com/HHS/simpler-grants-protocol/blob/e6a49e8cda8e973ee02dddfc19fe89eb18189820/lib/ts-sdk/src/extensions/types.ts#L120-L162) which restricts the keys of that record to a known set of supported objects (i.e. [`ExtensibleSchemaName`](https://github.com/HHS/simpler-grants-protocol/blob/e6a49e8cda8e973ee02dddfc19fe89eb18189820/lib/ts-sdk/src/extensions/types.ts#L105-L118))

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+}
+
+// Client is a placeholder for the SDK's runtime client type (not shown here)
+interface Adapter {
+  meta: AdapterMeta;
+  client: Client; // one per source system
```
widal001: I'd recommend making this a method -- something like `getClient(config) -> Client` -- so that way a given adapter can request a standardized set of system-specific inputs and validate/override system-specific configs like auth token, max page size, timeouts, etc.

widal001: We could still set this up to be a singleton so that way the client is instantiated only once and each subsequent call to that method returns the same `Client` instance

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+//
+// defineAdapter builds adapter.schemas from extensions.schemas by:
+//   - extending the base CommonGrants schema with any declared customFields → common
+//   - interpreting the declarative mappings (ADR-0017) into toCommon / fromCommon functions
+//   - native defaults to Record<string, unknown> (extensions is JSON-safe; runtime
+//     Zod schemas cannot be included)
+//
+// Client config (auth, base URL, rate-limiting) is not JSON-safe and must be
+// provided separately at runtime — it is never part of extensions.
+function defineAdapter(extensions: AdapterExtensions, client: Client): Adapter;
```
widal001: See my note above on making client a method.

We'll also need some way to specify transformation logic, native schemas, etc. to generate the `ObjectSchemas` as part of `defineAdapter()`. And we'll also need to define a way to define custom filters.

While we don't need to lock all of that down in this ADR, since that's part of the goal of #646 and #657 I would recommend making the input to `defineAdapters()` an object (like [`DefinePluginOptions`](https://github.com/HHS/simpler-grants-protocol/blob/e6a49e8cda8e973ee02dddfc19fe89eb18189820/lib/ts-sdk/src/extensions/define-plugin.ts#L21-L31) is for `definePugin()` so that we can easily add new named optional inputs to it over time.

website/src/content/docs/governance/adr/0022-adapter-framework.md
widal001: We should think think more about what this means for the other aspects of extensions beyond custom fields.

For example, could we merge:
- Custom filters? (Probably)
- Transformation functions? (Probably not)
- Mappings? (Maybe? But this could get messy quick, depending on what we mean by merge)

We might want to rename this field or provide different merge functions for different parts of the extensions.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+:::note[Bottom line]
+Pure object-first / Plugin+Adapter is best if:
+
+- we want authoring a new object's support to feel self-contained during development
+- but we're willing to accept a top-level key set that grows with the protocol and scattered client config
+  :::
+
+- **Pros**
+  - Co-location: all of an object's config is in one branch during authoring
+  - Clear semantic split: "plugin" = what you install, "adapter" = what you construct in code — `Adapter` is a well-understood bridging pattern with a named, importable, type-hint enabled class
```
widal001: I'm not sure I'm totally sold on the clarity of this split 🤔 

In this case what happens to the existing `Plugin` and `definePlugin()`?

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+- **Cons**
+  - No named `Adapter` type to import, document, or type-hint against
+  - Client, auth, and transport have no natural home in this model
+  - DI requires passing each model's adapter separately rather than a unified `Schemas` object — callers must accept one adapted schema per object rather than a single `Schemas` unit (**DI-friendly: 🟡**)
+  - Top-level surface tracks the object list indirectly via function calls (`createAdapter(opportunitySchema, ...)`, `createAdapter(applicationSchema, ...)`), but there is no stable type that enumerates supported objects (**Stable surface: 🟡**)
+  - In Python, `create_adapter` must dynamically generate a new model class, which is less transparent
+
+### Option 2: Pure object-first structure, "Plugin" for registry / "Adapter" for SDK
+
+Config and runtime object both keyed by CommonGrants model name at the root. `meta` and `client` sit alongside object keys but are not themselves objects.
```
widal001: It feels like there might be two different decisions being mixed up in this option:

- Putting object-specific keys at the root level alongside `meta` and `client`
- The relationship between Plugins and Adapters, which still feels a bit unclear to me reading through this option.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+Usage:
+
+```ts
+// Client is configured separately — auth and transport are not JSON-safe
+const client = createClient({
+  baseUrl: "https://api.grants.gov",
+  auth: { type: "oauth2" /* ... */ },
+});
+
+const adapter = defineAdapter(
```
widal001: Instead of (or in addition to) showing an example of how an adapter, it could be helpful to show how a consumer would use an adapter authored by someone else.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+
+1. **Keep "plugin" as the registry term** for published npm/PyPI packages displayed in the website catalog. No change to `PluginSourceEntry` or `src/content/plugins/index.json`.
+
+2. **Introduce "Adapter" as the SDK term** for the runtime object that wires together meta, client, schemas, and extensions for a single source system.
+
+3. **Use functional grouping at the top level** with four keys — `meta`, `client`, `schemas`, and `extensions` — rather than grouping by object name at the root.
+
+4. **Use per-object grouping inside `schemas`** where it reflects real coupling: each object's native schema, CommonGrants schema, and bidirectional transforms are tightly coupled and change together.
+
+5. **Make `extensions` the serializable root** — the input to `defineAdapter()` and the unit operated on by `mergeExtensions()`.
```
widal001: I think the phrase "make `extensions` the serializable root" was in my comments in slack, but on further thought, I don't think `extensions` has to be strictly serializable. But having it largely match the input to `defineAdapter()` or at least the parts of an adapter that can be combined with other adapters using `mergeExtensions()` could be useful.

See [this comment](https://github.com/HHS/simpler-grants-protocol/pull/740/changes#r3120511807) about which parts of the adapter that might be.

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+```
+adapter.meta        // name, version, sourceSystem, capabilities
+adapter.client      // auth, transport, resource methods, filters
+adapter.extensions  // serializable input to defineAdapter(); used by mergeExtensions()
+adapter.schemas.<ObjectName> = { native, common, toCommon, fromCommon }
+```
+
+**TypeScript interface (Zod-based):**
+
+```ts
```
widal001: I might recommend adding these to a `<details><summary /></details>` section so ADR readers can shrink them up.

widal001: Same is true for other large code blocks

website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -0,0 +1,421 @@
...
+);
+
+// Combine extensions from multiple packages before constructing the adapter
+const merged = mergeExtensions(baseExtensions, grantsGovExtensions);
+const mergedAdapter = defineAdapter(merged, client);
+```
+
+**Python interface (Pydantic-based):**
+
+```python
```
widal001: We could also explore switching this to a `<Tabs />` block instead of stacking the Python and TypeScript examples on top of one another.

website/src/content/docs/governance/adr/0022-adapter-framework.md
widal001: It's worth noting that the need to make the options an object/interface is unique to TypeScript, since python allows you to add named optional params to the function's root.

widal001 (commented):
Thanks for updating this draft @jcrichlake !

I left a number of specific comments on the TypeScript section (which we'll likely want to carry over to the Python sections) but the big areas are:

- Clarifying what's happening with `Plugin` and `definePlugin()`. Are we:
  - Replacing them with `Adapter` and `defineAdapter()`
  - Keeping the names but evolving the interface and implementation
  - Keeping both (strongly recommend against this)
- Updating `defineAdapter()` inputs so that it can include, meta information, per-schema transformation logic, and route-specific custom filters, etc.
- Updating the `client` attribute to be a `getClient()` function signature since the adapter consumer will likely need to provide (at a minimum) an Auth token to get the final client instance.

website/src/content/docs/governance/adr/0022-adapter-framework.md
widal001: But it's probably worth stating explicitly, we don't necessarily want to require that `toCommon()` and `fromCommon()` transformation functions use a declarative `mapping` format.

website/src/content/docs/governance/adr/0022-adapter-framework.md
jcrichlake: So the existing `plugin` mergeExtensions code has a flag for what to do in this scenario called `onConflict`  We'd likely want to do the same here. The possible inputs are firstWins, lastWins or Error

website/src/content/docs/governance/adr/0022-adapter-framework.md
jcrichlake: Updated the decision to add another point to address this.

website/src/content/docs/governance/adr/0022-adapter-framework.md
jcrichlake: Updated this to make it a list of enums per Billy's suggestion.

website/src/content/docs/governance/adr/0022-adapter-framework.md
jcrichlake: @SnowboardTechie and @widal001 for awareness. I updated the ADR to make everything reference plugin as opposed to adapter.  In my mind plugin was the more 'open' term that would preserve more options for us in the future when adding functionality.  I could see things like having a lifecycle hook or some of the data fetching pieces fitting better into the term plugin. 

That being said I'm open to either if folks have strong feelings either way.

website/src/content/docs/governance/adr/0022-adapter-framework.md
jcrichlake: Great callout, I've updated the ADR so that mappings are not required but supported.

website/src/content/docs/governance/adr/0022-adapter-framework.md
jcrichlake: I explored this but it would require changing the ADR from .md to mdx which I wasn't sure if we wanted to do that at this juncture

website/src/content/docs/governance/adr/0022-plugin-framework.md
```diff
@@ -0,0 +1,669 @@
...
+//     for equivalent configs automatically
+//
+// toCommon / fromCommon may be plain hand-written functions or generated via
+// buildTransforms(). Either way they are provided in schemas, not derived from extensions.
+function definePlugin(options: DefinePluginOptions): Plugin;
+
+// Combine multiple extension objects (e.g. from separate packages)
+function mergeExtensions(
+  ...extensions: PluginExtensions[],
+  onConflict = "firstWins",
```
SnowboardTechie: Small snag: TypeScript requires the rest parameter to be last (TS1014), so `onConflict` after `...extensions` won't compile. An options-object tail would keep the spread ergonomics:

```ts
function mergeExtensions(
  extensions: PluginExtensions[],
  options?: { onConflict?: "firstWins" | "lastWins" },
): PluginExtensions;
```

Worth locking down here since this snippet will likely get copy-pasted into the SDK.

website/src/content/docs/governance/adr/0022-plugin-framework.md
```diff
@@ -0,0 +1,669 @@
...
+# to_common / from_common may be plain hand-written callables or generated via
+# build_transforms(). Either way they are provided in schemas, not derived from extensions.
+def define_plugin(
+    meta: PluginMeta | None = None,
+    get_client: Callable[[ClientConfig], Client] | None = None,
+    extensions: PluginExtensions | None = None,
+    schemas: dict[ExtensibleSchemaName, ObjectSchemasInput[Any, Any]] | None = None,
+    filters: dict[ExtensibleSchemaName, dict[str, CustomFilterSpec]] | None = None,
+) -> Plugin: ...
+def merge_extensions(*extensions: PluginExtensions, on_conflict: "first_wins") -> PluginExtensions: ...
```
SnowboardTechie: Mirror of the TS note above, plus a Python quirk: per PEP 484, `on_conflict: "first_wins"` gets parsed as a forward reference (bare string annotations reference a type named `first_wins`), and the default from the TS signature doesn't carry over. Suggested:

```python
from typing import Literal

def merge_extensions(
    *extensions: PluginExtensions,
    on_conflict: Literal["first_wins", "last_wins"] = "first_wins",
) -> PluginExtensions: ...
```

website/src/content/docs/governance/adr/0022-plugin-framework.md
```diff
@@ -0,0 +1,669 @@
...
+
+Top-level keys are functional (`meta`, `client`, `schemas`, `extensions`). Per-object grouping is used only inside `schemas`, where it reflects real coupling between native schemas, CommonGrants schemas, and bidirectional transforms.
+
+<details>
+<summary>Option 3 code example</summary>
+
+```ts
+interface Plugin {
+  meta?: PluginMeta;
+  client?: Client;
```
SnowboardTechie: Tiny thing: this sketch still shows `client?: Client`, but the `Plugin.getClient` field at line 120 and the plugin-shape prose at line 45 both use `getClient`. The Decision section calls "`getClient()` is memoized" out as an Option 3 benefit, so updating this to `getClient?: (config: ClientConfig) => Client;` keeps the chosen-option sketch consistent with the rest of the ADR. Also lines up with @widal001's open thread on the same naming.

website/src/content/docs/governance/adr/0022-plugin-framework.md
```diff
@@ -0,0 +1,669 @@
...
+- Auth, transport, and rate-limiting are system-level concerns that belong to a single client, not distributed across per-object branches.
+- The top-level Plugin surface should be short and stable — adding new protocol objects should not expand the top-level key set.
+
+## Decision
+
+We decided to:
+
+1. **Keep "plugin" as the unified term** for both the published npm/PyPI packages in the website catalog and the runtime SDK object. No change to `PluginSourceEntry` or `src/content/plugins/index.json`. The existing `definePlugin()` function is expanded to accept the full set of top-level fields described below.
+
+2. **Use functional grouping at the top level** with four keys — `meta`, `client`, `schemas`, and `extensions` — rather than grouping by object name at the root.
```
SnowboardTechie: `filters` ends up a bit orphaned: this point enumerates four top-level keys, but the interfaces at lines 125 and 141 include a fifth `filters` key. Two options: enumerate five keys here, or move filters under `schemas.<Object>.customFilters` to keep the top-level surface at four. The `"customFilters"` value in `PluginCapability` (line 56) suggests the second was the original intent. Easier to pick a lane now than revisit after the ADR lands.

website/.cspell.json
```diff
@@ -35,7 +35,12 @@
     "USCDI",
     "rescope",
     "rescoped",
-    "underrepresents"
+    "underrepresents",
+    "commongrants",
```
SnowboardTechie: Heads up: these five entries (`commongrants`, `HBCUs`, `TCCUs`, `cfda`, `CFDA`) don't appear in the ADR body (grep only turns up the existing `CommonGrants` PascalCase form). Looks like a carry-over from an earlier draft. Probably cleanest to drop them from this PR.

SnowboardTechie (changes requested):
Solid writeup. The three-option analysis with the side-by-side criteria table makes the tradeoffs legible, and the parallel TS/Python interfaces with matching examples keep the two SDKs in lockstep.

Left five line comments. Most are tightening passes on the embedded type signatures rather than pushback on the decision itself, but since landed ADRs end up treated as immutable in this repo and these snippets will likely get copy-pasted into the SDK, worth catching here. Summary:

- Two signature bugs on `mergeExtensions` / `merge_extensions` that won't compile as written
- A couple of internal consistency nits on `client` vs `getClient` and where `filters` lives
- One cspell cleanup that looks like carry-over from an earlier draft

Nothing blocking the direction.

website/src/content/docs/governance/adr/0022-plugin-framework.md
widal001: Two notes on this comment @jcrichlake @SnowboardTechie 

- We want to preserve the `error` option and make it the default (for TypeScript at least) because that option preserves the detailed type inference we want for plugins. See [this docstring for more details](https://github.com/HHS/simpler-grants-protocol/blob/6b6d5e58d92b4ca19a63b2b708ed0738cefa289f/lib/ts-sdk/src/extensions/merge-extensions.ts#L25-L32)
- Also, we shouldn't be copying and pasting anything from this ADR into the SDK if the SDK already has an equivalent concept implemented in the current plugin framework, unless we compare the two and ensure that the change proposed in this ADR actually matches the _behavior_ we want.

Personally, I'm not too worried about locking down the exact syntax in this ADR because it will likely need to change as we start actually implementing these decisions. Instead we should just create a final task to clean up this ADR when we're ready to publish the next minor versions of these SDKs to update this ADR to match the actual implementation.

website/src/content/docs/governance/adr/0022-plugin-framework.md
widal001: In addition to the point that @SnowboardTechie raised above, one other thing I want to note is that during implementation we'll want to explore if there's an easy way to allow different plugins to provide their own custom `ClientConfig` or something similar.

Since each plugin might support a different auth scheme, and have different options for configuring things like max page size and max request timeout. And we don't want plugin consumers to have to guess which auth method to use for a given plugin.

website/src/content/docs/governance/adr/0022-plugin-framework.md
```diff
@@ -0,0 +1,669 @@
...
+
+```
+plugin.meta        // name, version, sourceSystem, capabilities
+plugin.getClient   // (config: ClientConfig) => Client; memoized by definePlugin()
+plugin.extensions  // serializable; used by mergeExtensions()
+plugin.schemas.<ObjectName> = { native, common, toCommon, fromCommon }
+```
+
+<details>
+<summary>TypeScript interface (Zod-based)</summary>
```
widal001: Thanks for adding these details and summaries.

For readability and easier comparison across Python and TypeScript, I might actually lean toward using `<Tabs>` and break these examples into a couple of sections:

```
### Example interface

<Tabs>
  <TabItem> TypeScript </TabItem>
  <TabItem> Python </TabItem>
</Tabs>

### Example author usage

<Tabs>
  <TabItem> TypeScript </TabItem>
  <TabItem> Python </TabItem>
</Tabs>

### Example consumer usage

<Tabs>
  <TabItem> TypeScript </TabItem>
  <TabItem> Python </TabItem>
</Tabs>

### Consequences
...
```

Right now it looks a little crammed:

Image

You can see an example of this in [the custom fields overview](https://commongrants.org/custom-fields/overview/#defining-custom-field-specs)

widal001 (approved):
I left a few comments on a couple of the threads that Bryan started and one more stylistic comment about using `<Tabs>` to make it easier to compare Python and TypeScript patterns.

But otherwise I think we're aligned enough on the pattern here to call the first cut of this ADR done. We'll definitely need to come back and update it though as we learn more about the implementation pattern later.

website/src/content/docs/governance/adr/0022-plugin-framework.md
widal001: We'll likely need an explicit `filters` key in `DefinePluginOptions` passed to `definePlugin()` but the resulting `Plugin` interface _shouldn't_ have a `filters` key since the `Client` returned by `getClient()` should already support those filters by default.

website/src/content/docs/governance/adr/0022-adapter-framework.md
widal001: It's an easy switch! Basically just changing the file extension, then Astro renders everything else exactly the same.

website/src/content/docs/governance/adr/0022-plugin-framework.md
SnowboardTechie: Yeah, my bad, I missed the error default entirely. Agreed on deferring the exact syntax to post-implementation cleanup.

website/src/content/docs/governance/adr/0022-plugin-framework.md
jcrichlake: Great call out! I've added these and it looks much cleaner!

website/.cspell.json
jcrichlake: Removed these, thanks for catching that!

Issue #745, opened 2026-04-20
[SDK] Port transformWithMapping to @common-grants/sdk; add handler-registry API


SnowboardTechie:
### Summary

Port `website/src/lib/transformation.ts` into `lib/ts-sdk/src/extensions/transformation.ts`, update the website to import from `@common-grants/sdk`, and expose `buildTransforms()` with a per-call `handlers?` argument as the public API in both SDKs — matching ADR-0022 Decision #8 (handlers registered per call, not globally). The Python equivalent (`common_grants_sdk/utils/transformation.py`) is already in the SDK but needs its public API formalized.

### Acceptance criteria

- [ ] `website/src/lib/transformation.ts` is moved to `lib/ts-sdk/src/extensions/transformation.ts` with only the `JsonValue` import path updated
- [ ] `website/src/lib/transformation.ts` re-exports from `@common-grants/sdk` (no duplication)
- [ ] `@common-grants/sdk` exports `buildTransforms(toCommonMapping, fromCommonMapping, handlers?)` returning `{ toCommon, fromCommon }` and `safeTransform(transform, inputs)` returning `{ results, errors }`
- [ ] `common-grants-sdk` exports `build_transforms(to_common_mapping, from_common_mapping, handlers=None)` and `safe_transform(transform, inputs)` with matching semantics
- [ ] Default handler set (`field`, `switch`) is exported and documented; `Handler` type is exported for custom-handler typing
- [ ] Name collisions between user-supplied handlers and the default set raise at `buildTransforms()` load time (not silently shadowed)
- [ ] Per-call handler scoping documented with an example

### Related

- Parent epic: #656
- Origin spike: #657
- ADR: [ADR-0022 — Plugin framework](https://commongrants.org/governance/adr/0022-plugin-framework) (canonical URL once merged; currently drafted in #740 + #753)
- Source file: `website/src/lib/transformation.ts`
- Python source: `lib/python-sdk/common_grants_sdk/utils/transformation.py`

SnowboardTechie:
FYI: ADR 0022 Decision #8 in #753 was updated with one handler-registry guardrail worth noting before implementation starts.

[...]

Issue #744, opened 2026-04-20
[SDK] Implement plugin-output + mapping-definition validation


SnowboardTechie:
### Summary

Implement the validation strategy defined in ADR-0022, Decisions #7 and #9: `definePlugin()` wraps the final `toCommon` / `fromCommon` with runtime schema validation (`.parse()` / `model_validate()`) and produces a `TransformResult<T>` of `{ result, errors }` unconditionally; mapping definitions are validated at `buildTransforms()` call time. Transformation errors extend a single `PluginError` base with structured context.

### Acceptance criteria

- [ ] Implement [ADR-0022](https://commongrants.org/governance/adr/0022-plugin-framework) Decisions #6–9
- [ ] Review-driven refinements not captured in the ADR text:
  - `buildTransforms()` treats declared `customFields` as valid target paths during field-path resolution (per @jcrichlake, [#753 thread](https://github.com/HHS/simpler-grants-protocol/pull/753#discussion_r3138157676))
  - `PluginError` does not carry an `index` field; batch position attribution is caller-side via `.map` / list comprehension (per @jcrichlake, [#753 thread](https://github.com/HHS/simpler-grants-protocol/pull/753#discussion_r3138231034))
  - `PluginError.sourceValue` excluded from default string / repr output — strengthens Decision #9's adopter-redaction stance into a defensive SDK default
- [ ] Unit tests cover each refinement above plus the ADR-specified behaviors

### Related

- Parent epic: #656
- Origin spike: #657
- ADR: [ADR-0022 — Plugin framework](https://commongrants.org/governance/adr/0022-plugin-framework) (currently drafted in #740 + #753)
- ADR-0017: [Mapping format](https://commongrants.org/governance/adr/0017-mapping-format)
- Decisions #7–9 were shaped by #753 review, notably @widal001's [`TransformResult` signature thread](https://github.com/HHS/simpler-grants-protocol/pull/753#discussion_r3134167544)

jcrichlake:
Safe-transforms was a hold over from a now defunct design and is not needed. The other areas have been addressed.

Issue #656, opened 2026-04-01
[SDK] Support transformation logic in SDK


bryan-thompsoncodes:
## Summary

Define and implement data transformation support in the Python and TypeScript SDKs, including mapping specifications, built-in transformation handlers, and extensibility for custom transformations.

## Context

[ADR 0017](https://commongrants.org/governance/adr/0017-mapping-format) establishes a custom JSON mapping format for translating between platform-specific grant data and CommonGrants models. The format supports transformations like field extraction, value switching, type conversion, and string manipulation.

The Python SDK already implements a basic `transform_from_mapping()` function with `field` and `switch` handlers:

```python
from common_grants_sdk.utils.transformation import transform_from_mapping

mapping = {
    "title": { "field": "opportunity_title" },
    "status": { 
        "switch": {
            "field": "opportunity_status",
            "case": { "posted": "open", "closed": "closed" },
            "default": "custom"
        }
    }
}

result = transform_from_mapping(source_data, mapping)
```

The TypeScript SDK has no equivalent. Both SDKs need to support the full transformation vocabulary from ADR 0017.

## Acceptance Criteria

- [ ] TypeScript SDK has `transformFromMapping()` with parity to Python implementation
- [ ] Both SDKs support core handlers: `field`, `switch`, `const`
- [ ] Both SDKs support type conversion handlers: `numberToString`, `stringToNumber`
- [ ] Both SDKs support string manipulation: `concat`, `split`, `replace`, `toUpperCase`, `toLowerCase`
- [ ] Handler registry is extensible (users can add custom handlers)
- [ ] Mapping specifications are validated against a JSON schema
- [ ] Transformation errors include clear context (path, handler, value)

PR #753, opened 2026-04-21
docs(adr): #657 Q5-Q8 sub-decisions for ADR 0022

SnowboardTechie:
### Summary

- Part of #657
- Time to review: ~15 minutes

### Changes proposed

Extends the Plugin framework in [ADR-0022](https://cg-pr-753.billy-daly.workers.dev/governance/adr/0022-plugin-framework/) with four sub-decisions addressing the remaining questions from SPIKE #657 (Q5–Q8). Folded into the existing Decision list as items 6–9 on top of the structural decisions merged via #740.

- **Q5 Porting existing runtimes** — Expose `buildTransforms()` / `build_transforms()` as a public SDK utility that wraps the existing mapping runtimes. Thread `mappings?` as a JSON-safe sibling of `customFields` under `PluginExtensions.schemas.<Object>`; when declared and no explicit transform is supplied in `schemas.<Object>`, `definePlugin()` auto-invokes `buildTransforms()`. Each direction is author-provided — automatic inversion is structurally impossible for many-to-one handlers like `switch`. First-release handler set: `field`, `switch`.
- **Q6 Validation strategy** — `toCommon` / `fromCommon` return `TransformResult<T>` of `{ result, errors }` unconditionally. `definePlugin()` wraps the underlying transform output with runtime schema validation (`.parse` / `model_validate`) and appends any validation failures to `errors` rather than throwing. Mapping definitions are validated at `buildTransforms()` call time (structural errors do throw). Batch processing is a plain `.map` / list comprehension; no separate `safeTransform` variant.
- **Q7 Handler extensibility** — Per-call handler registry via optional `handlers` argument on `buildTransforms()`. No global registry. Name collisions with the default set raise at `buildTransforms()` call time. Handler-name lookup uses own-property access (`Object.hasOwn` TS / dict `in` Python) to block prototype pollution when mapping JSON is reconstituted from untrusted extensions via `mergeExtensions()`.
- **Q8 Error handling** — Structured `PluginError` base (TS interface extending `Error`; Python exception with explicit `__init__`) carrying `path` / `handler` / `sourceValue` / `cause`. `sourceValue` may contain PII when transforming applicant data; adopters redact before logging or re-raising. Specific subclasses documented alongside implementation in #744 / #745.

### Context for reviewers

**Scope.** This PR extends ADR 0022 with sub-decisions within the existing framework; it does not touch the structural decision (`definePlugin(options)` with functional top-level grouping). All four sub-sections link to the relevant implementation sub-issues under epic #656 (#744, #745).

**Base-ADR drift** from #740 noticed but kept out of this diff (signature / naming / orphan-key inconsistencies outside Q5–Q8 scope) is tracked in #759.

**Verification.** `+108 / −8` single-file diff; CI green. Read-through for internal consistency: Decision 6–9 language matches TS/Python interface additions; new `mappings` field is symmetric with `customFields`; `PluginError` documented alongside the utilities that raise it.

### Additional information

ADR grew from 679 → 779 lines (+100 net). The base ADR is already on the heavy side for the framework-level scope; additions are intentionally tight.


website/src/content/docs/governance/adr/0022-adapter-framework.md
```diff
@@ -37,6 +37,14 @@ We decided to:
...
 
+6. **Port the existing mapping runtimes as-is for the forward direction.** Reuse `website/src/lib/transformation.ts` and `common_grants_sdk.utils.transformation` unchanged — both already accept an injected handler registry. `fromCommon` is author-provided; automatic inversion is structurally impossible for many-to-one handlers like `switch`. First-release handler set is `field` and `switch`; expansion toward [ADR-0017](https://commongrants.org/governance/adr/0017-mapping-format)'s broader vocabulary is scoped under [#745](https://github.com/HHS/simpler-grants-protocol/issues/745).
+
+7. **Validate adapter output at runtime; validate mapping definitions at load time.** `defineAdapter` wraps `toCommon` / `fromCommon` with `schema.parse` / `model_validate` because compile-time types do not guarantee runtime shape. Mappings are checked when `defineAdapter` is called, failing fast on structural errors, unknown handlers, or unresolvable field paths. A non-throwing `safeTransform` / `safe_transform` variant returns aggregated errors. Implementation tracked under [#744](https://github.com/HHS/simpler-grants-protocol/issues/744).
```
widal001: I'd recommend wrapping this automated validation into a `generateFromCommonFunc()` and `generateToCommonFunc()` (or something similar) internal utility.

So that the adapter author can pass a mapping and a `native` schema to one of those utility functions and generate the `toCommon(native: NativeT) -> CommonT` and `fromCommon(common: CommonT) -> NativeT` methods that would be exposed by an adapter: `adapter.schemas.Opportunity.toCommon()`

widal001:
@SnowboardTechie in general I like your proposals of bringing the Python and TypeScript SDKs closer to parity on their support for declarative mapping based transformations, but I'm not sure we want to _require_ all `toCommon()` and `fromCommon()` transformation functions to use declarative mappings.

Instead I'd focus these changes to the ADR on defining how someone could provide functions that satisfy the `toCommon()` and `fromCommon()` functions in the `Adapter` interface and offering a set of utility functions for generating those functions from a mapping (see [comment above](https://github.com/HHS/simpler-grants-protocol/pull/753#discussion_r3120656603)).

That would allow adapter authors to use the mapping format if they wanted to, or to write their own transformation function for more complex transformation logic and validation.

SnowboardTechie:
Yeah, good call. Requiring mappings for every transform was too restrictive. Reworked the ADR in `7f8b174` so the primary interface is author-provided functions, with mappings as a convenience path:

- `Adapter.schemas.<Object>` exposes `toCommon` / `fromCommon` as functions, so authors can write them directly when the logic is complex.
- `generateToCommonFunc(mapping, native, common)` and `generateFromCommonFunc(...)` (Python: snake_case) are now public SDK utilities that wrap the existing mapping runtimes.
- `defineAdapter` picks up an optional third arg, `overrides`, for per-object `{ native, toCommon, fromCommon }`. If `overrides[Object].toCommon` is supplied it wins. Otherwise, when `extensions.schemas.<Object>.mappings` is declared, `defineAdapter` auto-generates via `generateToCommonFunc`.
- `fromCommon` stays always author-provided (direct function or explicit `generateFromCommonFunc` call), since many-to-one handlers like `switch` can't be auto-inverted. The reverse utility only supports the reversible subset until ADR-0017 defines reversible mapping semantics.

This keeps the mapping-only path zero-ceremony (two-arg `defineAdapter(extensions, client)` still works, `mergeExtensions` flow unchanged) while giving authors with complex transforms a first-class way to bypass mappings. Happy to adjust the `overrides` shape or the auto-interpret-on-default behavior if either feels off.

website/src/content/docs/governance/adr/0022-adapter-framework.md
SnowboardTechie: Nice, added both in `7f8b174`. `generateToCommonFunc` / `generateFromCommonFunc` (Python: `generate_to_common_func` / `generate_from_common_func`) take `(mapping, nativeSchema, commonSchema, handlers?)` and return the transform function. `defineAdapter` calls them automatically when `extensions.schemas.<Object>.mappings` is declared and no explicit override is supplied, and authors can also call them directly and attach the result via the new `overrides` arg.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -40,6 +40,14 @@ We decided to:
...
+
+7. **Validate plugin output at runtime; validate mapping definitions at utility-invocation time.** `definePlugin()` wraps the final `toCommon` / `fromCommon` (however produced) with runtime schema validation, because compile-time types do not guarantee runtime shape. Mappings passed to `buildTransforms()` are checked at the call site, failing fast on structural errors, unknown handlers, or unresolvable field paths. A non-throwing `safeTransform` / `safe_transform` variant returns aggregated per-item errors for batch transformation. Tracked under [#744](https://github.com/HHS/simpler-grants-protocol/issues/744).
```
widal001: > A non-throwing `safeTransform` / `safe_transform` variant returns aggregated per-item errors for batch transformation.

I really like the idea behind this but have a few questions about how this fits into the overall spec:

- Are we thinking of `safeTransform` as just a utility function that plugin consumers could use? And if so could we maybe provide an example of what this might look like in the usage section?
- Haven't given this much thought, so might not be a good idea for various reasons, but have we given any thought to just changing the function signature for `toCommon()` and `fromCommon()` to return `TransformResult<T> {result: T, errors: PluginError[]}` basically forcing these functions to be safe transformation functions?

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -167,16 +186,38 @@ function mergeExtensions(
...
   toCommon: (native: TNative) => TCommon;
   fromCommon: (common: TCommon) => TNative;
 };
+
+// Non-throwing batch variant — applies a transform to each input and returns aggregated
+// per-item results and errors (see Decision #7). Return shape finalized in #744.
+function safeTransform<TIn, TOut>(
+  transform: (input: TIn) => TOut,
+  inputs: TIn[],
+): { results: (TOut | null)[]; errors: PluginError[] };
```
widal001: You can refine this in implementation, but one thing to consider here: If a single record has multiple errors, you might have a tough time collating errors with the results/inputs they affect.

An alternative approach would be to return something like `TransformResult[]` where `TransformResult` is:

```ts
interface TransformResult<T> {
  result: T,
  errors: PluginError[]
}
```

That obviously has a different tradeoff of requiring users to iterate through the result set again to isolate the results, but there are some other patterns maybe worth considering that balance both of these goals like:

```ts
interface TransformResults<TIn, TOut> {
  results: TOut[],
  errors: Record<TIn, PluginError[]>
}
```

Not sure if TypeScript would support a generic key in `Record<>` but generally speaking, we'll just want to balance the happy path (fetching all successfully parsed values) with the unhappy path (accessing all errors for the same unsuccessful record)

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -40,6 +40,14 @@ We decided to:
...
+
+7. **Validate plugin output at runtime; validate mapping definitions at utility-invocation time.** `definePlugin()` wraps the final `toCommon` / `fromCommon` (however produced) with runtime schema validation, because compile-time types do not guarantee runtime shape. Mappings passed to `buildTransforms()` are checked at the call site, failing fast on structural errors, unknown handlers, or unresolvable field paths. A non-throwing `safeTransform` / `safe_transform` variant returns aggregated per-item errors for batch transformation. Tracked under [#744](https://github.com/HHS/simpler-grants-protocol/issues/744).
+
+8. **Custom handlers are registered per utility call, not globally.** `buildTransforms()` accepts an optional `handlers` argument for registering additional handler names. Per-call scoping keeps behavior explicit and testable; name collisions with the default set raise at `buildTransforms()` call time rather than silently shadowing them. Handler-name lookup must not resolve inherited attributes, because mapping JSON can be reconstituted from untrusted sources via `mergeExtensions()`. Tracked under [#745](https://github.com/HHS/simpler-grants-protocol/issues/745).
```
widal001: Not saying we should change this, but I'm on the fence about linking to tickets that implement the decisions outlined in an ADR.

On the one hand, it's nice to be able to track (downstream) which PRs implement a given decision.

On the other hand, it makes this ADR feel more like a project plan than a record of past architectural decisions and their rationale. It's also worth noting that the other decisions in here don't have the equivalent linked issues.

Maybe worth pulling into a future huddle or KT session to discuss our approach here since we'll probably want to be consistent moving forward.

widal001 (approved):
I left a few comments that I'd like us to discuss as we move forward toward implementation, but none of these would necessarily block merging this PR.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
SnowboardTechie: Yeah, safeTransform is for consumers doing batch work, like a pipeline running toCommon over N records where you want to collect errors before failing. I'll add an example.

On changing the signatures, I think you're right. Having `toCommon`/`fromCommon` throw leaves the single-record path implicitly unsafe at the type level, and partial failure is routine for cross-schema transforms, not exceptional. `TransformResult<T>` as the unconditional return is probably the better default.

That folds your other comment in too: batch becomes `TransformResult<T>[]`, and `safeTransform` either goes away or becomes a .map helper.

I'll update Decision #7 to reflect this and pin the shape in #744. Lmk if there's a python or async wrinkle I'm missing.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
SnowboardTechie: Yeah, fair call. Honestly the ADR reads cleaner without the implementation refs sitting in the decision list. I like having a paper trail but I agree, that stuff belongs in the epic and PR history. Dropping #744/#745 from this PR.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
SnowboardTechie: I'd go with your first sketch. One thing worth naming: field handlers can emit warnings without invalidating a record, so items can legitimately carry both a result and errors. That makes me want to resist adding anything to TransformResult beyond result + errors[]. No success flag, no status enum. Let each consumer apply their own rule for what counts as success, since strict vs lenient will diverge.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -40,6 +40,14 @@ We decided to:
 
 5. **Make all top-level Plugin fields optional** so adopters can publish a plugin that provides only the features they need — for example, custom fields only — and expand to include transforms, client config, or additional schemas incrementally over time.
 
+6. **Plugin authors provide `toCommon` / `fromCommon` as functions; mappings are one way to generate them.** The SDK exposes `buildTransforms()` / `build_transforms()` as a public utility wrapping the existing mapping runtimes. `PluginExtensions.schemas.<Object>` gains an optional `mappings` key carrying JSON-safe `toCommon` / `fromCommon` mapping objects; when those are declared and no explicit transform is supplied in `schemas.<Object>`, `definePlugin()` invokes `buildTransforms()` automatically. Mappings for each direction are author-provided — `buildTransforms()` does not invert one direction into the other, because many-to-one handlers like `switch` are not reversible.
```
jcrichlake, line 43: I don't think this needs noted in the ADR but wanted to bring up for awareness that we should consider custom fields in the mappings so any `buildTransforms()`/ `build_transforms()` calls should have custom fields registered as the first step.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
jcrichlake: Would it be best to just have our transforms be non-batching for our first pass on this work? As it stands the different features we are adding are fairly large in scope and anyone adopting this framework could add in their own batching in whatever way they prefer.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
SnowboardTechie: Good callout. Agree this belongs on the implementation side rather than the ADR. Logged it on #744 so the custom-field registration step doesn't get lost when we wire up `buildTransforms()`.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
SnowboardTechie: Yeah, that's basically where we ended up in 42cc90b after Billy's thread above. `toCommon` / `fromCommon` are single-record returning `TransformResult<T>`, and the batch examples are just plain `.map` / list comprehension over that, no helper. You may have caught that already, flagging in case you were reading an earlier snapshot. Happy to lock "no batch helper in v1" into Decision # 7 explicitly if it's worth pinning.

PR #755, opened 2026-04-22
[Issue #733] ADR for organization identifiers

### LINKED ISSUE #733: [Org IDs] ADR: Organization identifier format
### Summary

We need an extensible way to map multiple identifiers to a given organization that can replace the hardcoded `ein`, `uei`, and `duns` fields on `OrganizationBase` and ideally be reused across other models (`OpportunityBase`, `ApplicationBase`, `AwardBase`).

### Options

This decision comprises four sub-questions:

1. **Collection structure:** Flat array or keyed map?  
2. **Top-level ID format:** UUID or composite identifier string?  
3. **Registry code format:** org-id.guide codes (`US-EIN`) or namespaced (`us:ein`)?  
4. **Identifier structure:** How much metadata belongs on each identifier instance?

### Decision Criteria

* **Cross-system matching:** Can identifiers match the same record across systems?  
* **Within-system location:** Can an identifier locate and update a record in a given system?  
* **Standards interop:** Does the approach work with org-id.guide?  
* **Pattern alignment:** Does it follow existing CommonGrants conventions?  
* **Format validation:** Can well-known ID formats (EIN, UEI) be validated at the schema level?  
* **Extensibility:** Can new identifier types be added without protocol changes?  
* **Multi-ID support:** Can a record have multiple IDs in the same registry?  
* **Reusability:** Can the same types be used across models?

### Approvers

_No response_

### Acceptance criteria

- [ ] The approvers for this decision have been identified (ideally before work on the ADR starts)
- [ ] The ADR has been added to the list of ADRs in `website/src/content/docs/decisions/adr/` with a number prefix like `0002-decision-summary.md`
- [ ] The ADR has been reviewed and approved by the approvers listed above
- [ ] The ADR satisfies requirements that are outlined in the ADR template

widal001:
### Summary

Proposes an identification system that supports both uniquely identifying an organization record _within_ a given system and matching organization records _across_ multiple systems.

- Fixes #733 
- Time to review: 10 minutes

Adds `/governance/adr/0023-org-ids`

#### Instructions to review

Go to https://cg-pr-755.billy-daly.workers.dev/governance/adr/0023-org-ids/ and review the ADR.

#### Notes

- The decisions outlined in this PR were informed by conversations with stakeholders from Fluxx and 360Giving/Open Data services.
- The exact implementation of the proposed identifiers approach will be finalized in the next release of the `@common-grants/core` library, so this ADR may continue to evolve slightly between this draft and that release

Screenshot 2026-04-22 at 5 32 44 PM


website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+  }
+}
+```
+
+At a glance, the key decisions for this identification system are:
+
+1. **Top-level ID:** A pure UUID on each resource, with the hosting system's own ID also surfaced inside the collection as `systemId`.
+2. **Collection structure:** A map keyed by short base-identifier aliases, with a catch-all `otherIds` map for extensions.
+3. **Map key format:** Model-defined camelCase aliases (`ein`, `uei`, `duns`), not raw registry codes.
+4. **Identifier metadata structure:** A hybrid shape with `registry` (code), two small registry-level fields (`scope`, `kind`), and instance-level fields (`id`, `compositeId`, `uri`, `verifiedAt`). Full registry metadata (issuer, human-readable name, schema, lookup template) lives in an external CommonGrants registry catalog.
```
jcrichlake: Should the Full registry metadata list of fields have something that links it to either the small registry-level fields or the instance level fields?

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+5. **Multiple IDs per registry:** `id` for the primary, plus an optional `allIds` array of `{ id, status }` objects for additional and archived IDs.
```
jcrichlake: In the example above `allIds` doesn't include the value listed for `id`.  I'd say it should either include that value or `allIds` should get renamed to otherIds.

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+6. **Parent organization:** `parent` carries `id`, `name`, and optionally an embedded `identifiers` collection. Parent identifiers do not leak into the child's collection.
+7. **Registry code format:** org-id.guide codes (`US-EIN`, `GB-CHC`) for well-known public registries, `namespace:type` (`grants.gov:org`, `candid:bridge`) for everything else.
+8. **Registry governance:** CommonGrants maintains a registry catalog that wraps org-id.guide where applicable and adds CommonGrants-specific metadata (schema URLs, kind, lookup templates). A documented promotion flow moves a registry from `otherIds` to a base identifier on the relevant model.
```
jcrichlake, line 107: Should we use future tense here since I don't believe we have a documented promotion flow at this juncture?

jcrichlake: Disregard, I read further

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+
+1. **Proposal.** An adopter or maintainer opens a GitHub issue with the proposed code, scope, kind, example values, and relevant models.
+2. **Review.** Reviewers evaluate: is the registry coherent and documentable? Does it have a named issuer? Can values be validated?
+3. **Catalog entry.** On acceptance, a PR adds a catalog file (`commongrants.org/registries/<code>.yaml`) with the alias, scope, kind, schema, and lookup template. Adopters continue to publish the identifier under `otherIds`, using the cataloged alias.
+
+_Stage 2: Promote to a base identifier (high threshold)_
+
+Moving an identifier out of `otherIds` and into a base identifier slot requires a minor release of `@common-grants/core`, because it adds a reserved key to that model's `IdentifierCollection` extension that every CommonGrants-compliant system is then expected to recognize. Promotion should be reserved for identifiers with demonstrated broad adoption, cross-platform value, and a stable authoritative issuer.
+
+1. **Promotion proposal.** A maintainer proposes promoting a cataloged identifier to a base identifier on one or more models, including evidence of broad use and cross-platform value.
```
jcrichlake: How and where do they propose this? Is this a template on the website? Is this a GitHub issue?

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+2. **Model extension.** On acceptance, the alias is added as a base identifier on the relevant model's `IdentifierCollection` extension (e.g. `ein`, `uei` on `OrgIdentifierCollection`) and the identifier-specific schema is wired up in `@common-grants/core`.
```
jcrichlake: Is this done by the proposing party or by the maintainer of simpler-grants-gov at the time?

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+      description: "{registry}:{id}, for use in query parameters"
+    schema:
+      type: string
+      format: uri
+      description: JSON schema for validating the identifier format
+    uri:
+      type: string
+      format: uri
+      description: Link to this record in the registry's lookup service
+    verifiedAt:
```
jcrichlake, line 255: Would having a `verifiedBy` field also help?  I'm wondering if this could help folks sort out issues if someone is incorrectly verifying ids

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+
+  </TabItem>
+</Tabs>
+
+### Cross-system lookup route
+
+A new query parameter on list routes supports cross-system lookups without requiring consumers to know the local UUID:
+
+```
+GET /organizations?identifier=US-EIN:12-3456789
```
jcrichlake: A few questions come up for me here.  I'm presuming all of this information is open and free to the public.  In the case where we may have sensitive data being passed, should this get moved to a header instead? Also are we familiar with the various rules and regulations that we would be expected to comply with to ensure we don't step out of bounds?

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+A pure UUID is best if:
+
+- we want standard REST behavior and stable references that outlive identifier changes,
+- but we're willing to rely on the identifier collection for human-readable context.
+  :::
+
+**Example**
+
+```
+GET /organizations/01912a8b-7c3d-7890-abcd-ef1234567890
```
jcrichlake, line 463: Can we put a comment after this GET to explain what the specific ID being passed is?

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+#### Option 2: Prefixed UUID (also known as TypeID)
+
+:::note[Bottom line]
+A prefixed UUID (also called a [TypeID](https://github.com/jetify-com/typeid)) is best if:
+
+- we want resource IDs that are self-describing in logs and error messages,
+- but we're willing to give up the standard UUID tooling that most databases and libraries provide.
+  :::
+
+For organizations, the prefix would be `org_`. Adopting this would be a protocol-wide decision (Opportunity, Competition, Form, Application, Award, etc. all have UUIDs today), so it's flagged here but deferred to a future ADR that can apply the convention consistently across all models.
```
jcrichlake, line 490: It's kind of implied here but this would require changing all of those models and most likely some form of validation on the id for each of the other objects in the protocol

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+const ein = org.identifiers.find((i) => i.scheme === "US-EIN");
+```
+
+- **Pros**
+  - Mirrors org-id.guide's `scheme` + `id` model directly.
+  - Well-established in open data standards (OCDS, 360Giving, IATI).
+- **Cons**
+  - Accessing a specific identifier requires filtering by scheme.
+  - Per-registry validation is hard in a mixed array because JSON schema can't branch on `scheme` cleanly.
+  - Multiple IDs per registry appear as duplicate entries with no clear primary.
```
jcrichlake, line 627: This decision seems so one sided that I wonder if it needs walked through here vs some comment on the schema/sdk object.  You can leave it in but it's something to consider

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+    "verifiedAt": "2026-03-15T00:00:00Z"
+  }
+}
+```
+
+The CommonGrants registry catalog carries the rest:
+
+```yaml
+# commongrants.org/registries/us-ein.yaml
+code: US-EIN
```
jcrichlake, line 798: A little confused by this example. There are fields that exist in the json example that also exist in the yaml below. I wouldn't think that would be the case when it says the registry catalog carries the rest.  

Can you put a note explaining how the fuller catalog and the registry metadata can get linked to each other. Say I wanted to look up a `source` field for a specific registry metadata object in the full catalog. What field would I use to do that?

website/src/content/docs/governance/adr/0023-org-ids.mdx
```diff
@@ -0,0 +1,1127 @@
...
+</Tabs>
+
+### Cross-system lookup route
+
+A new query parameter on list routes supports cross-system lookups without requiring consumers to know the local UUID:
+
+```
+GET /organizations?identifier=US-EIN:12-3456789
+GET /opportunities?identifier=US-CFDA:93.575
+GET /applications?identifier=fluxx:app:app-98765
```
SnowboardTechie: Quick question on `compositeId` parsing: since some registry codes here contain `:` (`grants.gov:org`, `candid:bridge`, `fluxx:app`), splitting a composite string back into `(registry, id)` is ambiguous without prior knowledge of which codes are valid. For example, `fluxx:app:app-98765` could parse as `fluxx` + `app:app-98765` or `fluxx:app` + `app-98765`.

Would it be worth pinning a convention under [Registry code format](#registry-code-format), something like "the last `:` is the delimiter" or "consumers resolve the registry half against the catalog"? Happy to file it as a separate issue instead if you'd rather not expand this ADR.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Fair point! And per our conversation in today's huddle I made some changes in 8880d21. The new nested `registry` object on each Identifier now includes a required `url` field pointing to the catalog entry, so the relationship between inline and catalog fields is explicit rather than implied.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Good catch! And addressed in fe0ba41. The primary `id` value now always appears as one of the entries in `allIds`, so consumers can do membership checks against `allIds` alone without also inspecting `id` separately.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Good question, and it made me realize that this section was being prematurely specific about the flow. I changed this section in 7607021.

I wound up modifying the promotion flow to focus less on a specific series of steps that people should follow (since we don't know those exact steps yet) and more to focus on the concepts behind promotion, highlighting phase 1 as being community-driven, and phase 2 as being maintainer-driven.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Good question, also addressed in 7607021. Stage 1 review is done by a maintainer. Stage 2 is entirely maintainer-driven for now. Ideally, a future ADR or an update to `CONTRIBUTING.md` will define an RFC process for community-driven promotions. That's something we'll likely tackle in https://github.com/HHS/simpler-grants-gov/issues/8774

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: I like the idea behind this, but there's some nuance in how we approach it for implementation.

For now, I added `verifiedBy` as an optional string in d585fb1, with guidance that it should reference the system or service doing the verification (e.g. `irs.gov`, `candid.org`, `system:manual-review`) rather than a human name or email (to avoid transmitting PII unexpectedly).

A few questions I'd want to resolve before we commit to having a `verifiedBy` property:

- Format: URI, handle, namespaced identifier, or free string?
- How do we prevent PII from creeping in beyond a SHOULD NOT in the docs?
- Is there value in a canonical vocabulary, or do we let adopters pick?

These feel like implementation details the ADR doesn't have to finalize exactly. There are a handful of similar open questions that will become clearer once we start actually building against this, and I'd rather address them as a set than bake decisions into the protocol prematurely.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Good call-out. Two things:

1. Added a scoping callout in 20a4db4 that makes explicit that `identifiers` is intended for publicly available IDs only (EIN, UEI, CFDA, etc.) and that sensitive/private IDs are out of scope.
2. The broader question about whether any of this should move to headers or POST bodies I noted in this [slack thread](https://betagrantsgov.slack.com/archives/C05TSB1D1SP/p1776965432427859) since it touches more than just this ADR. Deferring to a subsequent decision.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Great catch! Based on our conversation in today's huddle we decided to avoid the composite ID entirely.

Addressed in 48add0e: cross-system lookup now uses separate `registry` and `id` query parameters (`?registry=US-EIN&id=12-3456789`), which sidesteps the parsing issue completely. `compositeId` has been removed from the Identifier schema.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Good idea, addressed in 223ba68. Added inline comments distinguishing the system-UUID lookup from the cross-system registry-scoped lookup.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Good point, I strengthened that bullet in 223ba68 to call out explicitly that prefixed UUIDs would need validation updates across every model with an `id` field (Opportunity, Competition, Form, Application, Award, etc.), not just the model change itself.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: We chatted about this in huddle today. Decided to leave it in for now since documenting the one-sided comparison is useful for transparency, but happy to revisit if more tradeoffs on the array-vs-map question surface later.

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001: Yeah I can see how this might be a tad confusing. Addressed in 8880d21.

The inline and catalog fields should have a bit of clearer relationship:
- Registry-level facts live inside a nested `registry` object on the `Identifier` record, and a required `registry.url` points directly at the corresponding catalog entry.
- The nested `registry` object is meant to be a subset of the catalog entry (e.g. `code`, `scope`, `kind`, `schema`) that are more relevant at the record level too.

As we work on implementation, we can decide if we want to strip more of these nested attributes out, or conversely pull more of them into the `Identifier.registry` object, but at least we now have the `registry.url` connecting the two which should make the relationship and the lookup easier.

SnowboardTechie (approved):
LGTM prior feedback all handled and the open questions around exact field shapes feel right to refine during implementation rather than pin down here. 👍🏻

jcrichlake:
Well done! Thanks for getting those comments addressed quickly and thoroughly!!

jcrichlake (approved):

PR #775, opened 2026-04-24
[Issue #715] form migration

### LINKED ISSUE #715: [Form library] Migrate remaining hand-written forms to TypeSpec composition
### Summary

Migrate the remaining four hand-written forms to TypeSpec composition under `website/src/specs/forms/`, following the pattern established for `key-contact` in PR #713.

Forms to migrate (still hand-written under `website/src/content/forms/<form>/`):

- `sf424-mandatory` — composes Org details, POC, AOR; depends on the AJV fix (#714) since it composes multiple QB types with shared downstream `$ref`s
- `sf424` — same shape as `sf424-mandatory`; same AJV dependency
- `sf424-individual` — single-person form (no separate org); may avoid the shared-`$ref` issue but should be validated
- `cd511` — small AOR-only signature form; likely avoids the AJV issue and can ship sooner

A draft `sf424-mandatory.tsp` exists at `/tmp/sf424-mandatory.tsp.stash` from the form-library exploration; pick it up as a starting point.

### Context

- Each form should be authored as a TypeSpec model that composes question-bank questions and applies field-level `x-overrides` to match the canonical Grants.gov PDF for that form
- Field-level overrides are the preferred authoring shape (see `website/src/specs/forms/README.md`)
- Parity with the legacy hand-written forms can be checked against the cross-form gap matrix produced earlier in this issue (locally generated; not committed)
- The example generator may need the AJV fix to render the larger forms; the small ones (cd511, possibly sf424-individual) may render today

### Suggested PR sequence

1. `cd511` — smallest, likely no AJV dependency
2. `sf424-individual` — second smallest
3. `sf424-mandatory` — full SF-424 mandatory with all sections
4. `sf424` — sibling of mandatory; share most of the spec

### Acceptance criteria

- [ ] `website/src/specs/forms/cd511.tsp` exists; renders at `/forms-new/cd511`; matches the canonical CD-511 form (use the legacy hand-written version as reference)
- [ ] `website/src/specs/forms/sf424-individual.tsp` exists; renders at `/forms-new/sf424-individual`; matches the canonical SF-424 (Individual) PDF
- [ ] `website/src/specs/forms/sf424-mandatory.tsp` exists; renders at `/forms-new/sf424-mandatory`; matches the canonical SF-424 (Mandatory) PDF
- [ ] `website/src/specs/forms/sf424.tsp` exists; renders at `/forms-new/sf424`; matches the canonical SF-424 PDF
- [ ] Each form's parity with the legacy version is documented (field check or screenshot diff)

jcrichlake:
### Summary

  - Fixes #715 
  - Time to review: 15 minutes

  ### Changes proposed

This PR adds new pathway for composing forms, currently this is being developed in parallel to the hand built forms until the process is vetted and approved. Once it is we will transition these forms over to the existing location of the hand built forms and remove the older forms. 

The new pathway is to use questions from the question bank to compose the form fields whenever possible to avoid having duplicate fields across several different fields. This is especially important for common fields like name and address. 

  ### Context for reviewers

  Compare `/forms-new/` forms  against `/forms/` forms to verify the old form against the new.  This applies to all 4 tabs on each form page. 

  To verify:
  1. `cd website`
  2. `pnpm typespec` — should compile with no warnings
  3. `pnpm test` — all 60 tests should pass
  4. `pnpm dev` to start the website to compare

  ### Additional information

CD511
Old
image

NEW

image

SF424 Individual

Old

pt1.
image

pt2.

image

pt3.

image

New

pt1. 

image

pt2.
image

pt3.
image

SF424 Mandatory

Old

pt1.
image

pt2.

image

pt3.

image

pt4.

image

pt5.
image

New

pt1.

image

pt2.

image

pt3.

image

pt4. 

image

pt5. 

image


website/src/lib/schema/strip-ids.ts
```diff
@@ -1,16 +1,24 @@
...
 export function stripIds<T>(schema: T): T {
   return JSON.parse(
-    JSON.stringify(schema, (key, value) =>
-      key === "$id" && typeof value === "string" ? undefined : value,
-    ),
+    JSON.stringify(schema, (key, value) => {
+      if (key === "$id" && typeof value === "string") return undefined;
+      if (key === "unevaluatedProperties") return undefined;
+      return value;
+    }),
```
widal001, line 15: If this is an attempt to remove the `unevaluatedProperties: not: {}` a better way to do this is to change [the emitter settings](https://typespec.io/docs/emitters/json-schema/reference/emitter/#seal-object-schemas)

You should be able to pass that in just for the extensions command in `package.json` instead of changing it in `tsconfig.json`

website/src/specs/question-bank/generics/question-address.tsp
```diff
@@ -139,23 +139,35 @@ enum USState {
 })
 model QuestionAddress {
   /** The primary street address line */
+  @minLength(1)
+  @maxLength(55)
   street1: string;
```
widal001: We don't necessarily want to apply these limits to every use of this question. I'd prefer to just leave them out for now.

website/src/components/forms-new/FormDetails.astro
widal001, line 1: This was an oversight in my original PR, but can we add a section above this FormDetails  on the `FormDetails.astro` page that includes the summary like the current pages do?

Image

website/src/lib/schema/strip-ids.ts
jcrichlake: Fixed here: 

https://github.com/HHS/simpler-grants-protocol/pull/775/changes/6c7b25a66785a0ec3fdad3da9d1f966dcdc4a494

SnowboardTechie (commented):
Looks great overall. Side-by-side parity strategy makes review easy, the `walkGroups` resolution is a nice touch, and the schema tab is much more readable now.

website/src/specs/forms/sf424-mandatory.tsp
```diff
@@ -1,28 +1,849 @@
...
+
+  /** Specify the type of submission (required if typeOfSubmission is Other) */
+  @example("")
+  @minLength(0)
+  @maxLength(50)
+  otherSubmissionTypeExplanation?: string;
+
+  /** 1.b. Frequency */
+  @example(Sf424SubmissionType.Application)
+  frequency: Sf424SubmissionType;
```
SnowboardTechie: Small thing: `frequency` is typed as `Sf424SubmissionType`, so the dropdown will show "Application / Preapplication / …" rather than frequency values. Probably wants its own `Sf424Frequency` enum sourced from the legacy form's values.

website/src/specs/forms/sf424-mandatory.tsp
```diff
@@ -1,28 +1,849 @@
...
+            condition: #{
+              scope: "#/properties/contactPhone/properties/countryCode",
+              schema: #{},
+            },
+          },
+        },
+      },
+    }
+  )
+  contactPhone: QuestionPocPhone;
```
SnowboardTechie, line 615: Heads up that the `extension` and `isMobile` HIDE rules both point `condition.scope` at `.../properties/countryCode` instead of their own scopes (same pattern in the `aorPhone` block around line 821). Doesn't break anything today since `schema: {}` matches anything regardless, but if those condition schemas ever get tightened it'll fire weirdly. The same blocks in `sf424.tsp` and `sf424-individual.tsp` are self-scoped, so aligning these to match should do it.

website/src/lib/forms/overrides.ts
```diff
@@ -63,16 +129,33 @@ export function applyUiOverrides(
...
     const scope = pathToScope(dottedPath);
     const control = controlsByScope.get(scope);
-    if (!control) {
-      throw new Error(
-        `x-overrides.uiSchema path "${dottedPath}" (scope "${scope}") ` +
-          `does not match any Control in the base UI schema.`,
-      );
+    if (control) {
+      Object.assign(control, patch);
+      continue;
```
SnowboardTechie: Small readability nit: both branches do the same `Object.assign(target, patch)`, so the `continue` chain could collapse into:

```ts
const target = controlsByScope.get(scope) ?? groupsByChildPrefix.get(scope);
if (!target) {
  throw new Error(
    `x-overrides.uiSchema path "${dottedPath}" (scope "${scope}") ` +
      `does not match any Control or Group in the base UI schema.`,
  );
}
Object.assign(target, patch);
```

Reads a bit more like "find target, error or assign" and a future third lookup just slots into the `??`. Totally optional.

website/src/lib/forms/loader.ts
```diff
@@ -169,7 +170,9 @@ function extractSchemaData(
   const composedUi = explicitUi ?? composeUiSchema(base.properties);
   const composedFromCg =
     explicitFromCg ?? composeMappingFromCg(base.properties);
-  const composedToCg = explicitToCg ?? composeMappingToCg(base.properties);
+  const baseToCg = composeMappingToCg(base.properties);
+  if (explicitToCg) deepMergeInto(baseToCg, explicitToCg);
+  const composedToCg = baseToCg;
```
SnowboardTechie: Tiny nit, the `baseToCg` rename can drop:

```ts
const composedToCg = composeMappingToCg(base.properties);
if (explicitToCg) deepMergeInto(composedToCg, explicitToCg);
```

website/src/components/forms-new/FormDetails.astro
```diff
@@ -32,16 +27,41 @@ const exampleJson = await generateSchemaExample(schemaRelPath, {
...
-
 const hasOverrides =
   Object.keys(item.overrides.uiSchema ?? {}).length > 0 ||
   Object.keys(item.overrides.mappingFromCg ?? {}).length > 0 ||
   Object.keys(item.overrides.mappingToCg ?? {}).length > 0;
+
+// Strip x-* extension fields and TypeSpec anchor fields from the raw schema
+// so the Schemas tab shows a clean, standard JSON Schema. Also reorder
+// top-level keys to match the conventional shape (title → description → type).
+function stripExtensions(value: unknown, depth = 0): unknown {
```
SnowboardTechie: Couple of small things if you're back in here:

- The `depth` parameter is incremented but never read.
- Arrays short-circuit out, so any `x-*` keys nested inside `allOf`/`oneOf`/`anyOf` would survive. Not an issue for today's forms, but a one-liner protects future ones: `if (Array.isArray(value)) return value.map(stripExtensions);`
- The conditional spreads on `description`/`type` are redundant since `JSON.stringify` drops `undefined`. Plain `{ title: item.label, description, type, ...rest }` gives the same output.

website/src/specs/forms/sf424-mandatory.tsp
jcrichlake: According to the legacy form these are the correct inputs.  Tagging @widal001 for input to see if he can clarify this

widal001: It's possible this was a bug in the original hand-rolled JSON schema. The authoritative source of truth is [the FID](https://www.grants.gov/forms/form-items-description/fid/712) which has the XML schema and the PDF.

website/src/specs/forms/sf424-mandatory.tsp
```diff
@@ -1,28 +1,950 @@
...
+
+/** Applicant type code */
+enum Sf424ApplicantType {
+  StateGovernment: "A: State Government",
+  CountyGovernment: "B: County Government",
+  CityOrTownshipGovernment: "C: City or Township Government",
+  SpecialDistrictGovernment: "D: Special District Government",
+  RegionalOrganization: "E: Regional Organization",
+  USTerritoryOrPossession: "F: U.S. Territory or Possession",
+  IndependentSchoolDistrict: "G: Independent School District",
```
widal001: I think these are duplicated with the SF424 mandatory. Even if there are 1-2 discrepancies, I'd factor this out to a shared instance.

Part of what we're trying to demonstrate here is how many LOC you can save if you re-use components across forms, so if we can get 90% coverage across both forms but only declare this once that's the main goal of this form library re-work. Not having 100% parity between the legacy forms and our forms (at least at the moment).

website/src/specs/forms/sf424-mandatory.tsp
```diff
@@ -1,28 +1,950 @@
...
+  MadeAvailableForReview: "a. This application was made available to the State under the Executive Order 12372 Process for review on",
+  SubjectButNotSelected: "b. Program is subject to E.O. 12372 but has not been selected by the State for review",
+  NotCovered: "c. Program is not covered by E.O. 12372",
+}
+
+/** Yes/No answer */
+enum Sf424YesNo {
+  Yes: "Yes",
+  No: "No",
+}
```
widal001: We should be doing the same for other fields like this. And any other generic elements that are shared across multiple forms.

widal001 (approved):
IMO this looks good enough to call "Done". There's probably some more consolidation especially around the mapping that we can/should do when we pick this work back up, but this is a nice demonstration that we can flexibly build forms from questions in the question bank and extend them with form-specific fields.

PR #785, opened 2026-04-28
ci: split website workflow into validate + deploy-preview

SnowboardTechie:
### Summary

Runs vitest on website PRs and decouples the Cloudflare preview deploy from check/test/audit outcome.

- Time to review: 10 minutes

### Changes proposed

- `.github/workflows/ci-website-preview.yml`:
  - Rename display name to `CI - Website: Validate and Deploy Preview`.
  - Split the single `build-and-deploy-preview` job into `validate` (install / build / upload-artifact / checks / tests / audit) and `deploy-preview` (download-artifact / wrangler deploy / post preview comment).
  - Run vitest in `validate` via a new `Run tests` step — previously vitest only ran via `ci-catalog-validation.yml`, which triggers on `pnpm-workspace.yaml` / `pnpm-lock.yaml` / `.github/dependabot.yml`, so website-only diffs bypassed tests entirely.
  - Gate `Run all checks`, `Run tests`, and `Audit dependencies` on `!cancelled() && steps.build.outcome == 'success'` so all three run regardless of which one fails first.
  - `deploy-preview` uses `pnpm dlx wrangler@4 deploy` against the uploaded artifact, so the deploy job no longer runs `pnpm install`. The `@4` pin matches the `^4.85.0` devDep in `website/package.json`.
  - Expose `validate.outputs.artifact_ready` from the `Upload build output` step; `deploy-preview` gates on that plus `pull_request && !dependabot`. A build failure blocks the preview; a test/check/audit failure does not.

### Context for reviewers

The job split serves two goals beyond fixing the test gap. Build runs once and the output is shared with `deploy-preview` via artifact, so the deploy job skips rebuild and workspace install. And with the per-step `!cancelled()` gating on `validate`, a lint failure no longer short-circuits tests or audit — all three surface in a single CI run.

Deploy decoupling from test outcome is intentional. Previews stay available for reviewing UI changes on a PR that happens to have an unrelated test or lint regression. If the team prefers strict gating, swapping `needs.validate.outputs.artifact_ready == 'true'` for `needs.validate.result == 'success'` is a one-line change (see the thread with @jcrichlake).

Branch-protection heads-up: the required check name was `build-and-deploy-preview`. After this lands it becomes `validate` and `deploy-preview`. Anyone with branch-protection rules keyed on the old name will need to update the repo settings.

Verified locally with `pnpm --filter website run build && pnpm --filter website run test` (60/60 vitest pass) and `actionlint` on the workflow file. This PR's own CI run is the first exercise of the new structure.

### Additional information

N/A — no UI changes.

### DIFF
--- a/.github/workflows/ci-website-preview.yml
+++ b/.github/workflows/ci-website-preview.yml
@@ -121,7 +121,10 @@ jobs:
         # pnpm dlx fetches wrangler on demand — avoids a full workspace install
         # just to get wrangler on PATH. Pin to major @4 to match the devDep in
         # website/package.json and avoid silent v5+ drift.
-        run: pnpm dlx wrangler@4 deploy --name cg-pr-${{ github.event.pull_request.number }}
+        # --allow-build list: wrangler pulls in workerd/esbuild/sharp, each of
+        # which wants to run a postinstall build; pnpm 10's default rejects
+        # those unless explicitly allowed.
+        run: pnpm --allow-build=workerd --allow-build=esbuild --allow-build=sharp dlx wrangler@4 deploy --name cg-pr-${{ github.event.pull_request.number }}
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
--- a/.github/workflows/ci-website-preview.yml
+++ b/.github/workflows/ci-website-preview.yml
@@ -45,6 +45,9 @@ jobs:
         # linting, formatting, spelling, astro checking
         run: pnpm run checks
 
+      - name: Run tests
+        run: pnpm run test
+
       - name: Audit dependencies
         # Calls the npm bulk advisory endpoint directly as a workaround for
         # pnpm audit 410 errors (pnpm/pnpm#11265). Remove this script and
--- a/.github/workflows/ci-website-preview.yml
+++ b/.github/workflows/ci-website-preview.yml
@@ -79,6 +79,11 @@ jobs:
         run: pnpm run checks
 
       - name: Run tests
+        # Run independently of checks so a real test failure still surfaces
+        # when an unrelated lint/format/spell/astro-check step has failed.
+        # Gated on build success because vitest needs the build outputs and
+        # the install tree.
+        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
         run: pnpm run test
 
       - name: Audit dependencies
--- a/.github/workflows/ci-website-preview.yml
+++ b/.github/workflows/ci-website-preview.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     paths:
       - website/**
-      - .github/workflows/ci-website.yml
+      - .github/workflows/ci-website-preview.yml
       - .github/workflows/ci-website-cleanup.yml
       - pnpm-lock.yaml
 
--- a/.github/workflows/ci-website-preview.yml
+++ b/.github/workflows/ci-website-preview.yml
@@ -17,13 +17,14 @@ env:
   CLOUDFLARE_WORKERS_DOMAIN: billy-daly.workers.dev
 
 jobs:
+  # Install, build, and run checks/tests/audit. All post-build steps are
+  # sequential but each runs regardless of the previous step's pass/fail so
+  # lint, test, and audit failures all surface in a single CI run.
   validate:
     runs-on: ubuntu-latest
     outputs:
-      # Captured from the upload step (not the build step) so deploy-preview
-      # only attempts to download when the artifact is actually present. The
-      # output is read regardless of validate's overall outcome, so the deploy
-      # can run even when a later step (checks/tests/audit) failed.
+      # Read from the upload step (not the build step) so deploy-preview only
+      # tries to download when the artifact actually exists.
       artifact_ready: ${{ steps.upload-build.outcome == 'success' }}
 
     steps:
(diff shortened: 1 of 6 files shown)


jcrichlake:
I just want to confirm, does this block deploys if tests are failing? I kind of think we shouldn't generate a preview environment either in that case.

SnowboardTechie:
@jcrichlake no, deploy isn't blocked on test failures. The gate is `artifact_ready` (build succeeded and the artifact uploaded), so a build failure stops the preview but a test, check, or audit failure doesn't.

This was intentional though, I've found preview environments useful on prior repos for visually confirming what a failing test is catching, and for reviewing UI changes on a PR that happens to have an unrelated test regression.

The argument for blocking is real: a passing preview on top of failing tests can mask a regression the reviewer might otherwise notice. If you'd rather make test-pass a hard gate, the change is one line, swap `needs.validate.outputs.artifact_ready == 'true'` for `needs.validate.result == 'success'`.

Leaning toward keeping it decoupled given prior experience with this pattern, but open to the stricter gate if you'd prefer it.

.github/workflows/ci-website-preview.yml
```diff
@@ -38,29 +44,105 @@ jobs:
...
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+          cache: "pnpm"
+
+      - name: Install dependencies
+        working-directory: .
+        run: pnpm install --frozen-lockfile
```
widal001: You might need to check me on this, but if we already have the build artifact I think we can just change the deploy preview command to `pnpm dlx wrangler deploy ...` and we can skip installing dependencies, which should save us some time.

.github/workflows/ci-website-preview.yml
```diff
@@ -38,29 +44,105 @@ jobs:
...
+      - name: Upload build output
+        # Uploaded immediately after build so a later failure (checks/tests/
+        # audit) doesn't suppress the artifact deploy-preview depends on.
+        #
+        # Path-structure invariant: every entry shares `website/` as common
+        # parent, so upload-artifact stores them with that prefix stripped
+        # (artifact root = website/). The deploy-preview download then uses
+        # `path: website` to restore them at website/dist, website/cache, etc.
+        # Adding a path outside `website/` here will shrink the common prefix
+        # and silently shift download locations — keep new paths under website/.
```
widal001: @SnowboardTechie  Can we try to state this a bit more succinctly and in plain language? Same is true for the other code comments here and the PR description:

I _think_ the CI/CD changes in this PR are trying to do a few main things:

- We're splitting out these CI steps so they can (largely) run in parallel after a build succeeds:
  - checks
  - tests
  - deploy preview
- We're uploading the build artifact so that even as we split out the deploy to a separate step we don't have to rebuild.
- All of these changes are designed to ultimately help us:
  - Surface multiple issues (linting issues, failed tests, etc.) on a single CI/CD run rather than having to wait for a fix to an earlier issue to surface an issue in a subsequent step.
  - Speed up development and (ideally) reduce our GH action runtimes

But the fact that I'm not totally sure and am having trouble reasoning about it suggests that we might need to simplify the PR description and the code comments here.

I'm all for using AI to streamline the creation of PRs, but I've noticed that AI-generated PRs tend to lose the forest for the trees a little bit, over explaining specific implementation details and consequently "burying the lead" (to mix metaphors) of what the PR is actually trying to accomplish.

.github/workflows/ci-website-preview.yml
```diff
@@ -38,29 +44,105 @@ jobs:
...
+
       - name: Run all checks
         # linting, formatting, spelling, astro checking
         run: pnpm run checks
 
+      - name: Run tests
+        # Run independently of checks so a real test failure still surfaces
+        # when an unrelated lint/format/spell/astro-check step has failed.
+        # Gated on build success because vitest needs the build outputs and
+        # the install tree.
```
widal001: FWIW I don't _think_ this is true currently since it's a later step in the same job. To actually have them run independently [the tests would need to run in a separate job](https://github.com/orgs/community/discussions/26291).

If they were run independently and in parallel you would see them listed as multiple jobs in the action summary, but we just see two:

Image

widal001: By contrast [this CI file](https://github.com/agilesix/cg-frontend-astro/blob/4a930c87bb865fa3f08337f54995cc963e0c6ad5/.github/workflows/ci.yml) does produce multiple concurrent runs because it splits these steps to separate jobs:

Image

A notable tradeoff though is that the more jobs you split out CI checks to the faster you run up your GitHub action usage both because you have to install deps multiple times and because each separate job might be rounded up to the minute (but I could be wrong on that)

.github/workflows/ci-website-preview.yml
SnowboardTechie: Good call. Switched to `pnpm dlx wrangler@4 deploy …` and dropped the workspace install from `deploy-preview`. One gotcha: pnpm 10's default rejects postinstall scripts unless allow-listed (`ERR_PNPM_IGNORED_BUILDS`), and `pnpm dlx` doesn't inherit the workspace's approved list, so the command ended up as `pnpm --allow-build=workerd --allow-build=esbuild --allow-build=sharp dlx wrangler@4 deploy …`. Pinned to major `@4` to match the `"wrangler": "^4.85.0"` devDep in `website/package.json`. Net result: deploy-preview dropped from 55s to 39s on the latest run.

.github/workflows/ci-website-preview.yml
SnowboardTechie: Fair. Rewrote the PR description to match the repo's template sections (it was ignoring three of them) and trimmed the workflow comments from ~35 lines to ~10, keeping only the three that document non-obvious behavior: the path-structure invariant on `upload-artifact` (real footgun, adding a path outside `website/` silently shifts download locations), the `pnpm audit` 410 workaround (pre-existing), and the `wrangler@4` + `--allow-build` context.

.github/workflows/ci-website-preview.yml
SnowboardTechie: You're right, "independently" was the wrong word. `!cancelled() && steps.build.outcome == '"'"'success'"'"'` doesn't create parallelism; it just prevents the default short-circuit-on-failure between sequential steps. Replaced the per-step comments with one comment above `validate:` saying exactly that: steps are sequential, but each runs regardless of the previous step's pass/fail so all failures surface in one run.

SnowboardTechie: Yeah, the per-job split is a real option and the tradeoff you name (N× install + minute rounding) is exactly the thing I'd want to measure before committing to it repo-wide. This PR is scoped to "run vitest on website PRs + decouple deploy from test/check/audit outcome," and the same question applies to `ci-lib-core`, `ci-lib-cli`, `ci-lib-sdk`, `ci-lib-python-sdk`, cleaner to evaluate holistically than spot-fix here. Keeping the current single-job-with-gating structure in this PR.

jcrichlake (approved):

PR #803, opened 2026-05-01
docs(adr): #759 resolve drift in mergeExtensions, getClient, filters

SnowboardTechie:
### Summary

- Fixes #759
- Time to review: 5 minutes

### Changes proposed

Three spec-internal drift items in [ADR-0022 plugin framework](https://commongrants.org/governance/adr/0022-plugin-framework/), flagged during [#740](https://github.com/HHS/simpler-grants-protocol/pull/740) review and deferred to land alongside the SDK sub-issues. All three resolved in one .mdx file:

1. **`mergeExtensions` / `merge_extensions` signatures.** Replace the single-literal `onConflict: "firstWins" = "firstWins"` (which can only ever hold one value) with the three-value union `"firstWins" | "lastWins" | "error"` per [Billy's reply on #740](https://github.com/HHS/simpler-grants-protocol/pull/740#discussion_r3127086113). TS switches from `...rest, options` to `extensions: PluginExtensions[], options?: { onConflict? }` to satisfy TS1014 (rest parameter must be last); the example call site at line 473 updates to the new array-arg form. Both signatures carry an inline `// provisional pending SDK implementation (#744)` marker — exact syntax will be pinned by the SDK PR per Billy's deferral framing.

2. **`client` vs `getClient` drift in the Option 3 sketch.** The chosen-option `interface Plugin` (line 149), the plugin-shape prose (line 56), and the Option 3 pros all already use `getClient`. Only the Option 3 sketch (now line 780) lagged with `client?: Client`. Updated for consistency.

3. **Orphan `filters` top-level key.** Decision #2 enumerates four top-level keys (`meta`, `client`, `schemas`, `extensions`), but the runtime `Plugin` types tacked on a fifth `filters` key. Per [Billy's resolution](https://github.com/HHS/simpler-grants-protocol/pull/740#discussion_r3127172197): `filters` belongs on `DefinePluginOptions` / `define_plugin()` as an authoring-time input but does not surface on the runtime `Plugin` shape, because filter handling is encapsulated in the `Client` returned by `getClient()`. Removed `filters?:` from the TS `interface Plugin` and the Python `Plugin` dataclass; left `DefinePluginOptions` / `define_plugin(... filters=...)` untouched. Decision #2 footnote and inline comments above each `Plugin` type document the intentional asymmetry.

### Context for reviewers

[...]

PR #810, opened 2026-05-07
[Issue #799] Transform PoC

### LINKED ISSUE #799: [SDK] Draft transforms PoC: Python
### Summary

Build a Python proof-of-concept for the plugin transformation interface. Validates the contract shape recommended in SPIKE #657 before locking the full implementation in both SDKs.

Issued separately from the TypeScript PoC to avoid SDK dependencies between them.

### Acceptance criteria

- [ ] PoC implements the transformation contract shape recommended in #657
- [ ] PoC demonstrates bidirectional transformation between a sample source format and CommonGrants
- [ ] Findings and design questions captured to inform full SDK implementation

jcrichlake:
### Summary

  - Fixes #799
  - Time to review: 15 minutes

  ### Changes proposed

  Implements a bidirectional transform framework as a proof-of-concept for the plugin system. Plugin authors can now define declarative mapping dicts that compile into `to_common` /
  `from_common` callables, enabling data translation between a source system's native format and the CommonGrants format without writing imperative transform code.

  **New utilities:**
  - `build_transforms()` — compiles a pair of mapping dicts into typed `(to_common, from_common)` callables with structural validation at call time
  - `TransformResult` — unconditional return shape carrying a `result` dict and a list of `PluginError`s (non-fatal; partial result always returned)
  - `PluginError` — structured error type with `path`, `handler`, `source_value`, and `cause` fields for programmatic error handling
  - `ObjectSchemasInput` — bundles `to_common` / `from_common` for a single object type; passed to `define_plugin()` via `transform_schemas`
  - `PluginMeta` — optional metadata attached to a plugin (`name`, `version`, `source_system`, `capabilities`)

  **Built-in mapping handlers** (in `utils/transformation.py`):
  | Handler | Description |
  |---|---|
  | `field` | Extracts a value via dot-notation path |
  | `const` | Returns a fixed literal value |
  | `match` | Case-based lookup (canonical ADR name) |
  | `switch` | Alias for `match` (backward compat) |
  | `numberToString` | Extracts numeric value, coerces to string |
  | `stringToNumber` | Extracts string value, coerces to int/float |

  Custom handlers can be registered per `build_transforms()` call and cannot override built-in names.

  **Sample grants.gov plugin** (`examples/plugins/grants_gov/`): demonstrates the full interface — bidirectional field mapping, status case conversion, funding amounts, key dates, and
  plugin metadata.

  **Documentation** (`extensions/README.md`): new "Bidirectional Transforms" section covering mapping format, handler reference, custom handlers, and roundtrip usage.

  **Tests** (`tests/extensions/`, `tests/utils/`): coverage for `build_transforms`, `TransformResult`, `PluginError`, `ObjectSchemasInput`, `PluginMeta`, and `transform_from_mapping`.

  ### Context for reviewers

  This is a **proof-of-concept** scoped to the transform layer only. It is intentionally limited:

  - `build_transforms()` validates mapping *structure* at call time but cannot validate that field paths resolve against real data (deferred to full SDK with schema introspection).
  - The PoC does not auto-invert one mapping from the other — both directions must be authored explicitly, because handlers like `match` are not reversible.
  - The `TODO (full SDK)` comments in `transforms.py` call out where `model_validate` integration belongs (in `define_plugin()`, which knows the target Pydantic model).

  **Verification:**

  Run the working example end-to-end:
  ```bash
  cd lib/python-sdk
  poetry run python examples/transforms.py

  This executes a grants.gov → CommonGrants → grants.gov roundtrip and prints a PASS/FAIL check for each mapped field.

  Run the test suite:
  cd lib/python-sdk
  poetry run pytest tests/extensions/ tests/utils/ -v

  Additional information

  Sample output from examples/transforms.py:

  ============================================================
  SOURCE DATA (grants.gov format)
  ============================================================
  { "data": { "opportunity_title": "Research into conservation techniques", "opportunity_status": "posted", ... } }

  ============================================================
  to_common: grants.gov → CommonGrants
  ============================================================
  Errors: none

  Result:
  { "title": "Research into conservation techniques", "status": { "value": "open", ... }, "funding": { "minAwardAmount": { "amount": 10000, "currency": "USD" }, ... } }

  ============================================================
  from_common: CommonGrants → grants.gov
  ============================================================
  Errors: none

  ============================================================
  ROUNDTRIP CHECK
  ============================================================
    [PASS] title: 'Research into conservation techniques' -> 'Research into conservation techniques'
    [PASS] status: 'posted' -> 'posted'
    [PASS] award_floor: 10000 -> 10000
    [PASS] award_ceiling: 100000 -> 100000

  Roundtrip result: ALL PASS


lib/python-sdk/common_grants_sdk/extensions/transforms.py
```diff
@@ -0,0 +1,121 @@
...
+    known = set(merged)
+
+    # Validate mapping structure at call time
+    _validate_mapping(to_common_mapping, known)
+    _validate_mapping(from_common_mapping, known)
+
+    def to_common(native: Any) -> TransformResult[Any]:
+        try:
+            result = transform_from_mapping(native, to_common_mapping, handlers=merged)
+            # TODO (full SDK): run model_validate on result and append validation
```
jcrichlake: I'm not sure if we should handle this TODO here or later down the road. Would love thoughts on this.

lib/python-sdk/common_grants_sdk/extensions/transforms.py
```diff
@@ -0,0 +1,145 @@
...
+            # Handler invocation — argument is runtime-only, do not recurse
+            continue
+        _validate_mapping(value, known_handlers, current_path)
+
+
+def build_transforms(
+    to_common_mapping: dict[str, Any],
+    from_common_mapping: dict[str, Any],
+    handlers: dict[str, Handler] | None = None,
+    common_model: type[BaseModel] | None = None,
```
jcrichlake, line 75: Calling this out as a needed change from the existing ADR.  Without having the common_model  in the signature we don't have a way to validate an extended class that contains optional fields.  @SnowboardTechie for awareness

lib/python-sdk/common_grants_sdk/extensions/transforms.py
```diff
@@ -0,0 +1,145 @@
...
+            return TransformResult(result=validated, errors=[])
+        except ValidationError as exc:
+            errors = [
+                PluginError(
+                    e["msg"],
+                    path=".".join(str(loc) for loc in e["loc"]),
+                )
+                for e in exc.errors()
+            ]
+            return TransformResult(result=result, errors=errors)
```
SnowboardTechie, line 161: When `common_model` is set, success returns the validated Pydantic instance (L124) but a `ValidationError` returns the raw dict (L133). Mixed `.result` shape from the same call. Intended, or want it uniform?

lib/python-sdk/common_grants_sdk/extensions/transforms.py
```diff
@@ -0,0 +1,145 @@
...
+    known = set(merged)
+
+    # Validate mapping structure at call time
+    _validate_mapping(to_common_mapping, known)
+    _validate_mapping(from_common_mapping, known)
+
+    def to_common(native: Any) -> TransformResult[Any]:
+        try:
+            result = transform_from_mapping(native, to_common_mapping, handlers=merged)
+        except Exception as exc:
```
SnowboardTechie, line 143: Two small ones on the try/except:

- `except Exception` catches everything, so a `KeyError` in a custom handler comes back as a `PluginError` rather than propagating.
- `handler=None` is hardcoded. ADR Decision #9 lists `handler` as one of the structured fields, but with the catch at this layer there's no way to know which handler raised.

Either worth tightening?

lib/python-sdk/common_grants_sdk/extensions/transforms.py
SnowboardTechie, line 75: Just checking, is the asymmetry deliberate? `common_model` validates `to_common` output. `from_common` has no equivalent parameter or validation pass.

lib/python-sdk/common_grants_sdk/extensions/README.md
```diff
@@ -433,6 +441,140 @@ Prefer unique, namespaced field names so `"error"` is never triggered.
...
+
+### Defining transforms
+
+Use `build_transforms()` to compile a pair of mapping dicts into `(to_common, from_common)` callables, then pass them to `define_plugin()` via `transform_schemas`:
+
+```python
+from common_grants_sdk.extensions import (
+    CustomFieldSpec,
+    ObjectSchemasInput,
+    PluginMeta,
```
SnowboardTechie: Small snag: this import will raise `ImportError` as written. `PluginMeta` is referenced in three places in the new content (this line, the key-concepts table at L49, and the example call at L505) but the package only exports `PluginExtensionsMeta` (types.py L78). Which name should be canonical?

lib/python-sdk/examples/transforms.py
```diff
@@ -0,0 +1,273 @@
...
+
+import json
+from typing import Any
+
+# When run as `poetry run python examples/transforms.py`, Python automatically
+# adds the script's directory (examples/) to sys.path. Import from there using
+# the `plugins.` prefix (not `examples.plugins.`) — the `examples.` prefix only
+# works in -c or interactive contexts where lib/python-sdk/ is sys.path[0].
+from plugins.grants_gov.cg_config import plugin
+from plugins.grants_gov.generated.schemas import Opportunity
```
SnowboardTechie, line 27: Heads up: the PR body's standalone verification command hits ModuleNotFoundError on a fresh clone, since `examples/plugins/grants_gov/generated/` isn't committed and the python call doesn't trigger `make plugins` (which `check-types` does).

```
$ poetry run python examples/transforms.py
ModuleNotFoundError: No module named 'plugins.grants_gov.generated'
```

Same setup as `opportunity_extensions`, not new in this PR, just newly visible. Worth a note in the PR body or examples README.

lib/python-sdk/examples/transforms.py
```diff
@@ -0,0 +1,273 @@
...
+    ]
+    all_pass = True
+    for field, original, roundtripped in checks:
+        ok = original == roundtripped
+        if not ok:
+            all_pass = False
+        status = "PASS" if ok else "FAIL"
+        print(f"  [{status}] {field}: {original!r} -> {roundtripped!r}")
+
+    print(f"\nRoundtrip result: {'ALL PASS' if all_pass else 'SOME FIELDS DIFFER'}")
```
SnowboardTechie: Tiny readability thing. The checks list at L186-208 covers 4 fields, the `from_common` mapping (cg_config L80-101) covers 6, and L181's comment already calls out that out-of-mapping fields drop. The "ALL PASS" print line doesn't carry that context. Worth wording it as the selective check it is?

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -362,10 +377,19 @@ Handler = Callable[[Any, Any], Any]
...
+# generated/schemas.py's Opportunity), not the base class — passing a base class
+# silently weakens validation of typed custom fields.
+# transform_from_mapping normalizes Pydantic model instances to plain dicts via
+# model_dump(mode="json") at the entry point, so from_common can receive the
+# validated model output of to_common and field paths still resolve correctly.
 def build_transforms(
     to_common_mapping: dict[str, Any],   # ADR-0017 mapping from native → CommonGrants
     from_common_mapping: dict[str, Any], # ADR-0017 mapping from CommonGrants → native
     handlers: dict[str, Handler] | None = None,
+    common_model: type[BaseModel] | None = None,  # must be the generated extended model, not the base
```
SnowboardTechie, line 396: Worth flagging before this lands: this PR adds `commonModel?` / `common_model` to `buildTransforms()` in the ADR, but Decision #7's prose is unchanged and still reads "`definePlugin()` wraps the underlying transform output with runtime schema validation." Prose and signature now describe two different validation boundaries.

lib/python-sdk/tests/utils/test_transformation.py
```diff
@@ -233,6 +234,108 @@ def handle_type(data, type_spec):
     assert result == {"id_str": "12345"}
 
 
+def test_const_string(input_data):
```
SnowboardTechie, line 240: Heads up on something pre-existing. L189 (`test_extend_with_concat`) writes `"concat"` into the module-level `DEFAULT_HANDLERS` without restoring. The new `test_handler_collision_raises` doesn't care today (its parametrize values collide regardless), but `build_transforms` now reads `set(DEFAULT_HANDLERS)` at runtime for its collision check. So a future test registering `"concat"` as a custom handler would falsely fail with "collide with defaults" depending on test order.

Easy unbreak is the `handlers={**DEFAULT_HANDLERS, ...}` pattern that `test_extend_with_type_conversion` already uses.

lib/python-sdk/common_grants_sdk/extensions/types.py
```diff
@@ -0,0 +1,147 @@
...
+
+    custom_fields: custom field declarations (merged by merge_extensions).
+    mappings: optional ADR-0017 declarative mappings. When present and no explicit
+        to_common / from_common is supplied in schemas[obj], define_plugin() will
+        auto-invoke build_transforms() on these (TODO — ADR-0022 Decision #6).
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    custom_fields: dict[str, Any] | None = Field(default=None, alias="customFields")
```
SnowboardTechie: Small thing: `custom_fields` is typed here as `dict[str, Any]`, but the legacy `SchemaExtensions` (specs.py) types the same data as `dict[str, CustomFieldSpec]`. Tightening to match would let the type checker catch shape errors.

lib/python-sdk/common_grants_sdk/extensions/transforms.py
jcrichlake: I updated the doc string to better explain this behavior

lib/python-sdk/common_grants_sdk/extensions/transforms.py
jcrichlake: Yes this is deliberate, but still a great question! The idea here is that the `to_common_mapping` will take an external and unknown data shape and transform it into either a standard common grants spec object OR it gets turned into a standard common grants spec object with custom fields attached.  In the latter case we want the ability to call pydantic `model_validate()` against  not just the base object but we also want the deeper validation for the custom fields to surface any errors within those fields due to bad mappings. 

We can't do the same for `from_common_mapping` since we wouldn't have the pydantic model for the external data shape

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -150,7 +229,10 @@ def transform_node(node: Any, depth: int) -> Any:
             # Returns: `extract_field_value(data, "opportunity_status")`
             if k in handlers:
                 handler_func = handlers[k]
-                return handler_func(data, v)
+                try:
+                    return handler_func(data, v)
+                except Exception as exc:
+                    raise HandlerError(k, exc) from exc
```
SnowboardTechie, line 241: Found something subtle worth surfacing before this ships, especially now that `build_transforms` is going public and `transform_from_mapping` is documented in the SDK README. The `for k, v in node.items():` loop just above returns after its first iteration regardless of whether the first key was a handler, so a handler key with sibling keys at the same level gets passed through as a literal output dict instead of being invoked:

```python
data = {"x": "VALUE-OF-X"}
mapping = {"metadata": "literal-prefix", "field": "x"}
transform_from_mapping(data, mapping)
# Expected: "VALUE-OF-X"
# Actual:   {"metadata": "literal-prefix", "field": "x"}
```

The output looks plausible enough that I think it could go unnoticed for a while. A few directions you could take it: reject handler-with-siblings at runtime, reject it earlier in `_validate_mapping` (which today `continue`s past handler keys without checking for siblings), or rework the loop to process all keys. Curious how you'd want to handle it.

lib/python-sdk/common_grants_sdk/utils/transformation.py
```diff
@@ -64,10 +64,78 @@ def switch_on_value(data: dict, switch_spec: dict) -> Any:
...
+    if val is None:
+        return None
+    s = str(val)
+    try:
+        return int(s)
+    except ValueError:
+        return float(s)
+
+
+class HandlerError(Exception):
```
SnowboardTechie: Worth flagging that introducing `HandlerError` as its own exception class is a behavior change for direct `transform_from_mapping` callers and anyone using `CommonGrantsBaseModel.dump_with_mapping` / `validate_with_mapping`. A bad `stringToNumber` input that used to raise `ValueError` now raises `HandlerError`, and `transform_from_mapping` is documented in lib/python-sdk/README.md so external callers are plausible.

Worth weighing whether you'd rather make `HandlerError` extend `ValueError` (keeps existing `except ValueError` paths working for free), audit and migrate known callers, or call it out as a deliberate break. No strong opinion from my side, just want to make sure it's a conscious choice.

lib/python-sdk/tests/extensions/test_transforms.py
```diff
@@ -0,0 +1,200 @@
...
+
+    to_common, _ = build_transforms(
+        {"title": {"boom": "anything"}},
+        {},
+        handlers={"boom": boom},
+    )
+    result = to_common(SOURCE_DATA)
+    assert len(result.errors) == 1
+    assert isinstance(result.errors[0], PluginError)
+    assert "handler exploded" in str(result.errors[0])
```
SnowboardTechie: Small suggestion on `test_exception_surfaces_as_plugin_error_not_raised`: adding asserts on `result.errors[0].handler == "boom"` and `isinstance(result.errors[0].cause, RuntimeError)` would lock in the new attribution path. Without them, a future `except Exception` fallthrough could regress the attribution and the test would still pass.

If the handler-with-siblings thread leads to a new validation rule, a test covering that shape would be a natural addition here too.

SnowboardTechie (commented):
Nice work on round 2. 7 of my 9 earlier findings are fully addressed, and the other 2 are documented-as-intended. The `HandlerError` refactor in particular reads much cleaner than the hardcoded `handler=None` we started with.

One latent bug worth surfacing before this ships: `transform_node` silently treats handler keys as literals when a non-handler sibling iterates first. The fact that `build_transforms` and `transform_from_mapping` are both public surface now makes it more load-bearing than before. Reproducer in the inline on utils/transformation.py.

Two small follow-ups on the round-2 fix itself, both flagged on individual threads: `HandlerError` is a behavior change for `dump_with_mapping` / `validate_with_mapping` callers (worth deciding whether to make it `ValueError`-compatible), and the new attribution path is not covered by an assertion in the exception test.

lib/python-sdk/common_grants_sdk/extensions/transforms.py
SnowboardTechie: The `HandlerError` shape is great, much cleaner than what we started with. One small thing worth a look: `test_exception_surfaces_as_plugin_error_not_raised` asserts the message text survives but does not pin down `result.errors[0].handler` or `cause`. Might be worth tightening so the new attribution path stays covered.

lib/python-sdk/common_grants_sdk/extensions/transforms.py
SnowboardTechie: Got it, that makes sense. The native side does not have a canonical Pydantic model to validate against, so there is nothing to mirror `common_model` for. Worth adding a sentence along these lines near the `common_model` param description in the docstring so future readers have the rationale at hand.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -188,25 +193,33 @@ interface DefinePluginOptions {
...
-  options?: { onConflict?: "error" | "firstWins" | "lastWins" }, // defaults to "error"
+  ...extensions: PluginExtensions[],
+  onConflict: "firstWins" = "firstWins",
 ): PluginExtensions;
 
 // Handler signature matches ADR-0017 runtime conventions.
 type Handler = (value: unknown, context: unknown) => unknown;
 
-// Utility: generates toCommon and fromCommon functions from separate declarative
+/// Utility: generates toCommon and fromCommon functions from separate declarative
```
SnowboardTechie:
```suggestion
// Utility: generates toCommon and fromCommon functions from separate declarative
```

lib/python-sdk/.coverage
SnowboardTechie, line 1: Not sure we want this commited?

SnowboardTechie (approved):
Round-3 looks great! The ADR has some drift against #803 in the merge_extensions / filters area, but IMO we should rebase `main` onto current `HOLD-transforms` after this merges.

Approving w/ 2 small suggests: `git rm lib/python-sdk/.coverage` and add `.coverage` to either `lib/python-sdk/.gitignore` or the root. It looks like Pytest-cov SQLite output landed in the tree. And another trimming an extra `/`.

PR #816, opened 2026-05-11
ci(deps-catalog): respect semver and split majors into draft PR

SnowboardTechie:
### Summary

Stop the weekly catalog-check workflow from bumping past major-version boundaries (which has broken `main` four of the last five weeks via zod 3→4, typescript 5→6, `@types/node` 20→25, etc.). Within-major bumps continue to flow into the existing auto-PR; cross-major bumps now go to a separate draft PR labeled `DO NOT MERGE` so they stay visible without blocking the safe updates.

- Time to review: 8 minutes

### Changes proposed

- `.github/scripts/update-catalog-deps.sh`: bumps now stay within a dep's compatibility boundary by default — `^X.Y.Z` (X≥1) within major X, `^0.Y.Z` within the 0.x line, `~X.Y.Z` within minor X.Y. A new `--majors-only` flag inverts the filter to surface cross-boundary bumps. Majors-only install uses `--ignore-scripts` so the lockfile regenerates cleanly even when the workspace `prepare` step would fail under the major bump.
- `.github/workflows/deps-catalog-check.yml`: trimmed to the within-major pass. Continues to produce the existing ready-for-review PR on `chore/update-catalog-deps`.
- `.github/workflows/deps-catalog-check-majors.yml` (new): majors-only pass. Produces a **draft** PR on `chore/update-catalog-deps-majors` labeled `DO NOT MERGE` that surfaces available majors for human-led migration. Runs on its own runner, so a failure here can't block the within-major pass.

### Context for reviewers

**Why.** The scheduled catalog-check has failed 4 of the last 5 weekly runs. Each failure looked the same: the script bumped a dep to a new major (zod 3→4 renamed `ZodError.errors`, typescript 5→6 deprecated `baseUrl`, `@types/node` 20→25, etc.), the workspace `prepare` step then broke during `pnpm install`, and the workflow exited non-zero before reaching `peter-evans/create-pull-request`. Collateral damage: every safe within-major update that week (typespec minors, prettier patches, typescript-eslint patches) got dropped along with the broken major.

**New behavior.**

1. Within-major bumps flow through to the existing ready-for-review PR with no breaking-change risk.
2. Cross-major bumps go to a separate draft PR. CI on that PR is expected to fail until a human drives the migration — that failing CI is the signal it needs attention. Reviewers cherry-pick a single major into a focused migration PR, or pull the draft branch and chip through it. Closing the draft just defers the work — it re-appears on the next Monday's run with whatever majors are still pending.
3. Each Monday both branches force-update against `main` with the current state, so neither goes stale.

**Residual risk we're accepting.** `^0.Y.Z` minor bumps (typespec `^0.80.0 → ^0.81.0`) flow into the within-major PR. If a future typespec minor breaks the build, the within-major PR's CI fails visibly — same bounded outcome as the majors draft, no `main` impact. No past failures from this path. Not worth pre-emptive per-dep machinery.

**Verification.**

- Within-major dry-run identifies 11 in-range bumps; typescript ^5, zod ^3, eslint ^9, vitest ^3 correctly held.
- Within-major full-apply: `pnpm install` plus all four workspace `prepare` steps complete successfully.
- Majors-only dry-run identifies 9 cross-boundary candidates. Full-apply with `--ignore-scripts` regenerates the lockfile cleanly. Peer-dep warnings (e.g. eslint-plugin-vitest 0.5.4 vs eslint 10) correctly surface the work a reviewer will need to do.
- `pull_request` trigger smoke-tests the updated script end-to-end against itself in both workflow files (each lists the shared script in its `paths:` filter).

### DIFF
--- a/.github/workflows/deps-catalog-check.yml
+++ b/.github/workflows/deps-catalog-check.yml
@@ -34,8 +34,11 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
-      - name: Check for outdated catalog deps
-        id: check
+      # ── Pass 1: within-major bumps ────────────────────────────────────────
+      # Safe updates that almost always work. Auto-mergeable after CI passes.
+
+      - name: Check for within-major updates
+        id: check_minor
         run: |
           set +e
           ./.github/scripts/update-catalog-deps.sh --dry-run
@@ -46,16 +49,16 @@ jobs:
           elif [[ "$EXIT_CODE" -eq 0 ]]; then
             echo "has_updates=true" >> $GITHUB_OUTPUT
           else
-            echo "::error::Catalog dep check failed with exit code $EXIT_CODE"
+            echo "::error::Catalog dep check (within-major) failed with exit code $EXIT_CODE"
             exit 1
           fi
 
-      - name: Run catalog update script
-        if: steps.check.outputs.has_updates == 'true'
+      - name: Apply within-major updates
+        if: steps.check_minor.outputs.has_updates == 'true'
         run: ./.github/scripts/update-catalog-deps.sh
 
-      - name: Create pull request
-        if: steps.check.outputs.has_updates == 'true' && github.event_name != 'pull_request'
+      - name: Open within-major PR
+        if: steps.check_minor.outputs.has_updates == 'true' && github.event_name != 'pull_request'
         uses: peter-evans/create-pull-request@v8
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/deps-catalog-check.yml
```diff
@@ -82,9 +88,70 @@ jobs:
...
             - [ ] Add changeset if peerDep ranges changed for `@common-grants/core`
           labels: |
             dependencies
           commit-message: "chore(deps): update catalog dependencies"
           delete-branch: true
+
+      # ── Pass 2: cross-major bumps (manual review) ─────────────────────────
+      # Drafts a separate PR for major bumps so they're visible but obviously
+      # not auto-mergeable. Reset the working tree first so pass 2 sees the
+      # original committed workspace + lockfile, not pass 1's modifications.
```
jcrichlake: Would it be better to split this out into a totally different file?  I'm wondering if that wouldn't avoid some of the complexity with having pass 1 and then pass 2

SnowboardTechie: Good idea, thanks. Split into `deps-catalog-check.yml` (within-major) and `deps-catalog-check-majors.yml` (majors-only) in 6e580f4. Each runs on its own runner, so the `git checkout HEAD --` reset between passes is gone and a failure in one path no longer blocks the other.

jcrichlake (approved):

PR #834, opened 2026-05-14
[chore] Removing releases without tags check

jcrichlake:
### Summary

Simple script update to get rid of the releases without tags check for this workflow.  It wasn't checking as expected and there's not really a great action step to be taken should this scenario actually occur.  It also may not even be possible.


widal001 (approved):
Thanks for making this change!

SnowboardTechie (approved):

PR #825, opened 2026-05-12
[Issue #798] Transforms PoC: TypeScript

SnowboardTechie:
Summary

  Adds a TypeScript proof-of-concept for the plugin transformation framework (issue #798), mirroring the Python PoC from PR #810. Plugin authors can now compile declarative mapping objects into typed toCommon/fromCommon callables, validate output against an extended Zod schema, and attach those callables to a plugin via
  definePlugin({ schemas }). Also includes corresponding updates to the Python SDK to maintain cross-SDK parity, and ADR amendments for the plugin framework (ADR-0022) and three-state null handling (ADR-0024).

  ---
  Changes proposed

  TypeScript SDK (lib/ts-sdk)
  - Added buildTransforms(toCommonMapping, fromCommonMapping, handlers?, commonModel?) — compiles a mapping pair into typed (toCommon, fromCommon) callables with structural validation at call time
  - Added TransformResult<T>, PluginError, transformFromMapping(), getFromPath(), and DEFAULT_HANDLERS — six built-in handlers: const, field, match/switch (alias), numberToString, stringToNumber
  - Updated definePlugin() to accept optional meta: PluginMeta and schemas: SchemasInput; custom fields are now co-located under schemas[Object].customFields rather than the top-level extensions key
  - Removed mergeExtensions from the public surface (consumers should use object spread instead)
  - Implemented three-state null handling per ADR-0024: null (publisher asserts "doesn't apply"), undefined (absent/unknown), and a present value are preserved end-to-end through toCommon/fromCommon
  - Added new types: Handler, SchemasInput, ObjectSchemasInput, ObjectSchemas, PluginMeta, PluginCapability, ObjectMappings, PluginExtensionsObjectConfig, PluginExtensions
  - Added runnable example: pnpm --filter @common-grants/sdk example:transforms

  Python SDK (lib/python-sdk)
  - Added transforms.py and types.py to extensions/ module
  - Updated generate.py, plugin.py, specs.py, transformation.py, and custom_fields.py for parity with TypeScript changes
  - Added tests/extensions/ with test coverage for plugin, transforms, and types
  - Added runnable example (examples/transforms.py) and grants.gov plugin example (examples/plugins/grants_gov/)

  ADRs (website/)
  - Updated ADR-0022 (plugin framework) to reflect the flattened schemas-centric contract

  ---
  Context for reviewers

[...]

PR #820, opened 2026-05-12
[Issue #650] Updating readme

jcrichlake:
Summary

  - Fixes #650
  - Time to review: 10 minutes

  Changes proposed

  Updated and expanded the Python SDK documentation:

  - lib/python-sdk/README.md: Rewrote with a table of contents, reorganized into Usage and Modules sections, updated Quick Start and Kitchen Sink examples to reflect the current API
  (client.opportunities instead of client.opportunity), and restructured the Modules section with a summary table linking to sub-module READMEs.
  - lib/python-sdk/common_grants_sdk/client/README.md: Rewrote with a table of contents, consolidated authentication and configuration docs, added an environment variable reference
  table, documented the low-level HTTP methods (client.get() / client.post()), and added an API reference section with method tables for Client, Opportunities, and Auth.
  - lib/python-sdk/common_grants_sdk/schemas/README.md: New file documenting the schemas module, covering validation patterns, type safety with OppStatusOptions, generic response
  schemas (Paginated[T], Sorted[T], Filtered[T, F]), data transformation utilities, and a full API reference.

  Context for reviewers

  These are documentation-only changes — no source code was modified. Verification was done by reading the current source to ensure all examples and API references reflect the actual
  SDK behavior (e.g., client.opportunities.list(), client.opportunities.search(), Auth.api_key(), Auth.bearer()).

  Additional information

  No screenshots needed — all changes are Markdown documentation files.

PR #838, opened 2026-05-15
757 extend define plugin

### LINKED ISSUE #757: [Py SDK] Extend define_plugin() to accept schemas with to_common/from_common
### Summary

Extend `define_plugin()` in `common-grants-sdk` to accept per-object schema transforms under the `schemas` slot defined in ADR 0022. Each entry under `schemas[ObjectName]` takes `ObjectSchemasInput(native=..., to_common=..., from_common=...)`, where `to_common`/`from_common` are plain hand-written callables. The declarative `build_transforms()` path lives in #745. Other Plugin slots (`get_client`, `filters`, `meta`, `extensions`) are out of scope here and covered by sibling tickets.

### Acceptance criteria

- [ ] `common_grants_sdk` exports types `ObjectSchemas[TNative, TCommon]` and `ObjectSchemasInput[TNative, TCommon]`
- [ ] `define_plugin()` accepts per-object `ObjectSchemasInput` under `schemas`, compiles it into `ObjectSchemas[T]` at the same key, and merges any `extensions.schemas[Object].custom_fields` into the resulting `common` model
- [ ] `native` defaults to `dict[str, Any]` when omitted
- [ ] Existing custom-fields-only plugins (shipped in #508) continue to work with no changes — all new options are additive
- [ ] Unit tests cover: (a) custom-fields-only plugin still works; (b) plugin with hand-written `to_common`/`from_common`; (c) typed inputs preserve their `TNative` / `TCommon` parameters through `ObjectSchemasInput[T1, T2]` → `ObjectSchemas[T1, T2]` at the output key, and pass Pyright strict

### Related

- Parent epic: #656
- Origin spike: #657
- ADR: [ADR 0022 — Plugin framework](https://commongrants.org/governance/adr/0022-plugin-framework) (canonical URL once merged; currently drafted on [branch 730-ADR-Draft](https://github.com/HHS/simpler-grants-protocol/blob/730-ADR-Draft/website/src/content/docs/governance/adr/0022-plugin-framework.md))
- Prior Py SDK work: #508 (`define_plugin()` + build-time generator baseline)
- Siblings: #744 (validation), #745 (`build_transforms()` + handler registry), #746 (migrate PA/CA examples), #645 (custom filters + `get_client`)
- TypeScript counterpart: #756

### Supersedes

Part of #743, split per [Billy's feedback](https://github.com/HHS/simpler-grants-protocol/issues/743#issuecomment-4292271439) following the #508 / #509 precedent.

jcrichlake:
### Summary

  - Fixes #757
  - Time to review: 30 minutes

  ### Changes proposed

  **Core framework (`common_grants_sdk/extensions/`)**
  - Added `types.py` — new plugin framework types per ADR-0022: `PluginExtensions`, `PluginExtensionsMeta`, `PluginExtensionsSchema`, `ObjectSchemasInput`,
  `ObjectSchemas`, `TransformResult`, `PluginError`, `Handler`, `ObjectMappings`, `PluginCapability`
  - Added `transforms.py` — `build_transforms()` generates `to_common`/`from_common` callables from declarative mapping dicts, with structural validation at call
  time, custom handler registration, and optional Pydantic model validation via `common_model`
  - Updated `specs.py` — `merge_extensions()` now accepts `PluginExtensions` (structured objects) instead of the flat `SchemaExtensions` TypedDict; added
  `_merge_meta()` for conflict-aware metadata merging
  - Updated `plugin.py` — `define_plugin()` now accepts `meta`, `extensions`, `schemas`, `get_client`, and `filters` (all optional); `Plugin` and `PluginConfig`
  updated to match
  - Updated `__init__.py` — exports `build_transforms` and all new ADR-0022 types
  - Updated `generate.py` — code generator now handles the new `PluginConfig` shape, emits `ObjectSchemas` instances in generated `__init__.py`, and injects
  transform callables at generation time

  **Transformation utilities (`utils/transformation.py`)**
  - Added `numberToString` and `stringToNumber` built-in handlers
  - Added `HandlerError` for attributing transform failures to specific handlers
  - Added `DEFAULT_HANDLERS` registry dict and `handlers` parameter to `transform_from_mapping`

  **Examples**
  - Added `examples/plugins/grants_gov/cg_config.py` — grants.gov sample plugin demonstrating all 6 mapping types (`field`, `const`, `match`, `numberToString`,
  `stringToNumber`, and custom handlers)
  - Added `examples/transforms.py` — end-to-end demo of bidirectional transforms with roundtrip verification, custom handlers (`join`/`split`), and typed model
  validation
  - Updated `examples/plugins/opportunity_extensions/cg_config.py` — migrated from `SchemaExtensions` dict format to `PluginExtensions` objects; added module
  docstring
  - Updated `examples/plugin_custom_fields.py` — updated call sites to use `schemas.Opportunity.common.model_validate()` and
  `extensions.schemas["Opportunity"].custom_fields`

  **Tests and docs**
  - Added `tests/extensions/test_plugin.py`, `test_transforms.py`, `test_types.py`
  - Updated `test_merge_extensions.py`, `test_plugin_generator.py`, `test_transformation.py`, `test_custom_fields.py`, `test_plugin_registry.py`
  - Updated `extensions/README.md`, ADR-0017, and ADR-0022

  ### Context for reviewers

  This PR extends `define_plugin()` to support the full plugin framework shape described in ADR-0022. The key design decisions:

  - `build_transforms()` validates mapping structure at call time (not generation time) and surfaces failures as `PluginError` entries in `TransformResult.errors`
  rather than raising — consumers choose strict vs. lenient handling
  - Both transform directions are author-provided; `build_transforms()` intentionally does not invert one into the other because many-to-one handlers like
  `match`/`switch` are not reversible
  - Custom handlers are registered per `build_transforms()` call only and must not shadow `DEFAULT_HANDLERS` (raises `ValueError` at call time)
  - The code generator injects `to_common`/`from_common` callables into generated `__init__.py` so `plugin.schemas.Opportunity.to_common` is populated without plugin
   authors needing to do it manually

  **Testing instructions:**

  From `lib/python-sdk/`:

  ```bash
  # 1. Install dependencies
  poetry install

  # 2. Run the full check suite (formatting, linting, type checking)
  make checks

  # 3. Run all tests
  poetry run pytest

  # 4. Regenerate plugin schemas from cg_config.py files
  make plugins

  # 5. Run the bidirectional transforms example (no server required)
  poetry run python examples/transforms.py

  # 6. Run the plugin custom fields example (no server required)
  poetry run python examples/plugin_custom_fields.py

  Steps 5 and 6 should produce no errors. The transforms example should end with Roundtrip result (5 mapped fields checked; unmapped fields dropped by design): ALL
  PASS.

  Additional information

  transforms.py output:
  ============================================================
  to_common: grants.gov → CommonGrants
  ============================================================
  Errors: none

  Result:
  {
    "title": "Research into conservation techniques",
    "status": { "value": "open", "description": "The opportunity is currently accepting applications" },
    "funding": {
      "minAwardAmount": { "amount": 10000, "currency": "USD" },
      "maxAwardAmount": { "amount": 100000, "currency": "USD" }
    },
    "keyDates": { ... },
    "customFields": {
      "legacyIdStr": { "value": "12345" },   ← numberToString: int 12345 → str
      "priorityScore": { "value": 75 }       ← stringToNumber: str "75" → int
    }
  }

  ============================================================
  ROUNDTRIP CHECK
  ============================================================
    [PASS] title: 'Research into conservation techniques' -> 'Research into conservation techniques'
    [PASS] status: 'posted' -> 'posted'
    [PASS] award_floor: 10000 -> 10000
    [PASS] award_ceiling: 100000 -> 100000
    [PASS] priority_score_str: '75' -> '75'   ← numberToString in from_common

  Roundtrip result (5 mapped fields checked; unmapped fields dropped by design): ALL PASS


lib/python-sdk/common_grants_sdk/extensions/plugin.py
```diff
@@ -5,77 +5,77 @@
...
+    All fields are optional so adopters can start with only what they need.
     """
 
-    extensions: SchemaExtensions
+    extensions: PluginExtensions | None = None
     meta: PluginExtensionsMeta | None = None
-    transform_schemas: dict[str, ObjectSchemasInput[Any, Any]] | None = None
+    schemas: dict[str, ObjectSchemasInput[Any, Any]] | None = None
+    get_client: Callable[[ClientConfig], Any] | None = None
+    filters: dict[str, Any] | None = None
```
SnowboardTechie: This is the scope thing from the body — `get_client` and `filters` are wired into `Plugin` (lines 57–58), the `define_plugin` kwargs (lines 63, 66), and `generate.py:494-500` (which emits the real `lru_cache` wrap). So whatever signatures they have today is what we ship.

If we wait and add them later as optional kwargs once their interfaces are settled, existing adopters won't notice. Could also split the difference — keep `get_client` (which has a more concrete signature) and drop just `filters`.

SnowboardTechie (commented):
Looked through this — really nice work on the layering between the new types and the generator. A few thoughts:

The "bring your own callable" path works cleanly. Anything matching `Callable[..., TransformResult]` drops into `ObjectSchemasInput` and `build_transforms()` is just one way to produce one. Might be worth a README mention that hand-written ones still need to return `TransformResult` — the type catches it but it's easy to miss when porting existing code.

The bigger thing I'd push on is `get_client` and `filters` (inline below) — they're not really stubs, they're plumbed through enough layers that we're committing to their shape today, and waiting to ship them as optional kwargs once the interfaces firm up would be a no-op for adopters.

One thing that didn't fit inline: `lib/python-sdk/.coverage` (53KB pytest-cov artifact) got committed. Probably just needs a gitignore entry and a `git rm`.

---

**While I was reading the transform surface this builds on** (these are in `transformation.py` and `extensions/transforms.py`, which already landed in `HOLD-transforms` via #810 — so not this PR's job to fix, but flagging in case it's useful before that branch merges):

- `transform_from_mapping` does `if hasattr(data, "model_dump"): data = data.model_dump(mode="json")` (around line 208). That call doesn't set `by_alias=True`, but `CommonGrantsBaseModel` doesn't set `by_alias` by default either, so dumping a Pydantic CG instance gives you snake_case keys (`min_award_amount`) while the from_common mappings use the camelCase aliases (`funding.minAwardAmount.amount`, see `examples/plugins/grants_gov/cg_config.py:94`). The path lookups silently miss and you get an empty result with no errors raised. The example happens to work because the chain is `to_common(dict) → dict → from_common(dict)` — both sides are dicts so it never matters. But `from_common(opportunity_model)` directly will break. Fix is `data.model_dump(mode="json", by_alias=True)` plus an `isinstance(data, BaseModel)` check instead of `hasattr` (the duck-type would match unrelated classes).
- `build_transforms`'s `common_model` parameter only applies to `to_common` output — `from_common` ignores it. Symmetric naming threw me a bit; might read better as `to_common_model=`.
- The `to_common` validation block only catches `ValidationError`. Anything else from `model_validate` (e.g. `TypeError` from a misconfigured root validator) bubbles up and breaks the errors-as-values contract the rest of the function maintains.
- `DEFAULT_HANDLERS` has `"switch"` aliased to `"match"` for backward compat, but the tests in `test_transforms.py` still use `switch` while the new `grants_gov/cg_config.py` example uses `match`. Two names for the same thing across canonical example vs. canonical tests is confusing for someone reading the code to learn the shape.

lib/python-sdk/common_grants_sdk/extensions/generate.py
```diff
@@ -393,34 +440,108 @@ def _render_generated_init_py() -> str:
...
+        ]
+
+    # Explicit schemas: assert then inject callables (and native type if provided) per object.
+    if explicit_objs:
+        inject_lines.append("assert config.schemas is not None")
+        inject_lines.append("")
+        for obj in sorted(explicit_objs):
+            inject_lines += [
+                f'assert config.schemas["{obj}"].to_common is not None',
+                f'assert config.schemas["{obj}"].from_common is not None',
```
SnowboardTechie: Heads up — `assert` gets stripped under `python -O` (and `PYTHONOPTIMIZE=1`). Anyone running with optimizations on loses these guards entirely and gets a confusing `TypeError` somewhere downstream instead of the clear error here.

Probably want something like:

```python
f'if config.schemas["{obj}"].to_common is None:',
f'    raise ValueError(f"Plugin object {obj!r}: to_common callable is required")',
```

Same story for the `from_common` line below and the `config.schemas` check on line 483.

lib/python-sdk/common_grants_sdk/extensions/specs.py
```diff
@@ -51,43 +54,107 @@ def _merge_fields(
...
+        if incoming_val is None:
+            merged[field_name] = current_val
+        elif current_val is None:
+            merged[field_name] = incoming_val
+        else:
+            # Both have non-None values — apply conflict strategy
+            if on_conflict == "error":
+                raise ValueError(
+                    f'merge_extensions: duplicate meta field "{field_name}"'
+                )
```
SnowboardTechie, line 65: Small one — this names the field but not the values, so a multi-plugin merge gives you a few identical `merge_extensions: duplicate meta field "name"` errors with nothing to grep on. Including the values would help:

```python
raise ValueError(
    f'merge_extensions: duplicate meta field "{field_name}" '
    f'(existing: {current_val!r}, incoming: {incoming_val!r})'
)
```

lib/python-sdk/common_grants_sdk/extensions/specs.py
```diff
@@ -51,43 +54,107 @@ def _merge_fields(
...
+            merged[field_name] = (
+                current_val if on_conflict == "first_wins" else incoming_val
+            )
+
+    return _PluginExtensionsMeta(
+        name=merged["name"],
+        version=merged["version"],
+        sourceSystem=merged["source_system"],
+        capabilities=merged["capabilities"],
```
SnowboardTechie, line 74: Style nit: `sourceSystem=` sticks out next to the other snake_case kwargs (`name=`, `version=`, `capabilities=`) as the lone alias. `populate_by_name=True` so both work — reads cleaner as `source_system=merged["source_system"]`.

SnowboardTechie:
HOLD-transforms synced from `main` on 2026-05-20: 12 commits — 7 Dependabot / catalog dep bumps (#829, #826, #822, #821, #819, #817, #808), 3 CI / docs chores (#834, #820, #816), 2 release version bumps [skip ci]. Merge was clean (no conflicts). You may want to rebase or merge HOLD-transforms back into your branch.

lib/python-sdk/common_grants_sdk/extensions/generate.py
```diff
@@ -334,13 +361,34 @@ def _render_schemas_py(extensions: SchemaExtensions) -> str:
...
+    # Build schema assignments: each attribute is an ObjectSchemas instance so
+    # callers get a unified interface (plugin.schemas.Opportunity.common for the
+    # model class, .to_common/.from_common for transforms).
+    # to_common/from_common default to None here; root __init__.py injects the
+    # real callables for any object that has transforms configured.
+    assignments: list[str] = [
+        f"        self.{name} = ObjectSchemas(native=dict, common={name}, to_common=None, from_common=None)"
+        for name in model_names
+    ]
+    for obj in sorted(mappings_only):
```
SnowboardTechie, line 377: Heads up: this loop covers the mappings-only case, but there's a third bucket that's currently invisible — objects declared in `config.schemas` (explicit transforms) with no `custom_fields` and no `mappings`. They never get added to `_Schemas`, so `_render_plugin_init_py`'s injected `schemas.Opportunity.native = ...` blows up at import with `AttributeError`.

Repro:
```python
config = define_plugin(
    schemas={"Opportunity": ObjectSchemasInput(to_common=_to, from_common=_from)},
)
# generate runs fine; import → AttributeError
```

Nothing today hits this since the canonical example declares both slots, but it'll bite the first "just transforms, no custom fields" plugin. Probably folds into this loop by also iterating `set(config.schemas.keys()) - explicit_cf_objs - mappings_only_objs` with `common` defaulting to the base SDK class, same shape as `mappings_only_objs`. Fine to defer.

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -51,12 +51,18 @@ We decided to:
 The resulting Plugin shape:
 
 ```
-plugin.meta        // name, version, sourceSystem, capabilities
-plugin.getClient   // (config: ClientConfig) => Client; memoized by definePlugin()
-plugin.extensions  // serializable; used by mergeExtensions()
-plugin.schemas.<ObjectName> = { native, common, toCommon, fromCommon }
+plugin.meta                          // name, version, sourceSystem, capabilities
+plugin.get_client                    // (config: ClientConfig) => Client; memoized by the code generator
```
SnowboardTechie, line 55: The rest of the doc still references `getClient` / `filters` throughout, but the merged code has dropped them for this slot. Anyone reading the ADR alongside the code will wonder why the Plugin types don't match. Probably wants a one-line "Implementation status" note near the Decision summary explaining that `get_client` / `filters` are deferred to sibling tickets.

SnowboardTechie (approved):
Nice work, this is feeling solid. The `by_alias` + `isinstance(BaseModel)` fix, catch-all `Exception` on both transform directions (with the new test), and dropping `get_client` / `filters` all read clean. `match` as canonical with `switch` kept as the tested alias is a nice landing.

Two small follow-ups dropped inline — neither needs to land here. Approving.

lib/python-sdk/examples/plugins/grants_gov/cg_config.py
```diff
@@ -105,39 +117,50 @@
...
+                    "priorityScore": CustomFieldSpec(
+                        field_type=CustomFieldType.NUMBER,
+                        name="Priority score",
+                        description="Numeric priority score coerced from a string via stringToNumber",
+                    ),
+                },
+            )
+        }
+    ),
+    schemas={
```
widal001: @jcrichlake @SnowboardTechie I know I've been a little out of the loop here, but I think something might have gotten lost in translation between our initial discussions around ADR-0022 and this implementation:

Based on this code it looks like we have the following attributes in `DefinePluginOptions`:

- `schemas` -- Keyed by object, holds the transformation code logic
- `extensions.schemas` -- ALSO keyed by object, but only contains the custom fields

I thought the whole point of consolidating transformation logic and custom fields under a single dictionary keyed by object was so that users only had to add one new entry when introducing a new object (e.g. `Applications`). But this still requires them to add two new entries.

I'd suggest we either consolidate these into a single root-level `schemas.<object>.{customFields, native_schema, to_common, from_common}` or we keep them separate and rename them:

- `transformations` with this shape `<object>.{native_schema, to_common, from_common}`
- `customFields` with this shape `<object>.CustomFieldSpec`

widal001 (changes requested):
@jcrichlake See my comment above.

I'll continue working through this PR when I have time, but I think we'd need to combine or clarify the naming around `schemas` and `extensions.schemas` which will probably trickle down into how the rest of this PR is implemented.

lib/python-sdk/examples/plugins/grants_gov/cg_config.py
widal001: @SnowboardTechie does the TypeScript SDK transformation changes also have this same issue?

lib/python-sdk/examples/plugins/grants_gov/__init__.py
```diff
@@ -6,9 +6,21 @@
...
+if config.schemas is None:
+    raise ValueError("config.schemas is required when explicit transforms are declared")
+
+if config.schemas["Opportunity"].to_common is None:
+    raise ValueError("Plugin object 'Opportunity': to_common callable is required")
+if config.schemas["Opportunity"].from_common is None:
+    raise ValueError("Plugin object 'Opportunity': from_common callable is required")
+schemas.Opportunity.native = config.schemas["Opportunity"].native or dict
+schemas.Opportunity.to_common = config.schemas["Opportunity"].to_common
+schemas.Opportunity.from_common = config.schemas["Opportunity"].from_common
```
widal001: Assuming this is auto-generated, this will get messy very quickly as we add new objects that support transformations.

We should explore doing this with a function or generics, to see if there's a way to preserve type hinting without explicitly re-declaring all of these.

If this is the best way to copy the native schemas and transformation methods over from the plugin config, then it should at at minimum be moved out of `__init__.py` so this file stays lean.

lib/python-sdk/common_grants_sdk/extensions/transforms.py
widal001, line 1: This file looks like it needs to be update to change the `Any` to a set of generics that actually leverage `native` and `common` schemas so that the input and output of `to_common()` and `from_common()` have proper type hinting.

lib/python-sdk/common_grants_sdk/extensions/plugin.py
widal001, line 1: The `dict[str, ObjectSchemasInput[Any, Any]]` in this file drops the per-Object generics. Even if I construct `ObjectSchemasInput[GrantsGovOpportunity, Opportunity]` as a caller, the typing widens to `Any` once it's stored in this dict.

Until we fix this, typing improvements elsewhere (like making `build_transforms` generic) won't actually be preserved in the `Plugin` instance because the types get widened back to `Any` here.

A couple of options worth exploring: a `TypedDict` keyed by object name with object-specific `ObjectSchemasInput` types, or maybe using something like `@overload`s on `define_plugin`

lib/python-sdk/common_grants_sdk/extensions/transforms.py
jcrichlake: I've attempted to address in subsequent commits.

lib/python-sdk/examples/plugins/grants_gov/cg_config.py
jcrichlake: Thanks for noting this! I went ahead and moved everything under the schemas object.  I think this was a hold over from the PoC which was a partial implementation.

SnowboardTechie: Good news on the TS side. #825 is still in draft, and current shape has `customFields` on `ObjectSchemasInput`, so the per-object entry under `transformSchemas` carries `native + customFields + toCommon + fromCommon` together. No second key per object to maintain.

One thing the consolidation surfaces: pulling `customFields` out of the JSON-safe `extensions` surface means `mergeExtensions` over customFields drops out of the consolidated path, and the ADR-0022 text still describes the old dual-keyed shape. Better tracked as ADR cleanup than a separate amendment.

widal001 (commented):
@jcrichlake if you've already addressed my comments, don't feel like you have to wait on a second review from me to merge!

I honestly probably won't be able to dig into this further but if the generated code has type hints and the `schemas` portion of `definePlugin()` and `Plugin` interfaces have been updated, then I trust you to merge it in.

Maybe we can just schedule a longer design/code review session for the Python/TypeScript SDK's once we're getting ready to ship a release?

jcrichlake:
> @jcrichlake if you've already addressed my comments, don't feel like you have to wait on a second review from me to merge!
> 
> I honestly probably won't be able to dig into this further but if the generated code has type hints and the `schemas` portion of `definePlugin()` and `Plugin` interfaces have been updated, then I trust you to merge it in.
> 
> Maybe we can just schedule a longer design/code review session for the Python/TypeScript SDK's once we're getting ready to ship a release?

I think this is a great idea @widal001 I'll go ahead and merge this in.

Issue #735, opened 2026-04-16
[CLI] Fix how `resolveAnyOf()` in `cg check spec` works


widal001:
### Summary

`resolveAnyOf()` in `lib/cli/src/commands/check/utils/flatten-schemas.ts` (lines 57-78) self-describes as *"a simplified approach that avoids complex union type resolution"* — it selects the first non-`$ref` option from an `anyOf` and discards the rest. It also doesn't process `oneOf` at all. This causes ~172 false positives when checking a TypeScript/Hono implementation against the base protocol, while accidentally letting equivalent Pydantic/FastAPI specs pass.

Three edge cases:

**1. Nullable types: `type` array vs `anyOf` (~132 false positives)**

Both representations are equivalent in OpenAPI 3.1, but the CLI only handles the `anyOf` form — and handles it incorrectly by stripping `{type: "null"}`.

Zod-openapi generates (fails):
```json
{
  "source": { "type": ["string", "null"], "format": "uri" },
  "funding": { "type": ["object", "null"], "properties": { "..." : "..." } }
}
```

Pydantic generates (passes — because `resolveAnyOf()` strips `{type: "null"}`, accidentally collapsing it to non-nullable which matches the base spec):
```json
{
  "source": { "anyOf": [{ "type": "string", "format": "uri" }, { "type": "null" }] },
  "funding": { "anyOf": [{ "$ref": "#/components/schemas/OppFunding" }, { "type": "null" }] }
}
```

**2. Discriminated unions: `oneOf` vs `anyOf` (~30 false positives)**

The base protocol (TypeSpec) generates `anyOf` for the `Event` union. Zod's `discriminatedUnion` renders as `oneOf`. The CLI processes `anyOf` (stripping it to one variant) but ignores `oneOf` entirely, causing a structural mismatch.

Zod-openapi generates (fails — `oneOf` not processed):
```json
{
  "postDate": {
    "oneOf": [
      { "type": "object", "required": ["name", "eventType", "date"], "properties": { "..." : "..." } },
      { "type": "object", "required": ["name", "eventType", "startDate", "endDate"], "properties": { "..." : "..." } },
      { "type": "object", "required": ["name", "eventType"], "properties": { "..." : "..." } },
      { "type": "null" }
    ]
  }
}
```

Pydantic generates (passes — `resolveAnyOf()` keeps only the first `$ref`, discarding other variants):
```json
{
  "postDate": {
    "anyOf": [
[...]

PR #855, opened 2026-05-26
[Issue #736] ADR: Nullable Optional Fields

### LINKED ISSUE #736: ADR: Should optional fields in the CommonGrants protocol be nullable?
### Summary

Both the Python SDK (Pydantic `Optional[X]`) and the TypeScript SDK (Zod `.nullish()`) render optional fields as nullable. This is intentional for cross-language compatibility since Python's `Optional[X]` requires nullability. However, the base protocol spec (TypeSpec-generated) declares these fields as optional but non-nullable. Once the `resolveAnyOf()` bug in `cg check spec` is fixed (see #735), both SDK implementations will correctly fail because they add nullable where the protocol doesn't allow it. A decision is needed on whether the base protocol should align with the SDKs.

### Options

1. **Make optional fields nullable in the base protocol.** Update the TypeSpec source to use nullable optionals, aligning with both SDKs. Consumers can send `null` or omit the field. This matches how Python and JavaScript naturally model "may not be present."
2. **Keep optional fields non-nullable in the base protocol.** Both SDKs stop using `.nullish()` / `Optional[X]` and rely on field absence. Python SDK would need a custom Pydantic config to omit `None` fields from serialization rather than sending `null`. More restrictive but cleaner wire format.
3. **Allow both representations.** The protocol accepts either absent or `null` as equivalent, and `cg check spec` treats `type: T` as compatible with `type: [T, "null"]`. Pragmatic but loosens the spec.

### Decision criteria

- Cross-language compatibility (Python, TypeScript, future SDKs)
- Wire-format clarity (should `null` and absent be semantically different?)
- Backwards compatibility with existing implementations
- Alignment with OpenAPI 3.1 / JSON Schema conventions

### Approvers

- CommonGrants protocol maintainers

### Acceptance criteria

- [ ] The approvers for this decision have been identified (ideally before work on the ADR starts)
- [ ] The ADR has been added to the list of ADRs in `website/src/content/docs/decisions/adr/` with a number prefix like `0002-decision-summary.md`
- [ ] The ADR has been reviewed and approved by the approvers listed above
- [ ] The ADR satisfies requirements that are outlined in the ADR template

jcrichlake:
### Summary

  - Fixes #736
  - Time to review: 5 minutes

  ### Changes proposed

  Adds ADR-0024 documenting the decision on optional field nullability in the CommonGrants protocol spec and SDKs.

  ### Context for reviewers

  This ADR resolves the question of whether optional fields in the base protocol (TypeSpec-generated) should be nullable. The current situation: both the Python SDK
  (`Optional[X]`) and TypeScript SDK (`.nullish()`) render optional fields as nullable, but the base protocol declares them as optional-but-non-nullable. Once the
  `resolveAnyOf()` bug in `cg check spec` is fixed (#735), both SDKs will fail validation.

  **Decision: Make optional fields nullable in the base protocol (Option 1).**

  The key insight driving this decision is that publishers need to express three distinct field states — "not provided" (field absent), "doesn't apply" (explicit
  `null`), and "has a value" — and this three-state model is only achievable with a nullable spec, especially for scalar types like dates where no natural sentinel
  value exists.

  No code changes are included in this PR; the ADR documents the rationale so the corresponding TypeSpec and implementation changes can follow in a subsequent PR.

  ### Additional information

  The three options evaluated in the ADR:

  | Option | Description | Chosen? |
  |--------|-------------|---------|
  | Option 1 | Make optional fields nullable in the base protocol | ✅ |
  | Option 2 | Keep non-nullable; update SDKs to remove `.nullish()` / `Optional[X]` | ❌ |
  | Option 3 | Allow both representations; treat `null` and absent as equivalent | ❌ |

### DIFF
--- a/website/.cspell.json
+++ b/website/.cspell.json
@@ -49,7 +49,8 @@
     "CFDA",
     "HBCUs",
     "TCCUs",
-    "widal001"
+    "widal001",
+    "categoricals"
   ],
   "ignoreWords": [
     "opblock",
--- a/website/src/content/docs/governance/adr/0024-optional-field-nullability.md
+++ b/website/src/content/docs/governance/adr/0024-optional-field-nullability.md
@@ -1,0 +1,123 @@
+---
+title: Optional field nullability in the protocol spec and SDKs
+description: ADR documenting the decision on whether optional fields in the CommonGrants protocol should be nullable at the spec level.
+---
+
+Both the Python SDK (`Optional[X]` via Pydantic) and the TypeScript SDK (`.nullish()` via Zod) currently render optional fields as nullable. This is intentional for cross-language compatibility, since Python's `Optional[X]` requires nullability. However, the base protocol spec (TypeSpec-generated) declares these fields as optional but non-nullable. Once the `resolveAnyOf()` bug in `cg check spec` is fixed (see [#735](https://github.com/HHS/simpler-grants-protocol/issues/735)), both SDK implementations will correctly fail validation because they add `nullable` where the protocol doesn't allow it. A decision is needed on whether the base protocol should align with the SDKs, or vice versa.
+
+Experience from production grant data publishing systems highlights a subtlety that shapes this decision: `optional + nullable` is not sufficient on its own. Publishers need to express three distinct field states, not two:
+
+| State | Meaning |
+| ----- | ------- |
+| Field absent | "Not provided" — the publisher did not supply this data |
+| Explicit N/A value | "Doesn't apply" — the field is intentionally irrelevant for this record |
+| Field present with value | "Has a value" |
+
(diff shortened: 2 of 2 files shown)


SnowboardTechie (approved):
👍🏻

SnowboardTechie:
👍 — approved.

Heads up that this ADR has handler-level implications the SDK PoCs didn't anticipate. The TypeScript transforms PoC (#825) had three coercing handlers (`numberToString`, `stringToNumber`, `switchOnValue`) that collapsed an explicit `null` source into `undefined`, which would have made the publisher's "doesn't apply" assertion unobservable through `toCommon` / `fromCommon` once ADR-0024 lands.

I've pushed alignment commits to #825 that:

- preserve `null` through the coercing handlers (return type widens accordingly)
- add an opt-in `case: { "null": "..." }` to `match` / `switch` for target-side translation, while keeping pass-through as the default
- document the three-state contract in the SDK README for custom-handler authors
- exercise the round-trip end-to-end via the existing example (`source_url: null` survives both directions)

ADR-0024's "SDKs already reflect this model and require no changes" framing is accurate for the validation surface (`.nullish()` already accepts both), but the transforms layer is downstream of that and needs its own pin. Worth a follow-up against the Python PoC (#810) for the same alignment — TS leads, Python parity tracked there.

No changes needed in ADR-0024 itself.

PR #864, opened 2026-05-27
Fixing pipeline

jcrichlake:
Small PR for addressing a broken plugin that is slightly behind in its updates.  We should be able to pull this out after the poetry-audit-plugin ships a fixed release .


widal001 (approved):
Thanks for catching and fixing this

PR #866, opened 2026-05-28
Jeff/update pipeline

jcrichlake:
Another PR to make a small change to the pipeline for the broken plugin.  This is purely to get the Python builds working again.


widal001 (approved):
Mixing pip installs and poetry can cause some unintended consequences, so might be worth keeping an eye on this, but approving to unblock!

Issue #645, opened 2026-03-31
[SDK] Custom filters


SnowboardTechie:
## Summary

Define and implement custom filter support in the Python and TypeScript SDKs, including defining custom filters, using them in API client methods, and exposing them through an implementation's API.

## Context

[ADR 0012](https://commongrants.org/governance/adr/0012-filtering/#custom-filters) establishes that API endpoints supporting filtering use POST operations with a `filters` parameter in the request body. Each filter conforms to a standard `Filter` schema with an `operation` (eq, neq, gt, gte, lt, lte, like, in, not_in, between) and a `value`.

Routes that support implementation-defined filters must include a `customFilters` property in the request body, whose values conform to the same `Filter` schema:

```json
{
  "filters": {
    "title": { "value": "example", "operation": "like" },
    "customFilters": {
      "agency": { "value": ["Department of Transportation"], "operation": "in" }
    }
  }
}
```

The SDKs need to give developers a way to define these custom filters, pass them into client methods, and expose them from their own API implementations — with type safety and validation throughout.

## Acceptance Criteria

- [ ] SDK users can define custom filters
- [ ] SDK users can use registered custom filters in API client methods (e.g., `search()`, `list()`)
- [ ] Custom filters are serialized into the `customFilters` property of the request body per ADR 0012
- [ ] SDK users can expose custom filters through an implementation's API
- [ ] Filter operations are type-safe and validated against the supported operation set

PR #884, opened 2026-06-09
Custom filters TS PoC (#868)

### LINKED ISSUE #868: [TS SDK] Draft custom filter PoC: TypeScript
### Summary

Build a TypeScript proof-of-concept for the custom filter interface. Validates the design shape recommended in SPIKE #646 before locking the full implementation in both SDKs.

Issued separately from the Python PoC to avoid SDK dependencies between them.

### Acceptance criteria

- [x] PoC implements the custom filter design recommended in #646
- [x] PoC demonstrates a custom filter applied to a sample CommonGrants search request (customFilters property, standard Filter schema per ADR-0012)
- [x] Findings and design questions captured to inform full SDK implementation

SnowboardTechie:
### Summary

- Fixes #868 — the TypeScript custom-filter PoC: implements the design recommended in SPIKE #646, demonstrates a custom filter on a sample CommonGrants search request (the `customFilters` property + standard Filter schema per ADR-0012), and captures findings to inform the full SDK implementation.
- Merges into the `HOLD-filters` bucket (cut from `HOLD-transforms`, where the top-level `schemas` plugin shape this `routes` key extends already lives).
- Time to review: ~20 minutes

> **Status:** Proof-of-concept. Not the production surface — it exists to retire design risk and produce findings that feed forthcoming ADR-0022 / ADR-0012 amendments.

### Changes proposed

**Declaration layer**
- `routes` key on `definePlugin()`, passed through unchanged to `Plugin.routes` (route-keyed custom-filter declarations).
- 11-value `CustomFilterType`, plus `CustomFilterSpec` / `RouteDeclarations` / `PluginRoutes` types.
- `BooleanComparisonFilterSchema` added to the SDK Zod schemas.

**Runtime**
- `classifyFilters()` — three-bucket classifier turning a flat consumer `filters` object into the ADR-0012 `OppFilters` wire body: default fields → top-level named fields, registered-custom + ad-hoc → `customFilters`, `gov.<system>@<filter>` keys → verbatim passthrough.
- `validateRoutes()` / `validateFilterCall()` raising structured `PluginError`.
- `F.*` helper namespace compiling to `{ operator, value }`.

**Type-safety proof (the central risk this PoC retires)**
- Compile-time `@ts-expect-error` tests proving unknown keys, operator/filterType mismatches, and value-shape mismatches fail to compile **with** `as const` — and silently widen **without** it. The build enforces these (an unused directive fails `tsc`).

**Docs / example**
- Runnable canonical grants.gov example (`examples/custom-filters.ts`), wired into the `ci` chain.
- New "Plugin custom filters (PoC)" section in the extensions README, with a prominent routes-vs-schemas distinction (custom filters attach to resource methods; custom fields attach to schemas).

No `lib/core` change; no changeset (PoC on a `HOLD-*` bucket, no production-dependency change).

### Context for reviewers

Verified with the real toolchain:
- `pnpm --filter @common-grants/sdk run build` (tsc + typespec) — green. The build is what enforces the compile-time narrowing proof; vitest strips types and does not validate the `@ts-expect-error` directives.
- `pnpm --filter @common-grants/sdk test` — full suite green (25 files).
- Exercised from a downstream consumer against the **tarball-installed** package shape (the real `exports` / `files` publish map, not a local `src` import) — the classifier surface resolves and classifies correctly from the consumer's vantage; no publish-map change was needed.

A findings doc maps the six #646 questions to what the PoC proved and records ADR-correction candidates; it lives as a planning artifact alongside this work (not in the diff).

### Additional information

Consumer-side classification (example output):
- `status`, `closeDateRange` → top-level named fields (default bucket)
- `agency`, `fundingProgram` → `customFilters` (registered custom filters)
- `legacyTag` → `customFilters` (ad-hoc passthrough)


lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -0,0 +1,357 @@
...
+    schema: NumberArrayFilterSchema,
+    operatorEnum: z.enum(["in", "notIn"]),
+  },
+  numberRange: {
+    schema: NumberRangeFilterSchema,
+    operatorEnum: z.enum(["between", "outside"]),
+  },
+  // integerComparison reuses NumberComparisonFilterSchema — see module JSDoc
+  integerComparison: {
+    schema: NumberComparisonFilterSchema,
```
jcrichlake: Is there a reason this doesn't use the integer schema?

lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -0,0 +1,357 @@
...
+/**
+ * All 11 valid CustomFilterType values — used for unknown-filterType detection.
+ */
+const VALID_FILTER_TYPES = new Set<string>(Object.keys(FILTER_TYPE_SCHEMAS));
+
+/**
+ * The default-filter field names from `OppDefaultFiltersSchema`.
+ * Custom filter names must not collide with these.
+ */
+const DEFAULT_FILTER_NAMES = new Set<string>(Object.keys(OppDefaultFiltersSchema.shape));
```
jcrichlake, line 74: One thing to note here is that eventually we will have different objects each with their own custom filters. We may want to pass the object in somewhere or make a note to handle this when we add award filters to the SDK.

lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -0,0 +1,357 @@
...
+  notIn: (value: unknown[]) => ({ operator: "notIn" as const, value }),
+  /** String pattern match — `{ operator: "like", value }` */
+  like: (value: string) => ({ operator: "like" as const, value }),
+  /** String pattern non-match — `{ operator: "notLike", value }` */
+  notLike: (value: string) => ({ operator: "notLike" as const, value }),
+  /** Range (inclusive) — `{ operator: "between", value: { min, max } }` */
+  between: (min: unknown, max: unknown) => ({ operator: "between" as const, value: { min, max } }),
+  /** Range (exclusive) — `{ operator: "outside", value: { min, max } }` */
+  outside: (min: unknown, max: unknown) => ({ operator: "outside" as const, value: { min, max } }),
+};
```
jcrichlake, line 125: Should the types of most of these be unknown? Or is it better to narrow them where possible?

lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -0,0 +1,357 @@
...
+ * @param methodKey - The method name (e.g. `"search"`)
+ * @param consumerFilters - The flat consumer-facing filters object
+ * @returns The classified `OppFilters` wire body
+ */
+export function classifyFilters(
+  routes: PluginRoutes,
+  resourceKey: string,
+  methodKey: string,
+  consumerFilters: Record<string, unknown>
+): z.infer<typeof OppFiltersSchema> {
```
jcrichlake, line 267: Again this could be other possible base objects beyond OppFilters so I'd think this should be agnostic

lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -0,0 +1,357 @@
...
+  // Resolve registered filter specs for this route-method
+  const registeredFilters: Record<string, CustomFilterSpec> =
+    routes[resourceKey]?.[methodKey]?.filters ?? {};
+
+  const defaultFields: Partial<z.infer<typeof OppDefaultFiltersSchema>> = {};
```
jcrichlake, line 272: Another instance of the opportunity specific filters

lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -0,0 +1,357 @@
...
+    } else {
+      // Bucket 2 (registered custom) or Bucket 3 (ad-hoc / gov.* namespaced)
+      // Run call-time validation — passes spec if registered, undefined if ad-hoc
+      validateFilterCall(spec, key, value);
+      customFilters[key] = value as z.infer<typeof DefaultFilterSchema>;
+    }
+  }
+
+  // Build wire body — omit customFilters key entirely when empty (match nullish shape)
+  const wireBody: z.infer<typeof OppFiltersSchema> = {
```
jcrichlake: And commenting here to hopefully make it easier to find these.

lib/ts-sdk/src/extensions/README.md
```diff
@@ -595,6 +603,165 @@ for (const err of out.errors) {
...
 
+## Plugin custom filters (PoC)
+
+> **Status:** Proof-of-concept (issue [#868](https://github.com/HHS/simpler-grants-protocol/issues/868), design spike [#646](https://github.com/HHS/simpler-grants-protocol/issues/646)). Feeds forthcoming ADR-0022 / ADR-0012 amendments. Contract follows [ADR-0012](https://commongrants.org/governance/adr/0012-filtering/) and [ADR-0022](https://commongrants.org/governance/adr/0022-plugin-framework/).
+
```
jcrichlake: I don't think we need the status in here.  This should ideally be ready to go once it's published up

lib/ts-sdk/src/extensions/README.md
```diff
@@ -595,6 +603,165 @@ for (const err of out.errors) {
...
+Plugins can declare custom filter specs per route-method. At search time, a single flat consumer `filters` object is classified into the ADR-0012 wire body: default fields land as named top-level fields; custom and ad-hoc keys land under `customFilters`.
```
jcrichlake: I wouldn't reference the wire body, I'd talk about how it ends up being passed in the request to the receiver of the call

lib/ts-sdk/src/extensions/README.md
```diff
@@ -595,6 +603,165 @@ for (const err of out.errors) {
...
+
+### Routes vs. schemas — a critical distinction
+
+> [!IMPORTANT]
+> Custom **filters** attach to resource **methods** (routes), not to schemas. Custom **fields** attach to schemas. These two extension points use different keys on `definePlugin()`:
+>
+> - `definePlugin({ routes: { <resource>: { <method>: { filters: ... } } } })` — custom filters (this section)
+> - `definePlugin({ schemas: { <Model>: { customFields: ... } } })` — custom fields (see [Build-time with plugins](#option-2-build-time-with-plugins))
```
jcrichlake: I'd condense these into 1 example showing the combined use

lib/ts-sdk/src/extensions/README.md
```diff
@@ -595,6 +603,165 @@ for (const err of out.errors) {
...
+  // Ad-hoc filter (not declared in the plugin) → customFilters passthrough
+  legacyTag: F.eq("conservation-2024"),
+};
+
+const wireBody = classifyFilters(
+  grantsGovPlugin.routes!, // routes is optional on Plugin; assert non-null when known
+  "opportunities",
+  "search",
+  consumerFilters
+);
```
jcrichlake: I'm slightly confused here, what is the purpose of this function? Does this provide a second entry point beyond the custom filters passed into the `definePlugin` function?

lib/ts-sdk/package.json
```diff
@@ -72,7 +72,7 @@
     "check:format": "prettier --check .",
     "check:types": "tsc --noEmit",
     "checks": "pnpm run check:lint && pnpm run check:format && pnpm run check:types",
-    "ci": "pnpm run checks && pnpm run build && pnpm run test && pnpm run example:transforms",
+    "ci": "pnpm run checks && pnpm run build && pnpm run test && pnpm run example:transforms && pnpm run example:custom-filters",
```
jcrichlake: I wouldn't add the custom-filters example to the CI.  Eventually we are going to want to have update this example to use some endpoint

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+            ? boolean
+            : T extends "dateComparison"
+              ? string
+              : T extends "dateRange"
+                ? { min: string; max: string }
+                : T extends "moneyComparison"
+                  ? { amount: number; currency: string }
+                  : never;
+
+/**
```
jcrichlake: Is this the cleanest way to do this?

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+  NonNullable<typeof grantsGovPlugin.routes>,
+  "opportunities",
+  "search"
+>;
+
+describe("custom-filters compile-time narrowing (as const)", () => {
+  it("type tests compile", () => {
+    // The real assertion is `tsc --noEmit` exit 0. This trivial Vitest assertion
+    // exists solely so Vitest registers the file as a test suite.
+    expect(true).toBe(true);
```
jcrichlake: This test will never fail so why do we have it?

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+  //
+  // `TypedConsumerFilters<GrantsGovSpecs>` has only agency | fundingProgram | awardCount.
+  // An unknown key triggers: "Object literal may only specify known properties" (ts2353).
+  it("unknown filter key is a compile error (asserted with @ts-expect-error)", () => {
+    const bad: TypedConsumerFilters<GrantsGovSpecs> = {
+      // @ts-expect-error -- unknown key 'notARealFilter' is not a registered filter (ts2353)
+      notARealFilter: { operator: "eq", value: "x" },
+    };
+    void bad;
+    expect(true).toBe(true);
```
jcrichlake: Same thing here

jcrichlake:
Left some comments, looks pretty good for a first draft.  One thing I'll flag as a big picture item, should updates to the various Opportunity client endpoint routes be included as part of this PR?

jcrichlake (changes requested):

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: Only because it didn't exist, the SDK had no integer filter schema to reuse. Added `IntegerComparisonFilterSchema` with `.int()` in 21127a9 and pointed `integerComparison` at it, so non-integer values now fail call-time validation.

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: Yeah, this is the right catch for the full implementation. The `routes` structure already keys by resource, so the declaration side handles it; what's opportunity-bound is this default-name lookup and the `OppFilters` return type. Kept the PoC scoped to `opportunities.search` per #868, but the full build needs a resource to default-filters-schema map so awards carry their own defaults. I'll make sure that's explicit in the #645 breakdown when the implementation issues get filed.

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: Same point, answered on the `DEFAULT_FILTER_NAMES` thread above.

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: The wide params are deliberate. `F.gt` and friends are shared across number, date, and money filter types, so the narrowing lives elsewhere: `TypedConsumerFilters` narrows operator and value shape per registered filter at compile time, and the Zod schemas enforce the same at call time. Narrowing the helpers themselves would mean per-filterType variants, which mostly duplicates what the typed layer already does. Could be worth it in the full SDK if we want the extra DX at the helper level.

lib/ts-sdk/src/extensions/README.md
SnowboardTechie: I matched the transforms section above, which carries the same banner. Same logic applies to both, so I'd drop the two together rather than leave one labeled. If that works for you I'll pull both in this PR.

lib/ts-sdk/src/extensions/README.md
SnowboardTechie: Reworded in f0f9757. The section now describes it as the body of the search request sent to the API, and the snippet variable is `requestBody`.

lib/ts-sdk/src/extensions/README.md
SnowboardTechie: Condensed in f0f9757, it's one `definePlugin` example now with `routes` and `schemas` side by side.

lib/ts-sdk/src/extensions/README.md
SnowboardTechie: Not a second registration entry point. `definePlugin` registers the declarations (which filter names exist and what type each is); `classifyFilters` is the search-time step that takes the actual filter values and builds the request body. In the full SDK the client calls it internally inside search, it's only standalone here because client wiring is out of the PoC's scope. Added that explanation to the section in f0f9757.

lib/ts-sdk/package.json
SnowboardTechie: That cuts both ways: `example:transforms` already runs in the `ci` chain, which is the precedent this followed. Both examples carry assertions, so they've been doubling as smoke tests, but if the direction is pointing these at real endpoints later, I'm good pulling both out of `ci` now. Want me to drop both in this PR?

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
SnowboardTechie: Good call to flag this. The smell was hiding a real bug: the moneyComparison row had range operators and a numeric amount, while the schema takes gt/gte/lt/lte and a decimal-string amount. Rewrote it in 7ea731d as a single `FilterTypeMap` lookup, one row per filterType that reads directly against its Zod schema, and added two compile-error tests so the money row can't drift again.

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
SnowboardTechie: It can't fail, it's a placeholder. The real test is the type check: every `@ts-expect-error` in this file sits on a line that must fail to compile, and if a future type change makes that line legal, the directive becomes unused and `tsc` fails the build with ts2578. So `check:types` is what fails when the narrowing breaks, not vitest (vitest strips types and can't check them). The `expect(true)` lines only exist because vitest errors on a file in `__tests__/` that declares no tests. If we want this cleaner in the full SDK, vitest has a `--typecheck` mode built for exactly this; kept plain `tsc` for the PoC since the build already enforces it.

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
SnowboardTechie: Same deal as above, the `@ts-expect-error` is the test and `tsc` is the gate.

SnowboardTechie:
On the endpoint-routes question: deliberately left out. The PoC stops at the classifier boundary to retire the typed-narrowing risk; wiring `classifyFilters` into the Opportunity client routes is full-implementation work under #645. Pushed three commits for the inline threads: a dedicated `.int()` schema for integerComparison, the compile-time filter map corrected for the money types and collapsed into a single lookup table, and the docs updated (operator catalog, entry-point explanation, request-body wording). Details in the threads.

lib/ts-sdk/src/extensions/custom-filters.ts
jcrichlake: Sorry I didn't think about this earlier, but if this doesn't align with the spec documents on the website we should revert it back to `NumberComparisonFilter` since the spec itself is the source of truth.  Changes flow from the Spec to the SDK not the other way around.

lib/ts-sdk/package.json
jcrichlake: I would say yes

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
jcrichlake: Couldn't we slim this down to something like the below? 

`  // Checked by `tsc --noEmit` only. Each @ts-expect-error is the assertion.

  import type { TypedConsumerFilters } from "@/extensions";
  // ... minimal imports

  type GrantsGovSpecs = /* ... */;

  // @ts-expect-error -- unknown key (ts2353)
  const _1: TypedConsumerFilters<GrantsGovSpecs> = { notARealFilter: { operator: "eq", value: "x" } };

  // @ts-expect-error -- wrong operator for numberComparison (ts2322)
  const _2: TypedConsumerFilters<GrantsGovSpecs> = { awardCount: { operator: "like", value: 5 } };

  // ... etc`

lib/ts-sdk/src/extensions/README.md
jcrichlake: I'd say go ahead and drop both

lib/ts-sdk/package.json
SnowboardTechie: Removed in ce89092. The example still runs manually via `pnpm example:custom-filters` for when it gets reworked against a real endpoint.

lib/ts-sdk/__tests__/extensions/custom-filters-types.spec.ts
SnowboardTechie: Done in 940db0d: dropped the vitest wrapper and the widening demo block, kept the six @ts-expect-error assertions, and renamed the file from .spec.ts to .ts so vitest no longer collects it. `tsc --noEmit` is the only gate now, and the as-const trap is still covered in prose in the extensions README.

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: Checked the spec before touching anything: lib/core's numeric filters are NumberComparisonFilter / NumberRangeFilter / NumberArrayFilter only, no integer model on the website either. Reverted in 5cc2d1e, integerComparison points back at NumberComparisonFilterSchema and the int-enforcing schema is gone. If we want int enforcement later it can come in through a spec change first.

lib/ts-sdk/src/extensions/README.md
SnowboardTechie: Dropped both in f9bc818. Pulled the (PoC) markers from the two section headings and the TOC too, same logic.

jcrichlake (approved):

PR #885, opened 2026-06-09
Custom filters Python PoC (#869)

### LINKED ISSUE #869: [Py SDK] Draft custom filter PoC: Python
### Summary

Build a Python proof-of-concept for the custom filter interface. Validates the design shape recommended in SPIKE #646 before locking the full implementation in both SDKs.

Issued separately from the TypeScript PoC to avoid SDK dependencies between them.

### Acceptance criteria

- [x] PoC implements the custom filter design recommended in #646
- [x] PoC demonstrates a custom filter applied to a sample CommonGrants search request (customFilters property, standard Filter schema per ADR-0012)
- [x] Findings and design questions captured to inform full SDK implementation

SnowboardTechie:
### Summary

- Fixes #869 — the Python custom-filter PoC: validates the design recommended in SPIKE #646 (alongside counterpart TS PoC, #884) in the Python SDK — a custom filter on a sample CommonGrants search request (the `customFilters` property + standard Filter schema per ADR-0012) — and captures findings on the language-forced divergences to inform the full SDK implementation.
- Merges into the `HOLD-filters` bucket, parallel to the TS sibling PR #884. `lib/python-sdk/` only — no file overlap with #884.
- Time to review: ~30 minutes

### Changes proposed

**For plugin authors**
- Declare custom filters per resource/method — `define_plugin(routes={"opportunities": {"search": {"filters": {"agency": CustomFilterSpec(...)}}}})` — from a catalog of string/number/integer/boolean/date/money filter types.
- `PluginRoutes` mirrors the TS SDK shape exactly: per-method declarations sit in a `RouteDeclarations` `TypedDict` with an optional `filters` key, so the registration literal and the validation error paths (`routes.<resource>.<method>.filters.<name>`) are identical across both SDKs. Cross-SDK consumer testing showed why this matters: with the earlier Python-only shape (specs directly at the method level), a consumer porting that literal to the TS classifier had every registered filter silently downgraded to ad-hoc shape-only validation — wrong-value-type and unknown-type inputs were accepted instead of rejected.
- `validate_routes()` catches bad declarations at registration time: unknown filter types, and names that shadow a core default filter under either spelling.

**For SDK consumers**
- One flat `filters` dict per search call: default, registered-custom, and ad-hoc filters mix freely, snake_case or camelCase keys both work, and the classifier produces the correct ADR-0012 request body — no bucket bookkeeping at the call site.
- Anything invalid surfaces as a single catchable `PluginError` that names the failing filter, no matter which bucket it came from.
- The request body carries exactly the values that passed validation — booleans stay booleans (`true`, never `1`), and nothing unvalidated reaches the wire.
- `eq`/`neq` work on numeric comparison filters per protocol v0.3, so the same filter call accepts or rejects identically in both SDKs.
- `f` helpers (`f.eq`, `f.between`, `f.in_`, …) build filter values without hand-writing operator dicts.

**Runtime-vs-compile-time correctness (the central cross-SDK divergence)**
- The TS PoC's central proof is compile-time: `as const` narrowing rejects unknown keys and operator/value mismatches at `tsc` time. Python has no equivalent — correctness here is enforced **at runtime by Pydantic v2**. The pytest suite covers the same rejection matrix the TS `@ts-expect-error` tests cover, but as runtime `PluginError` assertions. Python adopters get runtime correctness; the compile-time DX story is TypeScript-specific.

**Findings captured for the full implementation**
- Pydantic-specific wire gotchas (alias-form construction, `mode="json"` serialization) are documented at their source in the module docs.
- One known value-strictness divergence remains (`"42"` as a number: TS rejects, Python coerces) — the protocol-level value types in #895 are the natural place to settle it.

No `lib/core` change; no changeset (PoC on a `HOLD-*` bucket, no production-dependency change — same as #884; re-evaluated when the bucket promotes).

### Context for reviewers

Follow-ups from this PoC pair are filed under the #645 epic: #895 (protocol-level boolean/integer filter models + value strictness), #896 (resource-agnostic classifier), #897 (escape-hatch semantics), #898 (ADR text corrections).

Verified with the real toolchain (from `lib/python-sdk/`):
- `make checks` — black, ruff, and mypy green.
- `make test` — 363 passed. The one failing test (`test_generate_models_typecheck_with_pyright_strict`) is pre-existing on the base branch and unrelated.
- `poetry run python examples/custom_filters.py` — runs green end-to-end, including the `except PluginError` demo.
- Exercised from a downstream consumer against the **wheel-installed** package shape (Poetry-built wheel in a fresh venv, not a source-path import).
- Cross-SDK parity exercised from the downstream-consumer harness: the Python and TS (#884) PoCs consume an identical shared fixture corpus through the wheel/tarball-installed packages, both run 5/5 green (2 accept + 3 adversarial reject cases), and their canonical-JSON output is byte-identical per case (recursive key sort + matched serializer settings, compare step exits 0). This run is what surfaced the routes-shape asymmetry that the `PluginRoutes` alignment above resolves.

### Additional information

Consumer-side classification (example output):
- `status` and `close_date_range` (normalized to `closeDateRange`) → top-level named fields (default bucket)
- `agency`, `fundingProgram` → `customFilters` (registered custom filters)
- `legacyTag` → `customFilters` (ad-hoc passthrough)

### DIFF
--- a/lib/python-sdk/common_grants_sdk/extensions/filters.py
+++ b/lib/python-sdk/common_grants_sdk/extensions/filters.py
@@ -38,6 +38,7 @@ from common_grants_sdk.schemas.pydantic.filters.money import (
     MoneyRangeFilter,
 )
 from common_grants_sdk.schemas.pydantic.filters.numeric import (
+    IntegerComparisonFilter,
     NumberArrayFilter,
     NumberComparisonFilter,
     NumberRangeFilter,
@@ -135,17 +136,16 @@ f = _FHelpers()
 # ---------------------------------------------------------------------------
 
 #: Maps each CustomFilterType to the Pydantic model used to validate operator/value shape.
-#: ``integerComparison`` reuses ``NumberComparisonFilter`` — a dedicated integer schema
-#: with int-only validation is deferred (mirrors the TS SDK, where
-#: ``integerComparison`` reuses ``NumberComparisonFilterSchema``).
-#: ``booleanComparison`` uses the SDK-level ``BooleanComparisonFilter``.
+#: ``integerComparison`` and ``booleanComparison`` use the SDK-level
+#: ``IntegerComparisonFilter`` / ``BooleanComparisonFilter`` models (parallel to the
+#: TS SDK's ``IntegerComparisonFilterSchema`` / ``BooleanComparisonFilterSchema``).
 FILTER_TYPE_SCHEMAS: dict[CustomFilterType, type[BaseModel]] = {
     CustomFilterType.STRING_COMPARISON: StringComparisonFilter,
     CustomFilterType.STRING_ARRAY: StringArrayFilter,
     CustomFilterType.NUMBER_COMPARISON: NumberComparisonFilter,
     CustomFilterType.NUMBER_ARRAY: NumberArrayFilter,
     CustomFilterType.NUMBER_RANGE: NumberRangeFilter,
-    CustomFilterType.INTEGER_COMPARISON: NumberComparisonFilter,  # reuse — see note above
+    CustomFilterType.INTEGER_COMPARISON: IntegerComparisonFilter,
     CustomFilterType.BOOLEAN_COMPARISON: BooleanComparisonFilter,
     CustomFilterType.DATE_COMPARISON: DateComparisonFilter,
     CustomFilterType.DATE_RANGE: DateRangeFilter,
--- a/lib/python-sdk/common_grants_sdk/extensions/filters.py
+++ b/lib/python-sdk/common_grants_sdk/extensions/filters.py
@@ -17,6 +17,8 @@ SDK's ``as const`` literal narrowing).
 
 from __future__ import annotations
 
+from collections.abc import Mapping
+from types import MappingProxyType
 from typing import Any, Optional
 
 from pydantic import BaseModel, ValidationError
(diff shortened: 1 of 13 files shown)


lib/python-sdk/common_grants_sdk/extensions/filters.py
```diff
@@ -0,0 +1,461 @@
...
+    loc = ".".join(str(part) for part in first["loc"]) or "<root>"
+    return f"{first['msg']} (at {loc})"
+
+
+# ---------------------------------------------------------------------------
+# validate_filter_call — call-time validator
+# ---------------------------------------------------------------------------
+
+
+def validate_filter_call(
```
jcrichlake, line 264: Should this be exported in the __init__.py for this module?

lib/python-sdk/common_grants_sdk/extensions/filters.py
```diff
@@ -0,0 +1,461 @@
...
+# classify_filters — ADR-0012 request-body classifier
+# ---------------------------------------------------------------------------
+
+
+def classify_filters(
+    routes: PluginRoutes,
+    resource: str,
+    method: str,
+    consumer_filters: Mapping[str, DefaultFilter | dict[str, Any]],
+) -> OppFilters:
```
jcrichlake, line 347: Could we update the docstring for this to flag it as something we'll change in the future

lib/python-sdk/common_grants_sdk/extensions/filters.py
```diff
@@ -0,0 +1,461 @@
...
+            # Bucket 1: core default filter.
+            # OppFilters requires the alias form (e.g. "closeDateRange") when constructing
+            # via **kwargs because populate_by_name is not set on OppDefaultFilters.
+            # Normalize: camelCase aliases stay as-is; snake_case keys are mapped to their
+            # alias; keys with no alias (e.g. "status") are passed through unchanged.
+            # No validate_filter_call here: default values are validated at the wrapped
+            # OppFilters construction below, against the named field's REAL type (e.g.
+            # status → StringArrayFilter) — stricter than the permissive DefaultFilter
+            # check, and the single validation point for this bucket.
+            opp_key = _SNAKE_TO_ALIAS.get(key, key)
```
jcrichlake: I'd change this variable name to be more generic like maybe alias_key or something. Less rework when we genericize this function

lib/python-sdk/common_grants_sdk/extensions/types.py
```diff
@@ -19,6 +19,27 @@
...
+# PluginRoutes = {resourceName: {methodName: RouteDeclarations}}.
+RouteMethodFilters = dict[str, CustomFilterSpec]  # {filterName: spec}
+
+
+class RouteDeclarations(TypedDict):
+    """Filter declarations for a single route method (e.g. ``search``).
+
+    Filters are keyed by name and the key is optional — a method may appear
+    in the route map with no filter declarations, and future declaration
+    kinds can ride alongside ``filters`` without a shape change. Mirrors the
```
jcrichlake: Can we update the language here, it feels clunky, `future declaration kinds`?

jcrichlake (changes requested):
Overall very nicely done! A few small tweaks but otherwise I think this looks good!

lib/python-sdk/common_grants_sdk/extensions/filters.py
SnowboardTechie: Yeah, good catch. The TS side exports validateFilterCall from its extensions index and the module docstring here already lists it as public surface. Added it to the `__init__` exports.

lib/python-sdk/common_grants_sdk/extensions/filters.py
SnowboardTechie: Updated, and anchored it to [#896](https://github.com/HHS/simpler-grants-protocol/issues/896) so the flag points at the actual follow-up (deriving the default-name set and wire body from the declared resource instead of the Opp binding).

lib/python-sdk/common_grants_sdk/extensions/filters.py
SnowboardTechie: Good call, renamed to alias_key.

lib/python-sdk/common_grants_sdk/extensions/types.py
SnowboardTechie: Good call, reworded: it now just states the shape (filters maps name to spec, and the key is optional) and drops the future-kinds clause.

PR #906, opened 2026-06-10
Hold transforms

jcrichlake:
### Summary

  - Fixes #825, #810, #838
  - Time to review: 45 minutes

  ### Changes proposed

  Implements the bidirectional plugin transform framework for both the TypeScript and Python SDKs per ADR-0022.

  **Core additions**
  - `buildTransforms` / `build_transforms` — compiles declarative mapping objects into typed `toCommon` / `fromCommon` callables, with built-in handlers (`field`, `const`, `match`,
  `numberToString`, `stringToNumber`) and support for custom handlers
  - `transformFromMapping` / `transform_from_mapping` — low-level recursive mapping walker used by `buildTransforms`
  - `definePlugin` / `define_plugin` extended to accept `mappings` (auto-wires `buildTransforms` at definition time) or explicit `toCommon`/`fromCommon` callables; providing both is a
  runtime error (XOR constraint)
  - Schema validation wrapping — both paths (mappings and explicit callables) validate `toCommon` output against the compiled common schema and `fromCommon` output against
  `sourceSchema` when provided

  **Type system**
  - `TransformError` — structured error carrying `path`, `handler`, `sourceValue`, and `cause`
  - `TransformResult<T>` — unconditional return shape for all transform callables; errors are aggregated rather than thrown
  - `SchemaOnly` / `SchemaWithTransforms` (TS) — named consumer-facing types distinguishing schema-only entries from entries with transforms
  - `SchemaInput` XOR union (TS) enforced at compile time; runtime validation in both SDKs
  - `PluginMeta.name` and `sourceSystem` / `source_system` made required (matching contract across SDKs)
  - Removed `merge-extensions` (superseded by the plugin framework)

  **Examples and docs**
  - New `examples/transforms.ts` and `examples/transforms.py` demonstrating full round-trip with custom handlers, extended schemas, and three-state null preservation
  - README updated in both SDKs with transform API docs and mapping format reference
  - ADR-0022 updated to reflect final decisions

  ### Context for reviewers

  The transform system intentionally does not auto-invert mappings — both `toCommon` and `fromCommon` must be author-provided, because many-to-one handlers (e.g. `match`/`switch`) are
  not reversible.

  The Python and TypeScript SDKs share the same conceptual contract but differ in one intentional way: Python uses a codegen step (`generate.py`) to bake `customFields` into typed
  Pydantic subclasses before runtime, so `SchemaConfig` does not expose `custom_fields` or `mappings` for inspection. In TypeScript everything compiles in memory at `definePlugin()`
  call time, so those fields are kept on the compiled entry.

  Both SDKs were verified with full check suites (lint, format, types) and all examples run end-to-end.


lib/ts-sdk/src/extensions/define-plugin.ts
```diff
@@ -4,41 +4,83 @@
...
+   * and `toCommon` / `fromCommon` callables — for each extensible model.
+   *
+   * This is the single surface for custom field declarations.
+   *
+   * `definePlugin()` compiles this into runtime schemas: `common` is built via
+   * `withCustomFields()` when `customFields` are declared; `toCommon` / `fromCommon`
+   * are auto-wired from `extensions.schemas[Name].mappings` when no explicit
+   * callables are provided. Native input Zod-wrapping remains deferred.
+   */
+  schemas?: T;
```
widal001: Hmm this doesn't seem right. We shouldn't have both an `extensions` property and a `schemas` property, especially if `PluginExtensions` also has a `meta`.

While I appreciate the attempt of providing a serializable plugin config, but I don't actually think that a goal for plugins, especially if that serializable config doesn't contain custom fields.

lib/python-sdk/common_grants_sdk/extensions/plugin.py
```diff
@@ -1,28 +1,148 @@
...
+    extensions: PluginExtensions | None = ...,
+    schemas: TSchemas = ...,
+) -> PluginConfig[TSchemas]: ...
+
+
+def define_plugin(
+    meta: PluginExtensionsMeta | None = None,
+    extensions: PluginExtensions | None = None,
+    schemas: Any = None,
+) -> PluginConfig[Any]:
```
widal001: It looks like we might be doing the same thing here with having both `schemas` and `extensions`. We'll want to drop this from both the TS and the Python SDK.

lib/ts-sdk/src/extensions/transformation.ts
```diff
@@ -0,0 +1,347 @@
...
+ *
+ * Cross-SDK divergence: Python's PoC uses bare `dict.get(val, default)`, which
+ * accepts non-string `val` natively and collapses `null` into `default`. This
+ * handler instead preserves `null` (three-state) and fails loud with a
+ * descriptive error on a malformed spec.
+ *
+ * `match` is the canonical handler name; `switch` is provided as a
+ * convenience alias — both point at the same handler function. (No prior SDK
+ * API defined `switch`; the alias is for ergonomic preference, not a
+ * backward-compatibility obligation.)
```
widal001: A lot of this stuff doesn't really belong in the docstrings for this function, especially since the TS SDK users don't really need to know or care about the implementation details of the Python SDK.

lib/ts-sdk/src/utils/transformation.ts
widal001, line 1: I'd consider moving this into a new `src/utils/` sub-directory, since this kind of mapping-based transformation is a pure function that could be used outside of plugins or extensions, for example in our website's form pre-fill playground.

lib/ts-sdk/src/extensions/transforms.ts
```diff
@@ -0,0 +1,275 @@
...
+ *   `PluginError` even when several fields would have failed.
+ * - Zod-validation failures (`commonModel` provided): every `ZodIssue` is
+ *   flattened into a separate `PluginError`, so `errors` carries the full
+ *   set.
+ *
+ * Callers writing strict-mode handling should treat any non-empty `errors`
+ * as failure regardless of length.
+ *
+ * @param toCommonMapping - ADR-0017 mapping from native format → CommonGrants.
+ * @param fromCommonMapping - ADR-0017 mapping from CommonGrants → native format.
```
widal001: We shouldn't be referencing ADRs in public-facing docstrings that are meant for SDK users

lib/ts-sdk/src/extensions/transforms.ts
```diff
@@ -0,0 +1,275 @@
...
+ */
+export function buildTransforms<TNative = unknown, TCommon = unknown>(
+  toCommonMapping: Record<string, unknown>,
+  fromCommonMapping: Record<string, unknown>,
+  handlers?: Map<string, Handler>,
+  // Bivariant `any` accepts schemas with input/output asymmetry; `unknown`
+  // would reject them at the contravariant input position.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  commonModel?: z.ZodType<TCommon, z.ZodTypeDef, any>
+): BuiltTransforms<TNative, TCommon> {
```
widal001: If we're not providing a `nativeModel` then we aren't validating the output of `fromCommon` matches that schema/model at runtime and we aren't really getting the benefit of `TNative`.

lib/ts-sdk/src/extensions/transforms.ts
```diff
@@ -0,0 +1,275 @@
...
+ * @throws Error when either mapping is structurally malformed (sibling keys
+ *   on a handler-dispatch node).
+ */
+export function buildTransforms<TNative = unknown, TCommon = unknown>(
+  toCommonMapping: Record<string, unknown>,
+  fromCommonMapping: Record<string, unknown>,
+  handlers?: Map<string, Handler>,
+  // Bivariant `any` accepts schemas with input/output asymmetry; `unknown`
+  // would reject them at the contravariant input position.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
widal001: Can we make comments like these more plain language? This doesn't really help me understand _why_ any is used

website/src/content/docs/governance/adr/0022-plugin-framework.mdx
```diff
@@ -20,7 +20,7 @@ _How should the Plugin object be structured to support both custom field declara
 ### Decision drivers
 
 - The framework must be implementable in both the Python and TypeScript SDKs with as consistent an interface as possible.
-- The `extensions` config must be serializable and able to pass validation (JSON-safe), and must be combinable across multiple extension packages via `mergeExtensions()` (TypeScript) / `merge_extensions()` (Python).
+- The `extensions` config must be serializable and able to pass validation (JSON-safe).
```
widal001: Having `extensions` be serializable isn't actually a goal, and it seems like this was the decision driver that led to the duplicate `schemas` vs `extensions` split.

Apologies if I introduced this language in some earlier draft, but I think a more accurate statement is that we'd like the inputs to `definePlugin()` to be declarative wherever possible (with optional overrides for things like custom code-driven transformation logic)

lib/ts-sdk/src/extensions/types.ts
```diff
@@ -117,46 +117,227 @@ export const EXTENSIBLE_SCHEMA_MAP = {
...
+ * Loose configuration object for plugin-provided HTTP clients.
  *
- * Each key is the name of a model that supports `customFields`, and the value
- * is a record mapping custom field keys to their `CustomFieldSpec` definitions.
- * The `Partial` type is used so plugins only need to declare models they actually extend.
+ * Loose per-plugin client configuration. The PoC does not constrain the shape;
+ * the full SDK's `client` capability work decides the precise type.
+ */
+export type ClientConfig = Record<string, unknown>;
+
```
widal001: Why did we include this in this PR for transformations? I'd suggest leaving out anything related to the client until we start working on the custom filters.

lib/ts-sdk/src/extensions/types.ts
```diff
@@ -117,46 +117,227 @@ export const EXTENSIBLE_SCHEMA_MAP = {
...
+ * provenance string. `version` and `capabilities` remain optional because
+ * they can be inferred or omitted during early development.
+ */
+export interface PluginMeta {
+  /** Plugin display name (e.g. `"grants.gov"`). */
+  name: string;
+  /** Plugin version (semver, e.g. `"1.0.0"`). */
+  version?: string;
+  /** Name of the native source system (e.g. `"grants.gov"`). */
+  sourceSystem: string;
```
widal001, line 354: If we use the term `sourceSystem` here, I'd recommend being consistent when naming other things that reference this system like `sourceSchema` instead of `native`, etc.

lib/ts-sdk/src/extensions/types.ts
```diff
@@ -117,46 +117,227 @@ export const EXTENSIBLE_SCHEMA_MAP = {
...
+
+/**
+ * Serializable portion of plugin config — safe to store as JSON.
+ *
+ * Used to store JSON-safe serializable config alongside a plugin.
+ */
+export interface PluginExtensions {
+  meta?: Partial<PluginMeta>;
+  schemas?: Partial<Record<ExtensibleSchemaName, PluginExtensionsObjectConfig>>;
+}
```
widal001: If we're using `schema` most other places, including `ExtensibleSchemaName`, I'd be hesitant to introduce a new concept of `Object`

widal001 (commented):
@jcrichlake @SnowboardTechie It seems like y'all have made some really great progress on this! I left a few detailed comments below, but there's a couple bigger shifts my suggestions revolve around:

1. Collapsing `extensions` and `schemas` in `definePlugin()` so there's only one right way to define the inputs to a given schema, with XOR logic that expects _either_ `mappings` _or_ custom `toCommon()` and `fromCommon()` functions directly
2. Improving the runtime validation and type inference for `fromCommon()` by accepting a `sourceSchema` in `buildTransforms()`, similar to what is currently supported for `toCommon()`
3. Distinguishing between schemas that do and don't have transformation logic on them, so that we can narrow the type, avoid having to do `schemas.[schemaName].toCommon?() = result` and then `if (!result) {}`
4. Standardizing naming so we're consistent across types, function params, at the same conceptual level (e.g. Plugin, Schema, Transform, Source, Common)
5. Applying some additional compile time checks to enforce the use of `as const` to prevent accidental type widening if people forget.

Here's a more detailed summary of those suggested changes [transforms-review.md](https://github.com/user-attachments/files/28866591/ts-sdk-transforms-review.md)

And here's a simplified proof of concept that demonstrates how the type inference would work for the approach I'm proposing in suggestion 1 above [ts-sdk-transforms-typecheck-poc.ts](https://github.com/user-attachments/files/28866594/ts-sdk-transforms-typecheck-poc.ts)

While most of my comments and the sample focus on TypeScript, the general changes also mostly apply to Python, except for the nuanced type inference work, since Python's codegen both makes this harder and easier.

.changeset/transforms-poc-python.md
```diff
@@ -0,0 +1,33 @@
+---
+"common-grants-sdk": minor
+---
+
+Add a Python proof-of-concept for the plugin transformation framework (issue #799), mirroring the TypeScript PoC. Plugin authors can now compile declarative mapping dicts into typed `(to_common, from_common)` callables, validate `to_common` output against an extended Pydantic schema, and attach those callables to a plugin via `define_plugin(schemas=...)`.
```
widal001, line 5: Do we still want to call this a PoC? And it also allows plugin authors to provide custom `to_common` and `from_common` functions

.changeset/transforms-poc-typescript.md
```diff
@@ -0,0 +1,32 @@
+---
+"@common-grants/sdk": minor
+---
+
+Add a TypeScript proof-of-concept for the plugin transformation framework (issue #798), mirroring the Python PoC in PR #810. Plugin authors can now compile declarative mapping objects into typed `(toCommon, fromCommon)` callables, validate `toCommon` output against an extended Zod schema, and attach those callables to a plugin via `definePlugin({ schemas })`.
```
widal001, line 5: Same with this summary

widal001 (approved):
Excited to get this released!!

PR #916, opened 2026-06-15
refactor(ts-sdk): update transform helper types

### LINKED ISSUE #888: [SDK] Review transforms
### Summary

Transforms work is complete in both SDKs but is currently unmerged. Billy gives a final review before a new SDK version ships.

### Acceptance criteria

- [x] Transforms changes reviewed

widal001:
### Summary

Updates `ToCommon`/`FromCommon` helper types so that plugin authors don't need a fully resolved `CommonSchema` in order to annotate their custom `toCommon()` and `fromCommon()` functions.

- Fixes #888 
- Time to review: 2 minutes

- **Accepts `Date` as input to datetime types**
  - Adds a pre-processing step to date types (e.g. `IsoDateSchema`) so that both string (`"2024-12-01"`) and date inputs (`Date(2024, 12, 1)`) are accepted
  - This allows us to wrap the `toCommon()` function with an extra `commonSchema.safeParse()` check (which ensures that the function returns a valid type) even if the `toCommon()` function already returns a `Date()` type for fields like `createdAt` and `lastModifiedAt`
- **Updates `ToCommon`/`FromCommon` helper types**
  - Provides a `TransformTypes` with named inputs `{ model, sourceSchema, cusomFields }` that replaces the positional args in `ToCommon` and `FromCommon`
  - This allows plugin authors to simply provide `ToCommon` and `FromCommon` the same custom field definitions they're providing to `definePlugin()` instead of having to generate a fully resolved `commonSchema` just for the signature
- **Improves examples and docs**
  - Updates `examples/transforms.ts` to use `definePlugin()` with the `mapping` option rather than invoking `buildTransforms()` directly (since this happens behind the scenes)
  - Adds a few other transformation examples (e.g. providing custom transform functions, demonstrating compile-time checks, etc.)
  - Adds extra explanation around when and why `as const` is required to the README.
- **Requires `sourceSchema` when `mappings` or `toCommon`/`frommCommon` is provided:** This allows us to always validate the output `fromCommon()` matches `sourceSchema` and avoids propagating undefined checks on `sourceSchema` when a schema supports transformations.

Most of these are pretty neutral changes that just improve type inference and/or help plugin authors avoid having to generate their own `CommonSchema` to annotate custom transformation functions.

But the bigger change to the interface that's worth discussing is the shift to requiring `sourceSchema` when defining transformations (either via `mappings` or `toCommon`/`fromCommon`). Personally I think this is expected behavior: If a plugin is going to provide transformations, we should know what it's transforming to/from, and the workaround for plugin authors if they don't want to have to provide an explicit source schema is `z.object({}).passthrough()`. But we can talk through this further if y'all disagree.

Screenshot 2026-06-15 at 6 52 39 PM

### DIFF
--- a/lib/ts-sdk/__tests__/extensions/define-plugin.spec.ts
+++ b/lib/ts-sdk/__tests__/extensions/define-plugin.spec.ts
@@ -207,7 +207,7 @@ describe("definePlugin", () => {
 
       const plugin = definePlugin({
         schemas: {
-          Opportunity: { toCommon, fromCommon },
+          Opportunity: { sourceSchema: z.object({}).passthrough(), toCommon, fromCommon },
         },
       });
 
@@ -234,7 +234,7 @@ describe("definePlugin", () => {
 
       const plugin = definePlugin({
         schemas: {
-          Opportunity: { toCommon, fromCommon },
+          Opportunity: { sourceSchema: z.object({}).passthrough(), toCommon, fromCommon },
         },
       });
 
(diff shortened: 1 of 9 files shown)


lib/ts-sdk/__tests__/extensions/define-plugin.spec.ts
```diff
@@ -253,6 +253,59 @@ describe("definePlugin", () => {
       expect(plugin.schemas.Opportunity.toCommon).toBeUndefined();
       expect(plugin.schemas.Opportunity.fromCommon).toBeUndefined();
     });
+
+    it("does not expose transform callables as callable on a schema-only entry", () => {
```
widal001, line 257: This and the following test validate compile-time checks that distinguishing between `SchemaOnly` vs. `SchemaWithTransforms` allows us to provide plugin consumers

lib/ts-sdk/__tests__/schemas/zod/types.spec.ts
```diff
@@ -163,11 +163,16 @@ describe("ISODate Schema", () => {
...
+    expect(parsed).toBeInstanceOf(Date);
+    expect(parsed.getUTCFullYear()).toBe(2025);
+    expect(parsed.getUTCDate()).toBe(15);
+  });
+
   it("should raise an error for an invalid ISODate", () => {
     // Must be in YYYY-MM-DD format
     expect(() => ISODateSchema.parse("01-01-2025")).toThrow();
-    // Must be a string, not a Date object
-    expect(() => ISODateSchema.parse(new Date())).toThrow();
```
widal001, line 170: Now dates are explicitly accepted, after being cast to strings via `.toISOString()` in pre-processing

lib/ts-sdk/src/extensions/transform-helpers.ts
widal001, line 1: These types basically allow plugin authors who want to hand write their own transformation functions to annotate them with `ToCommon` and `FromCommon` by providing `customFields` instead of having to use those custom field specs to first generate a `CommonSchema` (which is what `definePlugin()` is already doing for them)

lib/ts-sdk/src/extensions/README.md
```diff
@@ -92,7 +93,7 @@ opportunity.customFields?.category?.value; // string
 
 **Key points:**
 
-- Pass `as const` to the specs object so TypeScript can infer literal `fieldType` values and preserve the specific keys.
+- `as const` on the specs is only needed when you assign them to a variable before the call; inline specs are inferred correctly without it. Forgetting it on a hoisted variable is a compile error. See [When you need `as const`](#when-you-need-as-const) for details.
```
widal001, line 96: I dug a bit deeper into the nuances of when `as const` was needed and clarified that in the README

jcrichlake (approved):

PR #919, opened 2026-06-16
chore(deps): remediate high-severity transitive audit advisories

SnowboardTechie:
### Summary

- Clears the high-severity `pnpm audit` gate that fails the `validate` job on every dependency PR (5 high-sev advisories in transitive/website deps).
- Unblocks #917 and #914 — both are clean bumps that only fail on this shared gate.
- No changeset — only the unpublished `website` package and transitive deps change.
- Time to review: ~5 minutes

### Changes proposed

- **`pnpm-workspace.yaml`** — three new floor-overrides for transitive deps the version-update bots can't reach, plus a fix to one stale exact-pin:
  - `esbuild` → `>=0.28.1` (GHSA-gv7w-rqvm-qjhr — RCE via `NPM_CONFIG_REGISTRY`)
  - `form-data` → `>=4.0.6` (GHSA-hmw2-7cc7-3qxx — CRLF injection)
  - `ws` → `>=8.21.0` (GHSA-96hv-2xvq-fx4p — memory-exhaustion DoS)
  - `vite` 6.x: `'6.4.2'` exact-pin → `>=6.4.3` floor (GHSA-fx2h-pf6j-xcff — `server.fs.deny` bypass; the pinned `6.4.2` had itself become vulnerable)
- **`website/package.json`** — `astro` `^6.4.5` → `^6.4.6` (GHSA-2pvr-wf23-7pc7 — host-header SSRF). `astro` is a direct dependency, so it's bumped in place rather than overridden.
- **`pnpm-lock.yaml`** — regenerated (net reduction as the floors dedupe stale transitive copies).

### Context for reviewers

The split is deliberate: the three transitive deps aren't declared in any manifest, so neither Dependabot nor the catalog workflow can bump them — overrides are the only lever, written as floors (`<vulnerable>: '>=patched'`) so versions still float upward freely and don't lock to stale. `astro` is a direct `website` dep, so it's bumped directly.

The `vite` change also fixes a latent problem worth noting: the previous `vite@…: '6.4.2'` was an exact pin, and a later advisory (`<=6.4.2`) grew to cover it — so the pin was actively holding vite on a vulnerable version. Rewriting it as a floor lets it patch forward.

Verified locally (the `validate` audit is scoped to the `website` surface):

- `node .github/scripts/audit-deps.js --level high --filter website` → `No advisories at or above "high" severity` (694 packages; previously 5 highs).
- `pnpm --filter website build` → 150 pages built, all internal links valid.

Resolved versions after the change: astro 6.4.7, form-data 4.0.6, ws 8.21.0, esbuild 0.28.1, and no vulnerable vite ≤6.4.2.

### Additional information

The `astro` bump is the safe half of #912; with it landed here, #912 is left carrying only the Starlight 0.40.0 upgrade, which is blocked on an upstream type incompatibility (Starlight 0.40.0 ships source that doesn't typecheck under astro 6.4.x).


jcrichlake (approved):
LGTM

PR #920, opened 2026-06-16
fix(ts-sdk): guard date schemas against invalid Date input

SnowboardTechie:
### Summary

- Follow-up to #916: an invalid `Date` passed to the new date schemas throws a `RangeError` past `safeParse` instead of coming back as a validation error.
- Time to review: 2 minutes

- **Guard `acceptDate` against an invalid `Date`.** It called `toStr` (`.toISOString()`) on any `Date` instance, but an invalid `Date` (e.g. `new Date("garbage")`) is still `instanceof Date`, so `toISOString()` threw a `RangeError` from inside `z.preprocess`. That escapes `safeParse`, which is meant to be total, so an invalid date in a transform's output crashes instead of landing in `TransformResult.errors`. The guard (`!isNaN(val.getTime())`) routes an invalid `Date` to the string validator, which rejects it as a normal validation error.
- **Regression tests** for `ISODateSchema` and `OffsetDateTimeSchema` asserting `safeParse(new Date("not a date"))` returns `success: false` rather than throwing.

- Verified with the SDK toolchain: `pnpm typespec` + `pnpm test` (28 passed), `tsc --noEmit` clean, `eslint` / `prettier` clean. The two new tests fail against the pre-fix code with `RangeError: Invalid time value` and pass with the guard, so they discriminate.
- No changeset: the affected `acceptDate` code is unreleased (lives only on `HOLD-transforms`) and #916 added none, so this rides under the existing `transforms-poc-typescript.md` when the bucket merges. Happy to add one if per-fix granularity is preferred.

Before vs after, against the SDK's own zod (`ISODateSchema`):

```
=== current (HOLD-transforms) ===
  invalid Date  new Date('garbage'):  THREW past safeParse -> RangeError: Invalid time value
=== with guard ===
  invalid Date  new Date('garbage'):  safeParse returned error -> Expected string, received date
```

A valid string (`"2024-12-01"`) and a valid `Date` parse fine either way.

### DIFF
--- a/lib/ts-sdk/__tests__/schemas/zod/types.spec.ts
+++ b/lib/ts-sdk/__tests__/schemas/zod/types.spec.ts
@@ -170,6 +170,14 @@ describe("ISODate Schema", () => {
     expect(parsed.getUTCDate()).toBe(15);
   });
 
+  it("returns a validation error (not a thrown RangeError) for an invalid Date", () => {
+    // Regression: an Invalid Date is still `instanceof Date`, so toISOString()
+    // throws a RangeError inside preprocess that escapes safeParse. The guard
+    // routes an invalid Date to the string validator so safeParse stays total.
+    expect(() => ISODateSchema.safeParse(new Date("not a date"))).not.toThrow();
+    expect(ISODateSchema.safeParse(new Date("not a date")).success).toBe(false);
+  });
+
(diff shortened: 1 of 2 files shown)


widal001 (approved):
Thanks for catching this!

PR #924, opened 2026-06-16
refactor(py-sdk): adopt new plugin framework without codegen

### LINKED ISSUE #888: [SDK] Review transforms

widal001:
### Summary

Shifts the Python SDK to a new plugin framework that doesn't require a separate codegen step by leveraging Pydantic generics and sub-classes instead.

- Fixes #888 
- Time to review: 20 minutes

- Introduces a new Pydantic-based plugin framework and authoring DX
  - Introduces a few [new concepts](https://github.com/HHS/simpler-grants-protocol/blob/d2b8f7911fd92e39f2baf34d01d863d03a6e8624/lib/python-sdk/common_grants_sdk/extensions/README.md#key-concepts) to facilitate authoring `Plugins` without codegen, starting with the biggest changes:
    - `CustomField[V]` a pydantic generic that wraps a custom field value in a custom field pydantic schema. This largely replaces the `CustomFieldSpec` that the previous version and the TypeScript SDK use. While it's a pretty big shift, it follows a much more Pythonic approach to defining custom fields and is the cornerstone for the codegen-free Plugin type interference.
    - `CustomFieldSet` a pydantic base class that an author uses to declare a set of custom fields on an extensible schema. This essentially replaces the dictionary that mapped custom field keys to custom fields spec per schema (`"Opportunity": { "legacyId": CustomFieldSpec(...) }` e.g. `OppCustomFields[CustomFieldSet]`. Also a big shift, but again it aligns with a more Pythonic and pydantic-specific pattern
    - `schema(...)` factory function that accepts a sub-class of `CustomFieldSet`, along with `source_schema`, and `mappings` XOR `to_common`/`from_common` this is super close to the [`SchemaInput`](https://github.com/HHS/simpler-grants-protocol/blob/d2b8f7911fd92e39f2baf34d01d863d03a6e8624/lib/ts-sdk/src/extensions/types.ts#L285-L301) type in the TS SDK.
    - `PluginSchemas` basically the new equivalent of [`PluginSchemaInput`](https://github.com/HHS/simpler-grants-protocol/blob/d2b8f7911fd92e39f2baf34d01d863d03a6e8624/lib/ts-sdk/src/extensions/define-plugin.ts#L32) in the TS SDK, a class where the attributes are schemas that can be extended
    - `define_plugin()` Same name, but slightly different interface and now returns a `Plugin` directly instead of producing a config that needed to be converted into a `Plugin` via codegen.
- Removes the old codegen step and a lot of the associated tests.
- Introduces a new [`plugins.py`](https://github.com/HHS/simpler-grants-protocol/blob/d2b8f7911fd92e39f2baf34d01d863d03a6e8624/lib/python-sdk/examples/plugins.py) example that matches the [`transforms.ts`](https://github.com/HHS/simpler-grants-protocol/blob/d2b8f7911fd92e39f2baf34d01d863d03a6e8624/lib/ts-sdk/examples/transforms.ts) example in the TS SDK and highlights four common plugin authoring scenarios. **NOTE:** We might want to align the names of these examples/files but I didn't want to replace the existing `transforms.py` which seems to be demonstrating something different.

#### Instructions to review

- I'd mostly suggest reviewing [`plugins.py`](https://github.com/HHS/simpler-grants-protocol/blob/d2b8f7911fd92e39f2baf34d01d863d03a6e8624/lib/python-sdk/examples/plugins.py)
- Run it as is with `poetry run python examples/plugins.py`
- Make some direct changes to that file to see how this new framework impacts the plugin author and consumer experience. Such as:
  - Adding a new custom field
  - Adding a completely new plugin
  - Changing the transformation functions

#### Notes

- There are two big tradeoffs with this approach:
  - It shifts away from `CustomFieldSpec` and replaces it with `CustomField[V]` and `OppCustomFields[CustomFieldSet]` (authors can use their own name for `OppCustomFields`)
  - This makes it diverge with the runtime `with_custom_fields()` and with the TypeScript SDK syntax -- though it's worth noting that `with_custom_fields()` could be replaced with this pattern as well.
- But the upsides are pretty big both for overall conceptual alignment with the TS SDK and pythonic DX
  - Now `define_plugin()` and `definePlugin()` both produce a fully-typed `Plugin`, without an extra build step needed, which also allows multiple plugins to be defined in the same file (same with TypeScript)
  - We have 1000+ fewer LOCs to maintain, and provide plugin authors with greater mypy/pyright type errors as well as import errors if they define their plugins incorrectly
  - It's waaaay easier to provide hand-written `to_common` and `from_common` functions in a type safe way, which is what I imagine most plugin authors will do -- including us if we want to move the [transformation logic in `simpler-grants-gov`](https://github.com/HHS/simpler-grants-gov/blob/b1dfcdf770b9d9a1cc4b695e3f1ff0c71a1cd6a0/api/src/services/common_grants/transformation.py#L1) to a plugin
  - It's also a fair amount easier to reason about a plugin and how it was constructed. Arguably even easier than the TypeScript SDK.

#### Plugin author DX
Screenshot 2026-06-16 at 2 41 11 PM
Screenshot 2026-06-16 at 2 41 55 PM

#### Plugin consumer DX
Screenshot 2026-06-16 at 2 43 48 PM
Screenshot 2026-06-16 at 2 44 32 PM

#### Test run of file
Screenshot 2026-06-16 at 2 40 44 PM


lib/python-sdk/common_grants_sdk/schemas/pydantic/models/opportunity.py
```diff
@@ -0,0 +1,36 @@
...
+
+from pydantic import ConfigDict, Field
+
+from .opp_base import OpportunityBase
+
+CF = TypeVar("CF")
+
+
+class Opportunity(OpportunityBase, Generic[CF]):
+    """The common Opportunity model, generic over its custom-fields container.
```
widal001: This is one model that we could potentially remove if we pulled the `Generic[CF]` into the `OpportunityBase` itself. 

Or we might want to find a different place to define these since we'll need one per extensible schema.

lib/python-sdk/common_grants_sdk/extensions/types.py
```diff
@@ -88,58 +84,8 @@ class PluginExtensionsMeta(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
 
     name: str
-    source_system: str = Field(alias="sourceSystem")
+    source_system: str = Field(
+        validation_alias="sourceSystem", serialization_alias="sourceSystem"
+    )
```
widal001, line 89: The switch from `alias=` to this (`validation_alias=` and `serialization_alias=`) solves a Pyright/Pylance issue that shows a type issue when `alias=` alone is used, because Pyright/Pylance expects consumers of this model to use the alias rather than the attribute name.

jcrichlake:
Some top level questions/observations I had on this.
1. Do we want to leave the old code gen pathway in place for now? 
2.  The previous item gives us two classes named CustomField.  `opp_base.py` uses the old custom field. 
3. There's no runtime guard when `to_common` is passed without source schema.  Should this be required?
4.  `transform_frommapping({})` returns `None` instead of `{}`

Minor nit
- `test_plugin.py` doesn't test the schema_name mismatch path in define plugin

jcrichlake (approved):
Looks good on my end! Thanks for making these improvements!!

SnowboardTechie (approved):

PR #928, opened 2026-06-18
spike(sdk): classify custom filters behind opportunities.search() in both SDKs

### LINKED ISSUE #897: [SDK] Decide escape-hatch semantics for the custom-filter classifier
### Summary

Three classifier behaviors are shared by both PoCs (#884, #885) — deliberately kept symmetric rather than fixed one-sided — and each needs a decision during full implementation:

1. **`customFilters` is not a reserved name.** The default-name collision set is built from the default-filters schema, which doesn't include the `customFilters` key itself — so it can be registered as a custom filter or passed as a consumer key, nesting `customFilters.customFilters` in the wire body.
2. **Unknown resource/method silently downgrades.** Registered filters for an undeclared resource/method fall back to ad-hoc passthrough (permissive validation instead of the registered spec), with no signal.
3. **Ad-hoc routing is unobservable.** A typo'd default-filter key (e.g. `statuss`) silently lands in `customFilters`; a server that ignores unknown custom keys returns unfiltered results that look valid. No client error, no warning, nothing in the return value distinguishes "escape hatch" from "matched."

The passthrough itself is intentional per ADR-0012 (servers report unsupported filters in `filterInfo.errors`); these are decisions about client-side guardrails around it — keep, guard, or warn, decided once for both SDKs.

### Acceptance criteria

- [x] A decision is recorded for each of the three behaviors (keep as-is / guard / warn)
- [x] The decided behavior is implemented identically in TypeScript and Python
- [x] Tests in both SDKs cover the decided wire-body behavior

SnowboardTechie:
### Summary

- Spike: classify custom filters inside `opportunities.search()` in both SDKs, so consumers pass a flat `filters` argument instead of calling `classifyFilters` / `classify_filters` by hand. `routes` is bound at client construction.
- Fixes: #897 
- Relates to #645 (client wiring). Base branch is `HOLD-filters`; spike, not a release.
- Time to review: ~15 minutes.

### Changes proposed

- **Classification behind `search()`** (both SDKs): a flat `filters` argument is classified into the `OppFilters` wire body inside the client — default fields top-level, registered + ad-hoc filters under `customFilters`.
- **`routes` is client-bound**: `new Client({ routes })` (TS) / `Client(routes=…)` (Python), not per-call — it's fixed plugin config. TS `Client`/`Opportunities` are generic over the routes type so `search({ filters })` keeps filter-name autocomplete.
- **Validation is fail-soft**: `search()` never throws on a call-time filter problem. A filter that fails validation (default field against its real type, registered against its `filterType`, ad-hoc against the `{operator, value}` shape) is dropped from the request body and its error collected into the response's `filterInfo.errors` — mirroring ADR-0022's transform `{ result, errors }` model. A `status` collision (shorthand + `filters.status`) resolves `filters`-wins with a collected warning. Registration-time `validateRoutes` still throws. Strictness is per engine (Zod strict, Pydantic lax-coerces) — identical wire contract for valid input.
- **Python `filterInfo` / `sortInfo` surfaced**: the client now returns the server's response envelope (incl. `filterInfo.errors`) instead of empty placeholders, matching TS. `list` unchanged.

### Context for reviewers

- Back-compatible on the no-`filters` path. Python `search(status=…)` is now optional.
- Typed narrowing (TS) gives autocomplete + a `{ operator, value }` envelope check, not typo-rejection (ad-hoc keys are an open set); per-type value validation runs at runtime in `classifyFilters` (fail-soft — collected, not thrown).
- Client-side filter errors merge into the response's `filterInfo.errors` (one channel) flattened as `"path: message"`, ahead of any server-provided errors; consumers apply their own strict-vs-lenient rule (the same adopter choice ADR-0022 gives for transforms).
- Verified: TS `tsc` + `vitest` 521/521; Python `black` / `ruff` / `mypy` clean, `pytest` 381/381.

### Additional information

Throw-vs-collect (relates to #897) is settled in favor of **collect**: filter validation is fail-soft, surfacing errors in `filterInfo.errors` rather than throwing, consistent with ADR-0022's transform model. ADR-0022 scopes custom filters out, so this was a consistency choice, not a conformance fix. Registration-time `validateRoutes` stays fail-fast — a malformed plugin declaration has no result to return.


jcrichlake:
Small nit on the PR template, we usually have a `Fixes: ` field.  I think that is used to auto close issues on the project board.  But I'm not 100 percent sure on that

lib/python-sdk/common_grants_sdk/client/client.py
```diff
@@ -21,17 +23,27 @@ def __init__(
...
     ):
         """Initialize the CommonGrants client.
 
         Args:
             config: Optional Config instance. If None, a default Config is created.
             auth: Optional Auth instance. If None, API key authentication is used
                 with the key from config.
+            routes: Optional plugin ``routes`` declaration (fixed plugin config).
+                Bound and validated once here; used to classify registered custom
+                filters in ``opportunities.search``.
```
jcrichlake, line 36: Is this built in a way where it can support any of the possible routes in a client?

lib/python-sdk/common_grants_sdk/client/client.py
```diff
@@ -21,17 +23,27 @@ def __init__(
...
+
+        Raises:
+            FilterError: If ``routes`` declares an unknown ``filter_type`` or a
```
jcrichlake: Should there be another error if a route is given that isn't supporting custom filters? For example what if list is passed in?

lib/python-sdk/common_grants_sdk/client/client.py
```diff
@@ -182,7 +194,10 @@ def search(
                 path, json=request_data, params={"page": page, "pageSize": page_size}
```
jcrichlake, line 197: Shouldn't this have filters in it?

lib/python-sdk/common_grants_sdk/client/opportunities.py
jcrichlake, line 189: How is a user passing in the filters they wish to use to the client which is what eventually makes the call?

SnowboardTechie:
@jcrichlake ty for the catch! I added that to the description.

lib/python-sdk/common_grants_sdk/extensions/filters.py
```diff
@@ -411,52 +448,54 @@ def classify_filters(
             # via **kwargs because populate_by_name is not set on OppDefaultFilters.
             # Normalize: camelCase aliases stay as-is; snake_case keys are mapped to their
             # alias; keys with no alias (e.g. "status") are passed through unchanged.
-            # No validate_filter_call here: default values are validated at the wrapped
-            # OppFilters construction below, against the named field's REAL type (e.g.
+            # Validate per-field against the named field's REAL type (e.g.
             # status → StringArrayFilter) — stricter than the permissive DefaultFilter
-            # check, and the single validation point for this bucket.
+            # check. Fail-soft: an invalid value is collected and skipped.
             alias_key = _SNAKE_TO_ALIAS.get(key, key)
```
jcrichlake, line 594: I know this isn't changed, but is this all caps for any specific reason?

lib/python-sdk/common_grants_sdk/extensions/types.py
```diff
@@ -105,6 +117,25 @@ def __init__(
...
+    into ``errors``.
+
+    result: classified filters, containing only the keys that passed validation.
+    errors: aggregated ``FilterError``s for keys that failed; empty on success.
+
+    Registration-time validation (``validate_routes``) still raises — a malformed
+    plugin declaration has no result to return.
+    """
+
+    result: "OppFilters"
```
jcrichlake: Should this be a string? Also do we want it to be specifically OppFilters or a higher level pydantic object that will be re-used across different models?

lib/python-sdk/tests/client/test_client.py
```diff
@@ -1341,6 +1408,13 @@ def test_search_all_handles_error_on_subsequent_page(self, sample_item_data):
                     "totalItems": 5,
                     "totalPages": 3,
                 },
+                "sortInfo": {
+                    "sortBy": "lastModifiedAt",
+                    "sortOrder": "desc",
+                    "customSortBy": None,
+                    "errors": [],
+                },
+                "filterInfo": {"filters": {}, "errors": []},
```
jcrichlake, line 1417: None of these updates contain a populated filter object. Is that correct?

lib/ts-sdk/src/client/opportunities.ts
```diff
@@ -38,6 +64,39 @@ import { ArrayOperator } from "../constants";
...
+
+/** Raw `{ operator, value }` filter object, as produced by the `F.*` helpers. */
+type RawFilter = { operator: string; value: unknown };
+
+/**
+ * The declared custom-filter names for `opportunities.search` in a routes type.
+ * `definePlugin` preserves the literal `routes` type (its `const TRoutes` generic),
+ * so a plugin defined inline yields concrete filter-name literals here.
+ */
+type SearchFilterNames<R extends PluginRoutes> = R extends {
```
jcrichlake: Should this have Search included in its name? Or should it just be something like `CustomFilterNames`?

lib/ts-sdk/src/client/opportunities.ts
```diff
@@ -236,22 +305,24 @@ export class Opportunities {
...
-    options?: SearchOptions<S>
+    options?: SearchOptions<S, R>
   ): Promise<Filtered<z.infer<S>, OppFilters>> {
     const schema = options?.schema ?? (OpportunityBaseSchema as unknown as S);
 
-    // Build the base search body (without pagination)
-    const searchBody = this.buildSearchBody(options);
+    // Build the base search body (without pagination). Client-side filter
+    // problems are collected fail-soft into `filterErrors` rather than thrown.
+    const { body: searchBody, errors: filterErrors } = this.buildSearchBody(options);
```
jcrichlake, line 314: I'm wondering if we should make the fail-soft option configurable or not? We don't have to address it here, but raising this up as something we may want to consider if these API calls get expensive for consumers and providers. 

Especially when considering filters should in theory reduce the amount of data returned to the caller so as to lighten the load.

lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -189,35 +189,39 @@ export function validateRoutes(routes: PluginRoutes): void {
...
       );
     }
-    return;
+    return undefined;
   }
 
   // Registered filter — validate against the filterType's schema
   const schema = FILTER_TYPE_SCHEMAS[spec.filterType];
   if (!schema) {
     // Should not reach here if validateRoutes was called first, but guard anyway
```
jcrichlake, line 242: This implies there is a have to know about this call.  Is there a way to have validateRoutes be called inside of this validateFilterCall?

jcrichlake (commented):
Really nice work on the changes, a few comments and some questions to gain a better understanding.

lib/python-sdk/common_grants_sdk/client/client.py
SnowboardTechie: Yeah, the routes mapping takes any `resource.method` key, so structurally it's open. The validation scopes custom filters to *filter-capable* routes (the ones whose contract operation declares a `filters` param), which today is just `opportunities.search`. It's not opportunities-only by design, it's "wherever the contract supports filters," and `opportunities.search` is simply the first. As we add filter support to other routes, they join the set.

lib/python-sdk/common_grants_sdk/client/client.py
SnowboardTechie: Good call. It does now: `validate_routes` raises `FilterError` at construction when custom filters are declared on a route that isn't filter-capable. `opportunities.list` (no `filters` param in its contract) throws `Route "opportunities.list" does not support custom filters (supported: opportunities.search)`. The supported set is a deliberate allowlist for now, mirroring the contract's filter-capable routes; longer term it'd be cleaner to derive that from the contract directly so new routes don't need a manual entry. Both SDKs, regression test each.

lib/python-sdk/common_grants_sdk/client/client.py
SnowboardTechie: They're already in there. By the time it reaches `Client.search`, `request_data` is the fully-assembled body with filters included (the resource method builds it before calling down). This method is the generic transport, so it stays filter-agnostic on purpose. I added a one-line note here pointing that out, since it's easy to miss reading top-down.

lib/python-sdk/common_grants_sdk/client/opportunities.py
SnowboardTechie: Through the `filters=` param on `opportunities.search`: a flat `{name: {operator, value}}` dict (the `F.*` helpers build the entries). It gets classified and assembled into the request body before the client posts. The thread's gone outdated since, and the param + its docstring cover it now, so I'll resolve this one.

lib/python-sdk/common_grants_sdk/extensions/filters.py
SnowboardTechie: Nope, no significance, just emphasis. Reworded it to a plain "actual".

lib/python-sdk/common_grants_sdk/extensions/types.py
SnowboardTechie: On the string: it was a forward-ref, but redundant given `from __future__ import annotations` makes every annotation lazy anyway. Moot now, since the field is generic.

On `OppFilters` vs reusable: went reusable. `ClassifyResult` is now `Generic[T]` (matching `TransformResult` right below it), so the container isn't tied to opportunities. `classify_filters` returns `ClassifyResult[OppFilters]` since `opportunities.search` is the only filter-capable route today, but the container's ready to extend when the next one lands.

lib/python-sdk/tests/client/test_client.py
SnowboardTechie: Correct, and intentional. `Client.search` is the generic transport, so its tests only prove a populated `filterInfo` survives response validation, not that specific filter content is handled. The populated-`filterInfo` behavior (preservation + merging client errors ahead of server errors) lives in `opportunities.search`, and that's where the tests are: `test_opportunities.py` covers the populated `filterInfo.errors` cases, including merge order. Happy to add a populated case at the `Client` layer too if you'd rather pin it there, just didn't want to duplicate the opportunities coverage.

lib/ts-sdk/src/client/opportunities.ts
SnowboardTechie: Renamed it to `CustomFilterNames` (your second option). Lines up with the rest of the feature's vocabulary (`CustomFilterSpec`, `classifyFilters`, `SUPPORTED_CUSTOM_FILTER_ROUTES`), and since these types are module-scoped to opportunities the "Search" qualifier didn't feel necessary.

lib/ts-sdk/src/client/opportunities.ts
SnowboardTechie: Yeah, worth keeping on the radar. The wrinkle: fail-soft drops the filter client-side but still sends the request, so an invalid filter means a broader result set, exactly the cost you're describing. A consumer can inspect `filterInfo.errors` after the fact, but that doesn't save the call they already paid for. A configurable strict mode that throws before sending is really the only thing that prevents the wasted round-trip. Out of scope for the spike, but a real axis to revisit.

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: Good point on the implicit knowledge. I didn't fold `validateRoutes` into `validateFilterCall` (they're deliberately split: `validateRoutes` throws at registration, `validateFilterCall` is fail-soft per-value). Instead I took `validateFilterCall` off the public surface, so consumers now only see `validateRoutes` (registration) and `classifyFilters` (call-time, which runs `validateFilterCall` internally). Validation on the search path is unchanged: `validateRoutes` at construction, per-filter validation inside `classifyFilters`. The "have to know about this call" goes away because there's no longer a public call to know about.

lib/python-sdk/common_grants_sdk/extensions/types.py
jcrichlake: I'm a little confused, do we still need that as a string or no?

lib/ts-sdk/src/client/opportunities.ts
jcrichlake: I still see this as SearchFilterNames

lib/python-sdk/common_grants_sdk/extensions/types.py
SnowboardTechie: No string anymore. The field is `result: T` now (`ClassifyResult` is `Generic[T]`), so there's nothing to quote. The `"OppFilters"` was only a forward-ref to defer that name, and going generic removed the name entirely. Same shape as `TransformResult` right below it.

lib/ts-sdk/src/client/opportunities.ts
SnowboardTechie: That snippet's outdated. It's `CustomFilterNames` at the current head (`opportunities.ts:79`), and `SearchFilterNames` isn't anywhere in the tree anymore. GitHub's still showing the old hunk because the comment anchor went stale after the rename landed.

SnowboardTechie: Screenshot 2026-06-25 at 13 54 13

jcrichlake (approved):

PR #932, opened 2026-06-18
chore(deps): bump typescript to v6

SnowboardTechie:
### Summary

- Bumps the catalog `typescript` from `^5.9.3` to `^6.0.3` (latest 6.x).
- Co-bumps `typescript-eslint` `^8.61.0` → `^8.61.1`. Its `typescript` peer range is `>=4.8.4 <6.1.0`, so it supports TS 6.0.x — that compatibility is what gates this bump, and it passes.
- Handles two TS 6 breaking changes (deprecated `baseUrl`, stricter `exports` resolution) so the whole workspace stays green.
- Time to review: 5 minutes

### Changes proposed

- `pnpm-workspace.yaml`: `typescript` → `^6.0.3`, `typescript-eslint` → `^8.61.1`.
- `lib/ts-sdk/tsconfig.json` and `website/tsconfig.json`: removed the deprecated `baseUrl`.
- `lib/ts-sdk/tsconfig.json`: added a `json-schema-faker` `paths` entry pointing at its bundled declarations.
- `pnpm-lock.yaml`: regenerated.

### Context for reviewers

**Breaking changes handled**

1. **`baseUrl` deprecated (TS5101).** TS 6 errors on `baseUrl` and removes it as a module-resolution lookup root. Both `ts-sdk` and `website` used it only as a prefix for `paths`, so per the [TS 6 migration guidance](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html) it's removed and the `paths` values are made explicitly relative (`./src/*` etc.) — resolution is unchanged.

2. **Stricter `exports` resolution (TS2307).** `json-schema-faker@0.6.x` (a test-only devDependency) ships an `exports` map with only an `import` condition — no `require`/`default`. `ts-sdk` compiles in CommonJS mode, so under TS 6's stricter resolution `tsc` enters the conditional exports, finds no matching condition, and no longer falls back to the top-level `types` field (TS 5 did). `--traceResolution` confirmed `Saw non-matching condition 'import' … Module name 'json-schema-faker' was not resolved`. The package's bundled `.d.ts` is valid and runtime resolution (vitest/esbuild) is unaffected, so a `paths` entry points types at the bundled declarations. Downgrading to the last 0.5.x (which has a correct `exports` map) was rejected — 0.6.x changed `generate` to async, so it's an API change unrelated to this bump.

**Verification**

- `pnpm run ci` (core → changelog-emitter → cli → sdk → website) passes end-to-end, exit 0. Tests: 7 + 18 + 22 + 7 across the suites; lint, prettier, and CSpell clean.
- **No changeset.** `typescript`/`typescript-eslint` are build/dev tools, and the emitted public `.d.ts` of the published packages is byte-identical between 5.9.3 and 6.0.3 (verified by diffing declaration-only output of `ts-sdk` — 56 files, zero differences). No consumer-visible surface change.

### Additional information

`templates/express-js` uses `moduleResolution: "node"` (the node10 resolver, not the removed `classic`) and isn't in the pnpm workspace, so it's unaffected. No `outFile` or `target: es5` usage anywhere.

### CI note: undici override (dropped on rebase)

An earlier revision carried a `pnpm-workspace.yaml` override pinning `undici` to `7.28.0` to clear a high advisory ([GHSA-vmh5-mc38-953g](https://github.com/advisories/GHSA-vmh5-mc38-953g), TLS cert validation bypass) that was failing `validate` on `main` and all sibling catalog PRs. That override has been dropped: `main` now resolves `undici@7.28.0` transitively on its own, so it's no longer needed. Verified after rebasing onto current `main` — `audit-deps.js --level high` reports no high-severity advisories, with `undici` resolving to `7.28.0`.


karinamzalez (approved):
Still getting familiar with this repo, so mostly questions from me here-- the write-up was really helpful, especially the --traceResolution breakdown!

- On the version range: is ^6.0.3 the convention here, or would a tighter pin be preferred? I noticed typescript-eslint only officially supports TypeScript below 6.1, so I was curious whether the caret (which allows any 6.x) is intentional.
- Naive question: Is "no changeset for build/dev-tool bumps" a standard rule in this project, or judged case-by-case?

SnowboardTechie:
Yeah, the caret is just house style, every entry in the catalog is one. So `^6.0.3` is the pattern rather than a call about 6.x specifically. You're right it reaches past typescript-eslint's `<6.1.0` peer range on paper, but the lockfile pins `typescript@6.0.3` and CI installs with `--frozen-lockfile`, so 6.1 can't land without its own PR and a full CI run.

On the changeset, it's not written down and nothing enforces it, but the history is consistent: changeset if a consumer of a published package can see it. TypeSpec got one in #914 as a peer dep of core, js-yaml got one in #877 as a cli runtime dep. eslint v10 and `@types/node` v22 got none. typescript is in that group, build-only.

PR #933, opened 2026-06-18
chore(deps): bump zod to v4

SnowboardTechie:
### Summary

- Bumps the catalog `zod` range `^3.25.76` → `^4.4.3` (latest 4.x), and bumps the independently-pinned `templates/express-js` to match.
- `zod` is a **runtime dependency of the published `@common-grants/sdk` and `@common-grants/cli`**, so this is a release-warranting change — a changeset is included (per ADR-0010 + #821 precedent).
- Adapts everything v4 breaks, and replaces everything v4 deprecates, so nothing is left to rediscover when a v5 removes it.
- `pnpm run ci` is green across every package; the express-js template's own CI steps pass locally.
- No public API changes — the SDK/CLI surface is unchanged.
- Time to review: ~15 minutes

### Changes proposed

`zod: ^3.25.76` → `^4.4.3` in the `catalog:` block of `pnpm-workspace.yaml`, lockfile regenerated, plus the v4 source/test adaptations below.

**Breaking changes handled**

- **`z.record()` single-arg dropped** → two-arg `z.record(z.string(), …)`: `models.ts` (`otherDates`, `customFields`, `customFilters`), `with-custom-fields.ts` (`DEFAULT_VALUE_SCHEMAS.object` + the error-message example), two specs, and three sites in the express-js template.
- **`ZodError.errors` renamed** → `.issues`: the express-js template's validation middleware and service (5 sites).
- **`ZodType` generic** dropped its middle `Def` param: `z.ZodType<Out, Def, In>` → `z.ZodType<Out, In>` (client `OppSchema`).
- **`ZodObject` generics** collapsed (`UnknownKeysParam` / catchall removed): `HasCustomFields` → `z.ZodObject<Shape>`.
- **`AnyZodObject` removed** → `z.ZodObject` (`extensible-schema-coverage` test).
- **`SafeParseReturnType` renamed** → `ZodSafeParseResult` (client).
- **CLI validation message wording** changed `Invalid url` → `Invalid URL` (`error-handling` test assertion).

**Deprecated forms replaced**

v4 keeps these working but marks them for removal, so they're swept here rather than left for a future v5:

| Form | Replacement |
|---|---|
| `z.ZodTypeAny`, `z.ZodSchema` | `z.ZodType` (v4 exports both as deprecated aliases of `ZodType`) |
| `z.string().uuid()` / `.url()` / `.email()` | `z.uuid()` / `z.url()` / `z.email()` |
| `z.string().date()` / `.time()` / `.datetime()` | `z.iso.date()` / `z.iso.time()` / `z.iso.datetime()` |
| `.passthrough()` | `.loose()` |
| `.merge(B)` | `.extend(B.shape)` |

Covers SDK source, tests, examples, the CLI, the TypeScript guide on the docs site, and the express-js template. Found with `@typescript-eslint/no-deprecated` under type-aware linting rather than by grep, and worked down to zero remaining zod deprecations in `lib/ts-sdk` and `lib/cli`.

Each replacement was checked for equivalence before swapping: same accept/reject behavior, same parsed output, and same error codes and messages. One is worth a reviewer's attention: `.merge()` takes the **incoming** schema's unknown-key policy while `.extend()` keeps the **receiver's**, so they are not interchangeable in general. At both call sites here the objects are plain, so the swap is a no-op.

**Deliberately unchanged**

- **`.strict()`** is not deprecated in v4. It carries only a "consider `z.strictObject(A.shape)`" note, with no `@deprecated` tag, so those four sites are untouched.
- **ADR-0022's** pseudo-code still shows `z.ZodTypeAny`. It is a dated decision record and the alias still exists.
- **The SDK's own `statuses` deprecation** in `opportunities.ts` is unrelated to zod and has its own removal plan.

**Changeset**: `.changeset/zod-v4.md` — `@common-grants/sdk` and `@common-grants/cli`, the two published packages that depend on `zod` at runtime via `dependencies`. The `website` also consumes `zod` but is private and not changeset-tracked.

### Context for reviewers

**Local verification** (documented nix dev env, Node 22):

- `pnpm run ci` → green across **core, changelog-emitter, cli (119), sdk (548), website (69)**. No TS errors, prettier clean, cspell clean.
- **express-js template**, run the way its CI job does: `pnpm install --frozen-lockfile --ignore-workspace`, `pnpm run checks`, `pnpm run build`, all green. Smoke-tested at runtime as well: all 10 built-in mock opportunities parse against the migrated schemas, and bad `source` / `id` values are still rejected (`Invalid URL`, `Invalid UUID`).
- `pnpm audit --audit-level high`: the one `high` (`undici` via `website > wrangler > miniflare`) is **pre-existing on `main`** — `undici@7.24.8` is byte-identical between `main` and this branch. This bump introduces no new advisories.

**Downstream impact.** `zod` is a direct runtime dependency and the SDK's exported schemas are zod types, so this reaches consumers:

- A consumer on zod 3 passing their own schema into an SDK generic gets `TS2740` (`Type 'ZodString' is missing the following properties from type 'ZodType<...>'`).
- A consumer pinning zod 3 via an override cannot compile the SDK's `.d.ts` at all (`ZodPreprocess`, `ZodPipe`, `ZodTransform`, and `z.core` do not exist in zod 3).

`common-grants/ts-cg-grants-gov` is the second case today: it depends on `zod ^3.25.76` and sets `pnpm.overrides.zod: 3.25.76`. It needs a matching bump alongside this.

**Why the template is in scope.** `templates/express-js` pins `zod` directly rather than through the catalog, since it is a standalone package outside the workspace. It is not excluded from CI — `ci-template-express-js.yml` is path-gated on `templates/express-js/**` and runs install, checks, build, and audit. The repo's "templates are manually maintained" convention means they are excluded from Dependabot, so a deliberate refresh like this one is the intended path. Leaving it would ship a starter template teaching the v3 API against a v4 SDK.

`templates/quickstart` and `templates/fast-api` have no zod dependency and are untouched.


karinamzalez (approved):
Nothing blocking from me, I just have a few questions: 
- z.ZodTypeAny was updated to z.ZodType in responses.ts but left as-is in base.ts and elsewhere. Was that intentional (only changing it where the client's type inference needed it) or is it worth making them all consistent?
- Deprecated v4 forms left in place (.uuid(), .url(), .datetime(), .strict(), .passthrough(), .merge()).. is there a tracking issue for modernizing them before a future zod major removes them?
- Still getting familiar with Changeset! We marked `patch` for a major-version dependency bump. Re: the convention-- is `patch` standard for dependency bumps with no public API change, and what would push one of these to `minor` instead?

SnowboardTechie:
Good thing you asked, `ZodTypeAny` isn't a separate type in v4. Zod exports it as a deprecated alias of `ZodType`, so that swap was cosmetic, not an inference fix. I checked by putting it back on all five helpers. The real v4 fixes were in `client.ts` and `builder.ts`. `.strict()` isn't deprecated either, it only carries a "consider `z.strictObject()`" note, so those stay. Description is corrected on both.

No tracking issue, I swept the rest into this PR instead. That pulled in the express-js template, which needed real v4 fixes of its own. Details in the description.

You're right to flag the changeset. Patch is our standard for dep bumps, and what pushes one up is the bump reaching consumers, which this one does: on zod 3, handing us your own schema gets `TS2740`, and a zod 3 override can't compile our `.d.ts` at all. Moved the SDK to minor, CLI stays patch since it's an executable with nothing importable. `ts-cg-grants-gov` pins zod 3 today so it needs a matching bump.

Issue #10, opened 2024-12-26
ADR template and landing page


widal001:
### Summary

Create a landing page and template for the ADRs and update the instructions in the ADR issue template to reflect this.

### Acceptance criteria

- [ ] ADR template contains the required structure for an ADR and is ignored in the published docs
- [ ] ADR landing page lets users browse ADRs by topic, etc.
- [ ] ADR issue template has been updated to reference this new template and landing page

PR #934, opened 2026-06-18
feat(py-sdk): codegen-free typed custom-filter authoring DX

### LINKED ISSUE #938: [Py SDK] Typed custom-filter authoring (codegen-free)
### Summary

The custom-filter PoCs declare filters as runtime `CustomFilterSpec` dicts with no compile-time typing: an author gets no static signal that a declared filter's value shape lines up with what a consumer passes. #924 shifted the Python SDK to a codegen-free plugin framework (Pydantic generics and subclasses); the same approach can give custom filters a typed authoring experience with no build step.

This issue covers the static-typing layer over the existing runtime classifier, which stays unchanged. An author extends a standard-filters `TypedDict` with their custom filters and registers it on a route; a consumer gets each filter key narrowed to its declared value model, checked statically with no call-site annotations.

Drafted in #934: phantom `RouteFilters[TF]` carriers, an `OpportunityFilters` `TypedDict` base, `define_plugin(routes=...)` threading the typed routes onto the returned `Plugin`, and `Plugin.search_filters_type()` recovering the registered filter type via a constrained-`self` projection.

### Acceptance criteria

- [x] An author can extend a standard-filters `TypedDict` with custom filters and register it per route, with no build step
- [x] A consumer's filter keys narrow to their declared value models, checked statically (verified by `assert_type` in a runnable example)
- [x] The static-typing layer is purely additive: no behavior change to the existing throw-based runtime classifier (`classify_filters`, `validate_routes`, `f`, `FilterError`)
- [x] Authoring DX follows the codegen-free framework direction established in #924

SnowboardTechie:
### Summary

Makes custom filters typed and usable end-to-end in the Python SDK: an author registers the filters a route accepts, a consumer authors a filter dict the type checker narrows per key, and `client.opportunities.search(filters=...)` sends those filters to a real endpoint. Builds on the codegen-free plugin framework from #924.

- **What:** extend a standard-filters `TypedDict` (`OpportunityFilters`) with your custom filters and register them on the plugin (`define_plugin(routes=...)`); a consumer's filter keys narrow to their declared value models; `search()` classifies the filter dict into the request body and POSTs it.
- **Why:** custom filters were runtime-only `CustomFilterSpec` dicts with no compile-time typing, and the SDK client had no way to send them — so the typed authoring layer had no endpoint to flow to.
- **Closes #938** (sub-issue of #645); builds on the #924 framework.
- Time to review: ~15 minutes.

### Changes proposed

The typed authoring surface is plain `TypedDict`s; the runtime classifier (`classify_filters`, `validate_routes`, `FilterError`) and the route-keyed `PluginRoutes` shape (ADR-0022 Decision #10) are the contract.

- `schemas/pydantic/filters/opportunity.py`: `OpportunityFilters` — a `TypedDict` of the standard opportunity-search filters that an author extends to give each custom filter its own typed key. Clean value aliases (`StringArray`, `NumberComparison`, `DateComparison`, …) read at the call site. Range fields keep the explicit `*RangeFilter` names (the clean `*Range` names already export as the range value sub-models).
- `extensions/plugin.py`: `Plugin.routes` is a route-keyed `PluginRoutes` map (`routes[resource][method].filters[name] → CustomFilterSpec`, ADR-0022 Decision #10); `define_plugin(routes=...)` threads it onto the plugin, defaulting to an empty map.
- `client/opportunities.py`: `Opportunities.search()` gains optional `filters` and `routes`. The consumer's filter dict is classified via `classify_filters` into the three-bucket body (default named fields + a `customFilters` record) and POSTed; `status` becomes an optional shorthand merged into the filter set (`mode="json"` on the wire dump so date filters serialize to ISO strings).
- `extensions/filters.py`: the `f.*` builders return the precise filter model per value type (`f.in_([...]) → StringArrayFilter`, `f.gte(3) → NumberComparisonFilter`), so a dict built with `f.*` composes directly with the typed filter bag. Wire output unchanged.
- `examples/consumer_search_with_filters.py` (runnable): the end-to-end flow — register filters on the plugin, compose with `f.*`, search a live endpoint, get typed `OpportunityBase` rows back.
- `extensions/__init__.py`, `schemas/pydantic/filters/__init__.py`: export the aliases and filter types.
- Tests: `PluginRoutes` threading on `define_plugin`; client `search(filters=, routes=)` routing custom filters into `customFilters` and date filters serializing JSON-safe; `f.*` precise-model wire shapes.

### Context for reviewers

Run it against a CommonGrants endpoint on `http://localhost:8000` (see `lib/python-sdk/examples/README.md`). The Pennsylvania reference API accepts the `customFilters` body:
```
cd examples/pa-opportunity-example && make install && make dev   # one terminal
cd lib/python-sdk && poetry run python examples/consumer_search_with_filters.py   # another
```

```python
class OppSearchFilters(OpportunityFilters, total=False):
    agency: StringArray
    awardCount: NumberComparison

plugin = define_plugin(
    PluginSchemas(Opportunity=schema(common_schema=OpportunityBase)),
    routes={"opportunities": {"search": {"filters": {
        "agency": CustomFilterSpec(filter_type=CustomFilterType.STRING_ARRAY),
    }}}},
    meta=PluginMeta(name="grants.gov", source_system="grants.gov"),
)

filters: OppSearchFilters = {"status": f.in_(["open"]), "agency": f.in_(["NSF"])}
client.opportunities.search(search="...", filters=filters, routes=plugin.routes)
```
A consumer annotates the filter dict with their `OppSearchFilters` and gets per-key narrowing (`agency` must be a `StringArray`); the wrong model on a key is a static error under both mypy and pyright.

Verified: `make checks` green (black, ruff, mypy clean on 84 files), pyright clean on the changed files, 378 tests pass.

### Additional information

Author-facing keys are the camelCase wire names (`closeDateRange`, `minAwardAmountRange`); the runtime classifier also accepts snake_case and normalizes it, so a Python author can write either.

Custom-filter *semantics* are the deployment's concern (ADR-0022 — serving is out of scope): the SDK emits the correct request body; how a server applies custom filters is its own choice. The PA reference API, for instance, accepts and echoes `customFilters` in `filterInfo` while narrowing results on the standard `status` filter.


lib/python-sdk/common_grants_sdk/extensions/plugin.py
```diff
@@ -55,23 +74,56 @@ class PluginSchemas(Generic[_TOpportunity]):
...
 @dataclass(frozen=True)
-class Plugin(Generic[SchemasT]):
+class Plugin(Generic[SchemasT, _RoutesT]):
     """The plugin singleton consumers import.
 
-    ``schemas`` is a typed frozen dataclass, so ``plugin.schemas.Opportunity`` is
-    fully typed (dot access).
+    ``schemas`` and ``routes`` are typed frozen dataclasses, so
+    ``plugin.schemas.Opportunity`` is fully typed (dot access) and
+    ``plugin.search_filters_type()`` recovers the registered search-filters
```
jcrichlake: This misaligns with the ADR, is this an intentional variation?

lib/python-sdk/common_grants_sdk/extensions/routes.py
```diff
@@ -0,0 +1,65 @@
...
+
+@dataclass(frozen=True)
+class ResourceRoutes(Generic[_S]):
+    """A resource's filterable methods. One slot per method (``search`` to start)."""
+
+    # The slot is covariant so a plugin holding a concrete RouteFilters is usable
+    # where the base is expected — what the constrained-``self`` projection relies
+    # on. Covariance is sound here (frozen, read-only). mypy forbids a covariant
+    # TypeVar as a parameter (dataclass fields become ``__init__`` params); pyright
+    # accepts it. The ignore keeps both checkers green.
```
jcrichlake: I'm confused by what this is trying to tell me. Let's clarify it so that it's in plainer language.

lib/python-sdk/common_grants_sdk/schemas/pydantic/filters/opportunity.py
```diff
@@ -46,3 +63,25 @@ class OppFilters(OppDefaultFilters):
         description="Additional custom filters to apply to the search",
         alias="customFilters",
     )
+
+
+class OpportunityFilters(TypedDict, total=False):
+    """Standard filters for the opportunities search route (typed authoring DX).
+
```
jcrichlake, line 71: Either the class should contain the word Search or the doc string should omit search.

lib/python-sdk/examples/consumer_search_with_filters.py
jcrichlake, line 1: This file should probably call one of the existing `simpler-grants-gov` endpoints.  Most likely one of the common grants ones.  There's several already existing examples you can follow.  If you want some help setting up one of those endpoints locally let me know.

jcrichlake (changes requested):
Nice work! Thanks for putting this in place.  Left a couple comments where hopefully things can get tightened up a little bit.

lib/python-sdk/common_grants_sdk/extensions/plugin.py
SnowboardTechie: Thank you for catching this, it went deeper than the doc line. No, not an intentional variation: chasing the flag, `search_filters_type()` and the phantom `Routes`/`ResourceRoutes`/`RouteFilters` carriers behind it turned out not to work. The annotation pattern the docstring promised (`FiltersT = plugin.search_filters_type(); filters: FiltersT = ...`) fails both mypy and pyright, since Python won't take a runtime value as a type expression, so they bought no working DX the plain TypedDict doesn't already give.

Pulled all of it. Typed authoring now rides on the `OpportunityFilters` TypedDict (extend it, annotate your filter dict with the subclass), and `routes` is the route-keyed `PluginRoutes` map on `Plugin.routes` per ADR-0022 Decision #10.

lib/python-sdk/common_grants_sdk/extensions/routes.py
SnowboardTechie: That whole file came out. It was the phantom-carrier layer behind `search_filters_type()` (more in the `plugin.py` thread), so the confusing comment went with it.

lib/python-sdk/common_grants_sdk/schemas/pydantic/filters/opportunity.py
SnowboardTechie: Went with the docstring. Rewrote the first line to name what it is (the typed authoring surface you extend) rather than calling it the "search route", and kept the `OpportunityFilters` name since the sibling `OppDefaultFilters` follows the same no-Search convention.

lib/python-sdk/examples/consumer_search_with_filters.py
SnowboardTechie: Done. It uses the real `Client` against a CommonGrants endpoint on localhost:8000 now, like the other examples, instead of the stub. That meant teaching the client to actually send custom filters, which it couldn't before, so `Opportunities.search()` now takes `filters=`/`routes=` and builds the body through `classify_filters`.

jcrichlake (approved):

PR #940, opened 2026-06-23
[Issue #900] transforms example

### LINKED ISSUE #900: [TS SDK] Build example of transforms from SGG
### Summary

After expanding the TypeScript plugin to support transforms, build an example that uses the plugin to transform Simpler.Grants.gov opportunity data to and from CommonGrants.

### Acceptance criteria

- [ ] A working example demonstrates SGG opportunity data transformed via the `common-grants` TypeScript plugin
- [ ] The example is documented in the plugin repo

jcrichlake:
### Summary

  - Fixes #900
  - Time to review: 10 minutes

  ### Changes proposed
  > What was added, updated, or removed in this PR.

  Adds end-to-end round-trip transform example scripts for both the TypeScript and Python SDKs using the published `cg-grants-gov` plugin (v0.2.0). These examples replace the previous
  approach of referencing a local plugin with a dependency on the published package.

  **New files:**
  - `lib/ts-sdk/examples/grants-gov-transforms.ts` — TypeScript round-trip validation script
  - `lib/python-sdk/examples/grants_gov_transforms.py` — Python round-trip validation script

  **Updated:**
  - `lib/ts-sdk/package.json` — Added `@common-grants/cg-grants-gov@^0.2.0` dev dependency and `example:grants-gov` npm script
  - `lib/ts-sdk/tsconfig.json` — Added `@common-grants/sdk/*` path alias pointing to `./src/*`
  - `lib/python-sdk/pyproject.toml` — Added `cg-grants-gov = "^0.2.0"` dependency
  - `lib/python-sdk/poetry.lock` / `pnpm-lock.yaml` — Updated lockfiles
  - Minor: removed a blank line in `transformation.py`, added explicit `-> None` return types on two test functions

  ### Context for reviewers
  > Testing instructions, background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers. Explain how the changes were
  verified.

  Both example scripts perform the same end-to-end validation flow:

  1. Fetch a real opportunity from a running Simpler.Grants.gov API instance (`http://localhost:8080`) by opportunity ID
  2. Validate the raw response against `GrantsGovOpportunitySchema`
  3. Transform to CommonGrants format via `toCommon` / `to_common`
  4. Transform back to source format via `fromCommon` / `from_common`
  5. Assert that key fields survive the round-trip (exact matches: `opportunity_id`, `opportunity_number`, `agency_code`, `agency_name`, `opportunity_title`; semantic match:
  `opportunity_status` with known lossy mappings like `archived → closed`; presence checks: `created_at`, `updated_at`)
  6. Run a full deep-diff of raw vs. round-tripped output and report any remaining field differences

  The key change from prior iterations is that these examples now depend on the **published** `cg-grants-gov` / `@common-grants/cg-grants-gov` package (v0.2.0) rather than a local
  plugin path, making them usable as standalone reference implementations.

  **Running the TypeScript example:**
  ```bash
  # Requires a local Simpler.Grants.gov API running on port 8080
  pnpm --filter @common-grants/ts-sdk example:grants-gov <opportunityId>
  # or
  GRANTS_GOV_OPP_ID=<id> pnpm --filter @common-grants/ts-sdk example:grants-gov
  ```

  **Running the Python example:**
  ```bash
  cd lib/python-sdk
  poetry run python examples/grants_gov_transforms.py <opportunityId>
  # or
  GRANTS_GOV_OPP_ID=<id> poetry run python examples/grants_gov_transforms.py
  ```

  ### Additional information
  Python Screenshots
  
image

image

image

image

image

image

image

  
image

image

image


lib/python-sdk/pyproject.toml
```diff
@@ -11,6 +11,7 @@ python = "^3.11"
 pydantic = "^2.11.0"
 marshmallow = "^3.20.0"
 httpx = ">=0.27,<0.29"
+cg-grants-gov = "^0.2.0"
```
SnowboardTechie: The bigger thing I'd push on: this puts `cg-grants-gov` in the main runtime deps, but `cg-grants-gov` 0.2.0 itself requires `common-grants-sdk` (>=0.7.0), so the SDK now points at its own downstream plugin and pulls it plus transitives into every `pip install common-grants-sdk`, even though nothing in the shipped package imports it. The TS side has this under `devDependencies`. Worth moving to `[tool.poetry.group.dev.dependencies]` to match.

lib/python-sdk/examples/grants_gov_transforms.py
```diff
@@ -0,0 +1,244 @@
...
+
+from common_grants_sdk.client import Client
+from common_grants_sdk.client.config import Config
+
+# =============================================================================
+# Config
+# =============================================================================
+
+opp_id = sys.argv[1] if len(sys.argv) > 1 else os.environ.get("GRANTS_GOV_OPP_ID")
+api_key = "two_org_user_key"
```
SnowboardTechie, line 31: Small one: this `api_key` is dead, `Config(...)` re-inlines the same literal right below. TS uses its `apiKey` const, so either drop this line or pass it into `Config` for parity.

lib/python-sdk/examples/grants_gov_transforms.py
```diff
@@ -0,0 +1,244 @@
...
+    if diffs:
+        print(f"  {len(diffs)} field(s) differ after round-trip:")
+        for path, a_val, b_val in diffs:
+            print(f"    {path}")
+            print(f"      raw:  {a_val!r}")
+            print(f"      back: {b_val!r}")
+    else:
+        print("  PASS  all fields identical after round-trip")
+
+    print("\n=== All checks passed ===")
```
SnowboardTechie: Minor: the `deepDiff` above prints any field differences but never fails, so "All checks passed" prints even when it just listed some. The round-trip is intentionally lossy so the diff being informational is fine, just the final line reads a touch contradictory next to it. Maybe "explicit checks passed", or fold the diff count in. (Same on the TS side at `grants-gov-transforms.ts:251`.)

SnowboardTechie (commented):
Nice work on this. The round-trip plus deep-diff approach reads really clean, and having both SDKs exercised against a live opportunity is exactly the consumer-side proof these transforms needed. Left a couple of inline notes; the Python dependency-placement one is the only thing I'd want sorted before merge, the rest are small.

One that doesn't map to a diff line, more a question: [#900](https://github.com/HHS/simpler-grants-protocol/issues/900)'s second AC is "the example is documented in the plugin repo," and `cg-grants-gov` ships from the separate `ts-cg-grants-gov` / `py-cg-grants-gov` repos rather than this one. Is that doc step meant to land there in a follow-up, or are we treating the example plus run command here as enough to close it? Either way, the new example isn't in `lib/ts-sdk/examples/README.md` yet (which has a per-example section for the others), so adding it there seems worth doing regardless.

SnowboardTechie (approved):
Dependency move's clean, nits handled on both SDKs, CI green. Approving.

Issue #967, opened 2026-06-30
[TS SDK] Prepare SDK for custom filters release


widal001:
### Summary

Close a few remaining gaps in the plugin author's experience defining custom filters, and the plugin consumer's experience using a plugin-specific `Client` to send API requests with custom filters and parse API responses that include custom fields. 

This is the TypeScript counterpart of #966 and the gaps overlap with the Python SDK but are not directly 1:1.

#### Challenges for plugin authors

- `definePlugin()` doesn't provide intellisense for which routes/methods support custom filters, since `PluginRoutes` is `Record<string, Record<string, RouteDeclarations>>` (open string keys) instead of a closed, typed shape.
  ```ts
  const plugin = definePlugin({
    routes: {
      oportunites: {  // misspelled but no error
        search: { filters: { agency: { filterType: "stringArray" } } },
      },
    },
  });
  ```
- This also means `definePlugin()` doesn't flag route/method typos as compile errors. Those issues only get surfaced when a plugin consumer passes the routes into `new Client()`.

#### Challenges for plugin consumers

- Plugin consumers need to pass `plugin.routes` into `new Client()` to classify their custom filters, rather than getting a `Client` already scoped to the plugin via `plugin.getClient()`.
  ```ts
  // current
  const client = new Client({ baseUrl: "...", routes: plugin.routes });

  // desired
  const client = plugin.getClient({ baseUrl: "..." });
  ```
- `client.opportunities.search({ filters })` doesn't fail fast on a bad filter **value**. Ad-hoc (non-registered) filters should still pass through, but a malformed value, or a wrong-typed value on a standard or pre-registered custom filter, is silently dropped into the response's `filterInfo.errors` array (which is meant for errors returned by the API server) instead of being a compile error.
  ```ts
  // no compile error, no throw: the bad value is dropped into filterInfo.errors
  await client.opportunities.search({ filters: { agency: { operator: "bogus", value: 1 } } });
  ```
- `client.opportunities` methods don't use the plugin's schema to parse responses by default. Plugin consumers have to pass `schema` on every call, so custom fields don't get validated or typed unless the consumer remembers.
  ```ts
  const r = await client.opportunities.search({ filters });  // customFields untyped without `schema`
  ```
- A parsing error on a single row throws for the entire response (`rawItems.map(i => schema.parse(i))`), rather than collecting successful rows and errors.
  ```ts
  const result = await client.opportunities.search({ filters: { agency: F.in(["NSF"]) } });

[...]

PR #970, opened 2026-07-01
feat(py-sdk): typed custom-filters authoring surface and client (#966)

### LINKED ISSUE #966: [Py SDK] Prepare SDK for custom filters minor release
### Summary

Close a few remaining gaps in the plugin author's experience defining custom filters, and the plugin consumer's experience using a plugin-specific `Client` to send API requests with custom filters and parse API responses that includes custom fields.

#### Challenges for plugin authors

- `define_plugin()` doesn't provide intellisense support for the set routes/methods that support custom filters, since `PluginRoutes` is just a dictionary with string keys instead of a dataclass or TypedDict
  ```python
  plugin = define_plugin(
    routes={
      "oportunites": {  # misspelled but no error
        "search": {
          "filters": { "agency": CustomFilterSpec(filter_type=CustomFilterType.STRING_ARRAY) }
        }
      }
    }
  )
  ```
- This also means that `define_plugin()` doesn't flag typos in `routes=` either as IDE warnings or as runtime errors. Those issues are only caught when plugin consumers pass them into `Client()`
- The DX for defining custom fields and custom filters are a bit inconsistent (`PluginRoutes` is an open dictionary with string keys, `PluginSchemas` is a dataclass with fixed attributes)

#### Challenges for plugin consumers

- Plugin consumers need to pass `Plugin.routes` into `Client()` to classify their custom filters, rather than being able to access a `Client` already scoped with custom filters via `Plugin.get_client()`
  ```python
  # current
  client = Client(base_url="...", routes=Plugin.routes)
  
  # desired
  client = Plugin.get_client(base_url="...")
  ```
- If Plugin consumers want to get type hints for their filters they need to manually create their own `OpportunityFilters` sub-class like this, or import one that the plugin author creates. And because it's defined in two places there can be drift between this sub-class and the `PluginRoutes` input.
  ```python
  # need to define this in addition to passing `Plugin.routes` into `Client()`
  class OppSearchFilters(OpportunityFilters, total=False):
      agency: StringArray

  my_filters: OppSearchFilters = {"agency": f.in_(["NSF"]), "status": f.in_(["open"])}
  client.opportunities.search(search="ai", filters=my_filters)
  ```
- `client.opportunities.search(filters={...}`) doesn't provide IDE warnings or runtime errors for malformed filters and instead those errors are silently added to the `filters.errors` array in the response, which is meant to be reserved for errors returned by the API server.
- `client.opportunities` methods don't use `Plugin.schemas` to parse responses by default, meaning custom fields don't get validated or typed.
- Parsing errors for a single row raises an error for the entire response, rather than accumulating successful results and errors.
  ```python
  result = client.opportunities.search(filters={"agency": f.in_(["NSF"]), "status": f.in_(["open"])})

  # successfully parsed responses
  for opp in result.items:
       print(opp.custom_fields) # includes parsed custom fields

  # errors collected
  for err in result.errors:
     print(err.index)
     print(err.
  ```

### Suggested DX

Here are some examples of the plugin author and consumer experiences with these gaps fixed:

#### Plugin author experience

```python
from common_grants_sdk.extensions import (
    define_plugin, schema, PluginSchemas, PluginRoutes, ResourceRoutes, RouteFilters, PluginMeta, f,
)
from common_grants_sdk.extensions import CustomField
from common_grants_sdk.schemas.models import OpportunityBase
from common_grants_sdk.schemas.filters import OpportunityFilters, StringArray

# 1. CUSTOM FIELDS — declared once as a typed model. `CustomField[V]` is the single
#    source of truth: each field's value type is V, and that type flows through to
#    `opp.custom_fields.*` on every parsed response row
class OppCustomFields(BaseModel):
    program_code: CustomField[str]
    legacy_id: CustomField[int]

# 2. CUSTOM FILTERS — declared once by subclassing the SDK's base filter TypedDict.
#    This same type both registers the filter (via RouteFilters below) AND types the
#    consumer's `search(filters=...)` call site without a  separate CustomFilterSpec dict.
class OppCustomFilters(OpportunityFilters, total=False):
    region: StringArray

# 3. ASSEMBLE — bind the custom-field model to the Opportunity schema slot, and the
#    custom-filter type to the opportunities/search route. One call, two axes.
opportunity_plugin = define_plugin(
    schemas=PluginSchemas(Opportunity=schema(common_schema=OpportunityBase[OppCustomFields])),
    routes=PluginRoutes(opportunities=ResourceRoutes(search=RouteFilters[OppCustomFilters]())),
    meta=PluginMeta(name="grants-gov adapter", source_system="grants.gov"),
)
```

#### Plugin consumer experience

```python
client = opportunity_plugin.get_client(config)

result = client.opportunities.search(filters={"region": f.in_(["US-CA", "US-NY"])})
#                                              ^ "region" autocompletes; f.eq(...) here would be a
#                                                pyright error (region is a StringArray filter)

for opp in result.items:                 # result.items: list[OpportunityBase[OppCustomFields]]
    print(opp.custom_fields.program_code.value)   # typed `str`, from OppCustomFields

# result.errors collects any ParseFailure rows, preventing a single bad row from raising the entire response
```

### Notes

- I put a suggested [implementation plan](https://github.com/user-attachments/files/29510214/py-sdk-filters-client-plan.md) to address these gaps together based on the approach I took in [CommonBenefits](https://github.com/common-benefits/common-benefits/pull/13)
- There might be an opportunity to streamline the DX of `ResourceRoutes` and/or `RouteFilters` by using a `TypeDict`, but make sure that an alternative doesn't loosen IDE-level type hints or runtime errors.

### Acceptance criteria

- [x] Plugin authors
  - [x] Get intellisense support when defining custom filters for the routes and methods that support them
  - [x] Receive both an IDE warning and a runtime error if they misspell a route/method name or pass in routes that don't support custom filters when passing them into `define_plugin()`
  - [x] Provide IDE-level type hinting to plugin consumers without having to define custom filters twice and risking discrepancies
  - [x] Have a relatively consistent DX across defining custom filters and custom fields
- [x] Plugin consumers
  - [x] Can generate a `Client` instance that's already configured with a plugin's custom fields and custom filters using `Plugin.get_client()` (without having to pass in any configurations the plugin already has)
  - [x] Receive both IDE warnings and a runtime error if they provide an invalid input to a standard filter or a pre-registered custom filter
  - [x] Can get API responses parsed with pre-registered custom fields, without having to set the `schemas=` each time they invoke `client.opportunities.search()`
  - [x] Can access both the successfully parsed rows, and the rows with errors when a subset of rows in a search result contain malformed data
- [x] PR includes an example that tests the happy path of the features described above, and instructions for PR reviewers to validate that we correctly guard against the unhappy path(s)
- [x] Changes to the recommended approach and implementation plan are flagged and discussed in a meeting or Slack, and noted in  the PR.

SnowboardTechie:
### Summary

- Closes #966
- Time to review: ~20 minutes

Prepares the Python SDK (`common-grants-sdk`) for its custom-filters minor release by closing the plugin-author and plugin-consumer DX gaps:

- Plugin authors declare custom filters/fields once and get typed routes — a misspelled route/method is now a type error, not a silent runtime miss.
- Plugin consumers call `plugin.get_client(config)` and get a client already scoped with the plugin's filters and schemas: `search(filters=...)` is typed by the registered filters, and responses parse with the plugin's custom fields by default.
- Search/list responses partition successfully-parsed `items` from per-row parse `errors`, so one malformed row no longer fails the whole response.
- An invalid value on a standard or registered filter now raises `FilterError` before the request goes out — ad-hoc (unregistered) keys stay best-effort — and `filter_info.errors` is reserved for server-returned errors.

### Changes proposed

**Type gate**

- Switched the python-sdk type gate from mypy to pyright (basic mode) and wired pyright into `ci-lib-pysdk.yml` (added a Node setup step). **Tests are excluded from the gate** — flagging for team confirmation (basic mode was the decided call; the test scope is the team's).

**Authoring surface**

- `OpportunityFilters` is now an open `TypedDict` (PEP 728 `extra_items`, bound `CommonGrantsBaseModel`): standard keys typed to their value models, a `total=False` subclass gives each registered custom filter its own typed key, and unregistered keys still pass through. One source of truth for both registration and the consumer's `search(filters=...)` call site — replacing the two prior surfaces (`CustomFilterSpec` + an inert filter class).
- Typed route carriers: `PluginRoutes(opportunities=ResourceRoutes(search=OppSearchFilters))`, where the route slot holds the filter TypedDict class directly; `classify_filters` recovers the registered filters from it via `get_type_hints`.
- `Plugin.get_client()` projects the registered filter TypedDict and the plugin's Opportunity schema into a fully-typed client.

**Client**

- `BaseClient` (transport plumbing) + generic `Client[FiltersT, ItemT]` (typed facade). `Opportunities.search()` / `.list()` return `SearchResult` / `ListResult` with `{items, errors}` per-row partitioning (`parse_batch`), a plugin-bound default parse schema (no per-call `schema=` needed), and a per-call `schema=` override.

**Docs / versioning**

- Rewrote the existing changeset (`common-grants-sdk: minor`) to describe the new surface. Added `examples/typed_custom_filters.py` (in-gate `assert_type` proofs + a runnable runtime-raise demo) and `examples/typed_custom_filters_failures.py` (reviewer-run negative fixtures).

### Context for reviewers

**Verification.** `cd lib/python-sdk && make checks` (black, ruff, pyright basic → 0 errors) + `make test` (393 passed) + `make build`. The public package surface was also validated from a downstream consumer: the built wheel was installed into a fresh venv and exercised end-to-end (`define_plugin` → `get_client` → classify fail-soft → `parse_batch`); this caught and fixed a missing `parse_batch` export.

**Unhappy-path checks (please run):**

- Static: `cd lib/python-sdk && poetry run pyright examples/typed_custom_filters_failures.py` should report a route-typo error and a wrong-filter-value error. That file is excluded from the gate so its intentional errors don't fail CI.
- Runtime: `cd lib/python-sdk && poetry run python -c "import examples.typed_custom_filters as e; e.demo_invalid_filter_raises()"` should print `raised as expected: filters.region …` — the client raises before any request. (No API needed.)

**Cross-SDK.** The `classify_filters` primitive stays fail-soft in both SDKs (still collects, never raises); the raise is applied at the Python client boundary only. The TS client gets the matching client-side behavior in #967.

**Approach changes flagged** (per the issue's AC that recommendation changes be noted):

- **`RouteFilters` dropped**: the route slot holds the filter TypedDict class directly, per the issue's own note about streamlining `ResourceRoutes` / `RouteFilters` without loosening type hints or runtime errors.
- **Single-resource (Opportunity) scope**: resource-agnostic classification is deferred to #896.
- **Tests excluded from the pyright gate** (team confirm — see above).

### Additional information

Author + consumer DX (from `examples/typed_custom_filters.py`):

```python
class OppCustomFields(CustomFieldSet):
    program_code: Optional[CustomField[str]] = None
    legacy_id: Optional[CustomField[int]] = None

class OppSearchFilters(OpportunityFilters, total=False):
    region: StringArray

plugin = define_plugin(
    PluginSchemas(Opportunity=schema(common_schema=OpportunityBase[OppCustomFields])),
    routes=PluginRoutes(opportunities=ResourceRoutes(search=OppSearchFilters)),
    meta=PluginMeta(name="grants-gov adapter", source_system="grants.gov"),
)

client = plugin.get_client(config)
result = client.opportunities.search(filters={"region": f.in_(["US-CA", "US-NY"])})
#                                              ^ "region" autocompletes; f.eq(...) is a type error

for opp in result.items:            # typed with the plugin's custom fields
    print(opp.custom_fields.program_code.value)
for err in result.errors:           # per-row parse failures, collected not raised
    print(err.index, err.message)
```

Targets `HOLD-filters` (the custom-filters integration branch), not `main`.


jcrichlake (approved):

.changeset/python-typed-custom-filters.md
widal001, line 1: Just leaving a reminder that once we merge in the final set of PR changes we might want to look holistically at the changesets and make sure they:

1. Provide a relatively clean summary of what changed since the last release
2. Don't reference ephemeral design patterns that were introduced and then removed on this hold branch, since SDK consumers will only care about the final set of SDK changes that are shipped, not the intermediate steps.

lib/python-sdk/common_grants_sdk/client/opportunities.py
```diff
@@ -1,32 +1,40 @@
...
 
+# Bound is OpportunityBase[Any] (the custom-fields parameter is invariant, so a
+# bare OpportunityBase bound would reject OpportunityBase[OppCustomFields]).
+ItemT = te.TypeVar(
+    "ItemT", bound="OpportunityBase[Any]", default="OpportunityBase[Any]"
+)
+
+
+class Opportunities(Generic[FiltersT, ItemT]):
+    """Fetch opportunity data from the CommonGrants API.
```
widal001, line 29: While we don't need to make this change now, since we just have one resource, I'd encourage us to consider abstracting some of the common functionality to a resource base class that other resources can extend and wrap. 

We'll probably want to do this when we add `Awards` or `Organizations`, which could happen as soon as Sprint 6.8 or 6.9. That would also be a good opportunity to evaluate what lives on the `Client` class vs. what lives on the `Resource` base class. And I might recommend deferring #896 to this step because it might make more sense to make that a private helper (`Resource._classify_filters()`) instead of a public function, since SDK consumers should never have to call `classify_filters()` directly themselves.

lib/python-sdk/common_grants_sdk/client/opportunities.py
```diff
@@ -38,174 +46,171 @@ def path(self) -> str:
...
-
-        # Return the OpportunityBase from the response
-        return opportunity_response.data
+        return cast("ItemT", resolved.model_validate(success_response.data))
 
     def search(
         self,
-        search: str,
+        search: str = "",
         status: List[OppStatusOptions] | None = None,
```
widal001, line 111: Do we want to add a deprecated note in the docstrings for `status` so we can give people a heads up that we're removing it?

lib/python-sdk/examples/typed_custom_filters.py
```diff
@@ -0,0 +1,133 @@
...
+#    ``search(filters=...)`` call site — no second declaration to drift.
+class OppSearchFilters(OpportunityFilters, total=False):
+    region: StringArray
+
+
+# 3. ASSEMBLE — bind the custom-field model to the Opportunity schema slot and the
+#    custom-filter type to the opportunities/search route. One call, two axes.
+opportunity_plugin = define_plugin(
+    PluginSchemas(Opportunity=schema(common_schema=OpportunityBase[OppCustomFields])),
+    routes=PluginRoutes(opportunities=ResourceRoutes(search=OppSearchFilters)),
```
widal001, line 68: Nice work figuring out how to avoid the need for `RouteFilters` 🎉

lib/python-sdk/examples/consumer_search_with_filters.py
widal001, line 1: These examples were super helpful but this directory is getting a bit messy. After we get the custom filters stuff released, I'd like us to go back and clean up the examples, standardizing the naming and grouping related examples under a nested sub-directory if they're all testing the same thing, e.g. `examples/custom_filters/` or maybe even just `examples/extensions/` or `examples/plugins/`

widal001 (commented):
I left a few forward-looking comments re: abstracting some of the `Opportunities` code to a `Resource` base class that could also support future resources like `Awards` or `Organizations`. And personally I'd recommend deferring #896 until we get to that step, since we might solve the problem differently if/when we create a base `Resource` class.

I also added a few small reminders for us to:

- add a deprecation note to the status option for `search()`
- review the changesets on the hold branch to ensure they make sense to SDK consumers who haven't been following all of the intermediate steps on this branch
- clean up the `examples/` directory to standardize naming and group related examples more cohesively

All of these suggestions can wait until after we release though. The only thing I'd still like to get in before we release the Python features are #978

lib/python-sdk/common_grants_sdk/client/opportunities.py
SnowboardTechie: good idea, added with https://github.com/HHS/simpler-grants-protocol/pull/970/changes/9eae8c80e6a66f486b7f22d3cd133fb4f5a29d29

PR #975, opened 2026-07-02
feat(ts-sdk): typed custom-filters authoring surface and client (#967)

### LINKED ISSUE #967: [TS SDK] Prepare SDK for custom filters release

SnowboardTechie:
### Summary

- Plugin authors now get compile errors for route/method typos in `definePlugin()` (closed key unions) plus a runtime `FilterError` backstop at definition time, and registered filters type the consumer call site directly — one declaration, no second inert list.
- Plugin consumers get `plugin.getClient(config)`: a client pre-bound to the plugin's schemas and registered filters — responses parse with the plugin schema by default, `search({ filters })` autocompletes registered names and compile-checks value families.
- Filter validation is now fail-fast (an invalid value on any filter throws `FilterError` before the request is sent), and `search()`/`list()` partition rows into `{ items, errors }` so one malformed row no longer throws away the page.
- The resource layer is registry-driven for multi-resource readiness (organizations/applications/awards): `Resource` base with read primitives plus a `mutate` write primitive (`put`/`patch` added to the transport), so a new resource is a registry entry + typed slots.

- Fixes #967
- Time to review: 45 minutes

### Changes proposed

- **extensions/**: `PluginRoutes` closed to `Partial<Record<ResourceName, RouteMethods>>`; `FILTER_TYPE_SCHEMAS` literal-typed with a `CustomFilterInput<FT>` projection; `F.*` helpers made generic over value; `classifyFilters` flipped from fail-soft to fail-fast (`ClassifyResult` removed; the name stays consistent with Python's `classify_filters`); `definePlugin()` validates routes at definition time and wires `plugin.getClient()`.
- **client/**: new `client/resources/` (`Resource` base with `mutate`, `RESOURCE_REGISTRY`, moved `Opportunities`), new `client/results.ts` (`parseBatch`, `ParseFailure`, `ListResult`/`SearchResult`), `BuiltClient` projection + builder; `Client` de-genericized (constructor no longer takes `routes`) while keeping default base-schema resource slots for plugin-less use; transport gains `put`/`patch`; fetch helpers partition rows and use raw row counts for pagination.
- **Behavior contract**: invalid standard/registered filter values throw before the request; well-formed ad-hoc filters still pass through; `filterInfo.errors` now carries server-returned errors only; `get()` stays fail-hard; `statuses`+`filters.status` conflict now throws.
- **Tests/docs**: 545 tests (incl. compile-only type gates for typos, wrong value families, and the `BuiltClient` not-`any` projection); examples reworked to the consumer flow; extensions README updated; changeset (`minor`) documents the breaking changes.

### Context for reviewers

**Deviations from the [draft implementation plan](https://github.com/user-attachments/files/29517984/ts-sdk-client-extensions-plan.md) (flagged per the AC):**

- No `extensions/` file renames — types landed in the existing files, using the plan's own "the directory moves are optional cosmetics" clause. `client/resources/` + `results.ts` are genuinely new.
- `Client` keeps default (base-schema, no registered filters) resource slots rather than going strictly transport-only, so plain `new Client()` consumers keep working; the `routes` constructor param and generic are still removed.
- Registry is simplified at n=1 (single `schemaName` per entry, direct typed slot assignment instead of a loop); the `satisfies Record<ResourceName, …>` clause is the drift guard.
- The runtime route-typo backstop fires at `definePlugin()` (per this issue's AC) rather than at client construction (Python's timing).

**Deliberate TS-ahead-of-Python divergences (vs #966 / PR #970):**

- `onParseError: "collect" | "throw"` per-call option (Python always collects).
- A malformed **ad-hoc** filter value throws in TS ("no soft-degrade on input" per the plan); Python currently drops it as a non-strict error.
- Multi-resource registry + `mutate` write primitive (Python is opportunities-only).

**Unhappy-path verification (per the AC):**

- `examples/custom-filters.ts` — "Reviewer notes — unhappy paths" header block maps each guard to where it's proven; `pnpm example:custom-filters` runs the fail-fast demos with no server.
- Compile gates: `__tests__/extensions/plugin-routes-types.ts` (route/method typos, wrong value family, `BuiltClient` projection not-`any`), `__tests__/client/search-filters-types.ts`.
- Runtime: `__tests__/extensions/get-client.spec.ts` (FilterError before any request, definition-time backstop), `__tests__/client/opportunities.spec.ts` (no-request assertions, `{items, errors}` partition, server-only `filterInfo.errors`), `__tests__/client/resources.spec.ts` (write-verb mechanism via a test-only subclass).

**Verified:** `pnpm run checks` / `pnpm build` / 545 tests / examples all green; also exercised from a downstream-consumer harness against the packed tarball (definePlugin backstop, pre-request throw, `customFilters` wire nesting, partition, plugin-bound typed parsing all confirmed from the consumer import path).

**Migration heads-up:** `common-grants/ts-cg-grants-gov` does not compile against this branch (its checkout predates the current SDK surface generally, not just this PR); migration notes will accompany the release. Breaking changes are enumerated in the changeset.

### Additional information

```ts
const plugin = definePlugin({
  schemas: { Opportunity: { customFields: { programCode: { fieldType: "string" } } } },
  routes: { opportunities: { search: { filters: { region: { filterType: "stringArray" } } } } },
} as const);

const client = plugin.getClient({ baseUrl });
const result = await client.opportunities.search({ filters: { region: F.in(["US-CA"]) } });
//                                                            ^ autocompletes; F.eq() here is a compile error
result.items[0].customFields?.programCode?.value; // typed string — no per-call schema
result.errors;                                     // ParseFailure[] — malformed rows, with index + raw
```


lib/ts-sdk/src/extensions/custom-filters.ts
```diff
@@ -3,9 +3,10 @@
  *
  * Provides the classifier, validators, and F helper namespace for the custom-filters surface.
  *
- * - `classifyFilters` — transforms a flat consumer `filters` object into the
+ * - `categorizeFilters` — transforms a flat consumer `filters` object into the
```
widal001: Is there a reason we're renaming `classifyFilters` --> `categorizeFilters` everywhere?

If this is because that's the name of the private method in CommonBenefits, I'm okay with keeping it `classifyFilters` to stay consistent with the Python SDK.

Ideally we'd eventually make this a private function or method in both TypeScript or Python, since plugin authors and consumers shouldn't really need to use or know about it.

lib/ts-sdk/examples/custom-filters.ts
widal001, line 1: FYI - I refactored this example in #982 to remove the direct invocation of `categorizeFilters()` since SDK consumers should need to use this directly.

widal001 (approved):
I haven't gotten a chance to review this super in depth, but the plugin author and consumer interface seems solid!

I opened one small stacked PR that might be good to merge into this #982 to improve ad hoc filter validation.

There are a few extra safeguards that this branch introduces that we'll want to port over to Python: mostly the validation check at `define_plugin()` instead of `plugin.get_client()`

lib/ts-sdk/src/extensions/custom-filters.ts
SnowboardTechie: The rename was tracking the behavior change rather than the CommonBenefits name: when the fail-soft version got removed, the fail-fast throw-before-request classifier came in under a new name so the contract change would be loud. But since nothing consumes this surface yet and the plan is to bring Python to the same erroring pattern, I've reverted it in 94bba78. `classifyFilters` everywhere again, still fail-fast.

I'd land the privatization the same way, consumers shouldn't need it once `search()` is the path.

lib/ts-sdk/examples/custom-filters.ts
SnowboardTechie: Yeah, that's the better shape. The direct call was showing the request-body seam more than a real consumer flow.

PR #978, opened 2026-07-06
refactor(py-sdk): remove `routes` and `schemas` args from `Client`

### LINKED ISSUE #908: [SDK] Review custom filters
### Summary

Custom filters work is complete in both SDKs but unmerged. Billy gives a final review before a new SDK version ships.

### Acceptance criteria

- [x] Custom filters changes reviewed

widal001:
### Summary

Removes the `routes` and `schemas` arguments from the `Client` to make `Plugin.get_client()` the single correct pattern for getting a `Client` that has been scoped with a plugin's custom filters and custom fields.

- Fixes #908 
- Time to review: 3 minutes

- Removes `routes` and `schemas` from the `Client` args, to avoid confusing SDK consumers about if/when they should use those params vs. `plugin.get_client()`
- Adds `Client._bind_routes()` and `Client._bind_schemas()` to replace `Client` args as the mechanism for adding runtime scoping and IDE-level type inference for custom fields and filters.
- Enforces type checking on `examples/` and fixes one other pyright issue on `examples/custom_filters.py`

There were a few main motivations for this change:

1. The previous pattern introduced a few potential error scenarios that this refactor removes:
    1. If an SDK consumer passes in `routes` with the wrong type annotation, e.g. `Plugin[Any]` they'll get a pyright error when they pass valid filters into `search()`
    2. Even if an SDK consumer passes `routes` and `schemas` into `Client` correctly, the responses won't get properly typed with custom fields.
2. We want to avoid confusing SDK consumers who aren't using or creating plugins, since `routes` and `schemas` are only relevant in the context of customizing the client's default behavior, which is better handled through plugins.
3. We also want to avoid confusing Plugin consumers, who should only have one correct way of accessing a scoped client, per [the Zen of Python](https://peps.python.org/pep-0020/)

This is one of the errors plugin consumers can get when they use `Client(routes=...)` instead of `plugin.get_client()`

Screenshot 2026-07-06 at 9 28 28 AM

Switching to `plugin.get_client()` fixes this error.

Screenshot 2026-07-06 at 11 24 48 AM

### DIFF
--- a/lib/python-sdk/examples/custom_filters.py
+++ b/lib/python-sdk/examples/custom_filters.py
@@ -54,7 +54,7 @@ class OppSearchFilters(OpportunityFilters, total=False):
     fundingProgram: StringComparison
 
 
-routes: PluginRoutes = PluginRoutes(
+routes: PluginRoutes[OppSearchFilters] = PluginRoutes(
     opportunities=ResourceRoutes(search=OppSearchFilters)
 )
 
--- a/lib/python-sdk/common_grants_sdk/client/client.py
+++ b/lib/python-sdk/common_grants_sdk/client/client.py
@@ -1,11 +1,12 @@
 """HTTP client for the CommonGrants API.
 
 ``BaseClient`` is the transport plumbing (auth headers, paginated GET/POST). The
-generic ``Client`` layers the typed resource facade on top: it binds a plugin's
-route filters (``FiltersT``) and Opportunity schema (``ItemT``) so
+generic ``Client`` layers the typed resource facade on top. Scope it to a plugin
+with ``plugin.get_client(...)``: that binds the plugin's route filters
+(``FiltersT``) and Opportunity schema (``ItemT``) so
 ``client.opportunities.search(filters=...)`` is typed and responses parse with the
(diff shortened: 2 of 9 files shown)


lib/python-sdk/examples/custom_filters.py
```diff
@@ -54,7 +54,7 @@ class OppSearchFilters(OpportunityFilters, total=False):
     fundingProgram: StringComparison
 
 
-routes: PluginRoutes = PluginRoutes(
```
widal001, line 57: This fixes another pyright issue. The previous version widened the `PluginRoutes` types to the default `OppFilters` rather than using `OppSearchFilters`

SnowboardTechie (approved):
Nice work, this reads clean.

One small non-blocking thing: two docstrings didn't catch up with the move. The `client.py` module docstring still offers "Construct one directly, or via `plugin.get_client(...)`" as equivalent paths, and `_registered_filter_models` in `extensions/filters.py` still says `validate_routes` is "run at `Client` construction". Fine as a fast-follow if you'd rather land this as is.

Approving.

PR #988, opened 2026-07-08
fix(py-sdk): align python and typescript

### LINKED ISSUE #986: [Py SDK] Close feature gaps with TS SDK
### Summary

Close a few important feature/functionality gaps related to custom filters between the TypeScript and Python SDKs.

### Acceptance criteria

- [x] Invalid `routes=` inputs are raised to plugin consumers on `define_plugin()` rather than to plugin consumers at `Plugin.get_client()`
- [x] Invalid `filters=` inputs, including invalid ad hoc filter combinations, are raised to plugin consumers _before_ a request to the API has been made
- [x] Acceptable type annotations for custom fields are limited to `DefaultFilters` (or something similar) rather than the less restrictive `CommonGrantsBaseModel`

widal001:
### Summary

Update the Python SDK to close some feature and DX gaps that were previously addressed in the TypeScript SDK

- Fixes #986 
- Time to review: 15 minutes

- [Adds an `IntegerComparison` filter](https://github.com/HHS/simpler-grants-protocol/commit/7cdb018643d21de722e9682349820785ad06080a) and groups this (and other filters) into `VALID_FILTER_MODELS`
- [Validates custom filters in `define_plugin()`](https://github.com/HHS/simpler-grants-protocol/commit/e9c83ccaf9f4431bb16429ddf264194b2ff4d3b3) to prevent invalid filter types from slipping through and getting caught by the plugin consumer when they run `plugin.get_client()`
- [Raises invalid filters passed by plugin consumers](https://github.com/HHS/simpler-grants-protocol/commit/c74de075ffb7cc4799c9c535c8778f96af616f29) before a search request is sent to the API. This helps surface plugin consumer mistakes immediately instead of silently ignoring or dropping them.
- [Refactors the custom filters examples](https://github.com/HHS/simpler-grants-protocol/commit/97c454ea624bf52a991a2fc7255d2df5ae06b144) to highlight both happy and unhappy paths more clearly.

1. Checkout the PR, and go to: `lib/python-sdk/`
2. Run `poetry run python examples/mock_api_server.py` to start the mock server
3. Run  `poetry run python examples/custom_filters.py` to run the examples
4. Inspect the `examples/custom_filters.py` example for examples of which edge cases are handled

Screenshot 2026-07-08 at 4 29 04 PM


lib/python-sdk/examples/custom_filters.py
widal001, line 1: I refactored this file to more closely match the `lib/ts-sdk/examples/custom-filters.ts`

lib/python-sdk/examples/mock_api_server.py
widal001, line 1: Added this to support a self-contained examples test

lib/python-sdk/common_grants_sdk/schemas/pydantic/filters/integer.py
```diff
@@ -0,0 +1,45 @@
...
+    operators (``eq``/``neq``). The value must be an integer: a non-integral value
+    (e.g. 3.5) or a bool is rejected. Implements the core spec's
+    IntegerComparisonFilter (filters/numeric.tsp).
+    """
+
+    operator: ComparisonOperator | EquivalenceOperator = Field(
+        ...,
+        description="The comparison operator to apply to the filter value",
+    )
+    value: int = Field(..., description="The integer value to compare against")
```
SnowboardTechie, line 26: Python rejects 3.5 here, but TS `integerComparison` still aliases `NumberComparisonFilterSchema` (custom-filters.ts) and accepts it, so a registered `integerComparison` filter diverges across SDKs. Ad-hoc stays aligned since the classifier hits `NumberComparisonFilter` first. Given the PR title, do we want a TS follow-up to close it, or is Python leading the spec here on purpose?

SnowboardTechie (approved):
Ran through it against HOLD-filters, tests and CI look solid. Approving but with two notes:

The changeset is stale. `python-typed-custom-filters.md` still says ad-hoc keys "stay best-effort and are dropped if unusable," which is the behavior this PR reverses, and it doesn't mention the new `IntegerComparisonFilter` or the `ClassifyResult` / `FilterError.strict` removal.

lib/python-sdk/common_grants_sdk/schemas/pydantic/filters/integer.py
widal001: This is a good flag, but it warrants a deeper discussion. I'll start a thread in Slack but I think there are actually a few options:

1. Tighten the TS SDK `IntegerComparisonFilter`
2. Loosen the Python SDK `IntegerComparisonFilter`
3. Drop `IntegerComparisonFilter` and treat `NumberComparisonFilter` as a superset of both integer and float (which is how the JSON `number` type is treated)

PR #1006, opened 2026-07-15
feat(core): add ExtensibleEnum schema and templated model

### LINKED ISSUE #961: [Core] Add an `ExtensibleEnum` schema and templated model
### Summary

**Background:** We have a very consistent pattern for defining extensible enum fields (e.g. object fields that have a `value` property from a fixed `enum` with a `custom` option, an optional `customValue` property, and an optional `description` property). Examples of this pattern include [`OppStatus`](https://commongrants.org/protocol/models/opportunity/#oppstatus) and [`AppStatus`](https://commongrants.org/protocol/models/application/#appstatus) which define common status options, but allow implementations to extend these options with custom status values for their implementation in a structured way.

**Problem:** This pattern isn't formally documented or enforced anywhere which places the burden on maintainers to enforce consistency, results in duplicate code and changes if this pattern is modified, and could introduce drift amongst fields that follow this pattern.

**Recommendation:** We should introduce a new `ExtensibleEnum` base model like we have in [CommonBenefits](https://commonbenefits.dev/protocol/fields/extensible-enum/#extensibleenum) (along with a [templated equivalent](https://github.com/common-benefits/common-benefits/blob/c164f3431521834f1dfa7275af08448bfe742a1e/lib/core/lib/fields/extensible-enum.tsp#L38-L49)) and then update the existing extensible enum fields in CommonGrants to use this like [`AppStatus` in CommonBenefits](https://github.com/common-benefits/common-benefits/blob/c164f3431521834f1dfa7275af08448bfe742a1e/lib/core/lib/models/application.tsp#L27-L29).

### Acceptance criteria

- [x] A new `ExtensibleEnum` field and `ExtensibleEnumT` has been added to the core library
- [x] Existing extensible enum fields have been updated to use `ExtensibleEnumT` (e.g. `OppStatus` and `AppStatus`)
- [x] Any breaking changes this introduces have been reviewed with @widal001 and other engineers
- [x] The approach to implementing this either closely matches the CommonBenefits protocol, or the PR explains why an easier option exists

SnowboardTechie:
### Summary

- Fixes #961
- Time to review: 15 minutes

### Changes proposed

- Adds a `Fields.ExtensibleEnum` base model and `Fields.ExtensibleEnumT<T>` template to the core library, formalizing the extensible enum pattern (`value` + optional `customValue` + `description`) that our status-style fields all repeat by hand.
- Rewrites the five existing extensible enum fields — `OppStatus`, `AppStatus`, `ApplicantType`, `FormResponseStatus`, `CompetitionStatus` — as one-line instantiations of the template, so the pattern has a single source of truth.
- Adds the `custom` option to `FormResponseStatusOptions` at v0.4.0 — its docstring already listed `custom`, but the enum was missing it (the only one of the five that was).
- Adds a docs page for the new field at `/protocol/fields/extensible-enum/`, regenerates the committed schema artifacts, and re-pins the TypeSpec line references in the affected model pages.

### Context for reviewers

**Approach matches CommonBenefits** ([extensible-enum.tsp](https://github.com/common-benefits/common-benefits/blob/c164f3431521834f1dfa7275af08448bfe742a1e/lib/core/lib/fields/extensible-enum.tsp)): an untyped `ExtensibleEnum` plus a typed `ExtensibleEnumT<T>`, consumed as `model AppStatus is Fields.ExtensibleEnumT<AppStatusOptions>;`. One deliberate wording difference: CommonBenefits' template property docs say "typed to `T`", which leaks the template parameter into every emitted consumer schema — I reworded the property docs to be reader-facing instead.

**Breaking-change review** (per the issue's acceptance criteria):

- **No structural schema changes.** I diffed the full emitted output (JSON Schema + OpenAPI 0.1.0–0.4.0) against the branch point: `required`, `$ref` targets, types, and enum membership are all byte-identical, except the intended `custom` addition, which appears only in the 0.4.0 output (the `@added(v0_4)` gate works — 0.2.0/0.3.0 emit the enum without it).
- **Property descriptions change** in the emitted schemas for the five refactored models — each model's bespoke wording ("The status of the opportunity, from a predefined set of options") is replaced by the template's shared wording ("The selected value, from a predefined set of options"). This is docs-only, but it does alter published-version artifacts, so calling it out explicitly.
- The TS SDK fuzz suite (Zod vs. emitted JSON schema, 474 tests) passes with zero SDK changes, which is the working proof that validation semantics are unchanged. Python SDK is likewise untouched.

Also fixed in passing: the `Examples.Competition.status` example set `customValue: "custom"` while `value` was `open` — the custom value only applies when `value` is `custom`, so it's dropped.

Verified with: `pnpm --filter @common-grants/core run checks && build && typespec`, full emitted-schema diff against the branch point, `pnpm --filter @common-grants/sdk run build && test:coverage`, and a full website build (163 pages, links validated).

### Additional information

Defining a new extensible enum field is now:

```typespec
enum MyStatusOptions {
  active,
  inactive,
  custom,
}

model MyStatus is Fields.ExtensibleEnumT<MyStatusOptions>;
```


lib/core/lib/core/fields/extensible-enum.tsp
```diff
@@ -0,0 +1,69 @@
...
+model ExtensibleEnum {
+  /** The selected value. May be from a predefined set of options or a custom value */
+  value: unknown;
+
+  /** A custom value, used when the selected value does not fit a predefined option */
+  customValue?: string;
+
+  /** A human-readable description of the value */
+  description?: string;
+}
```
widal001: Instead of defining `ExtensibleEnum` and `ExtensibleEnumT` separately and having to keep them in sync, I'm pretty sure you can do this:

```tsp
model ExtensibleEnumT<T> {
  /** The selected value, from a predefined set of options */
  value: T;

  /** A custom value, used when the selected value is the `custom` option */
  customValue?: string;

  /** A human-readable description of the value */
  description?: string;
}

model ExtensibleEnum is ExtensibleEnumT<unknown>;
```

lib/core/lib/core/fields/extensible-enum.tsp
```diff
@@ -0,0 +1,69 @@
...
+
+/** A typed extensible enum: `value` is constrained to a caller-supplied enum (or
+ * other type), with a free-form `description` for annotation or custom-value
+ * details.
```
widal001: Something like this is more appropriate as the docs for `ExtensibleEnum`, which is the user-facing, publicly emitted JSON schema, and the `ExtensibleEnumT` should focus on how and why maintainers should use it to define extensible enums.

The example is great, though!

lib/core/lib/core/fields/extensible-enum.tsp
```diff
@@ -0,0 +1,69 @@
...
+// Model definitions
+// ########################################
+
+/** A value from an open-ended set of options: a `value` of unspecified shape,
+ * an optional `customValue` for caller-defined options, and an optional
+ * free-form `description`. Use this when the value set is open-ended and a
+ * typed constraint isn't appropriate. For typed variants, see `ExtensibleEnumT<T>`.
+ */
+@example(Examples.ExtensibleEnum.customStatus)
+@doc("A value from an open-ended set of options")
```
widal001: While it's _technically_ true that this version of an extensible enum is has a `value` of an unspecified shape, that's not what this model represents.

This model represents a pattern for letting APIs choose a value from a known set of options that can also be extended with a custom value when necessary.

And because this model generates the JSON Schema which we'll be adding to the docs for pattern, we should document it here rather than describing it as being (literally) unspecified.

lib/core/lib/core/fields/extensible-enum.tsp
widal001: Though we wouldn't want to use the phrase "caller-supplied" in the docs for the non-generic `ExtensibleEnum` to avoid confusing readers that it refers to API consumers

lib/core/lib/core/models/form-response.tsp
```diff
@@ -59,6 +50,9 @@ enum FormResponseStatusOptions {
   notStarted,
   inProgress,
   complete,
+
+  @Versioning.added(CommonGrants.Versions.v0_4)
+  custom,
```
widal001, line 55: Thanks for catching this!

lib/core/lib/core/fields/extensible-enum.tsp
```diff
@@ -0,0 +1,69 @@
...
+ *   closed,
+ *   custom,
+ * }
+ *
+ * model OppStatus is ExtensibleEnumT<OppStatusOptions>;
+ * ```
+ */
+@Versioning.added(CommonGrants.Versions.v0_4)
+model ExtensibleEnumT<T> {
+  /** The selected value, from a predefined set of options */
```
widal001, line 29: I think you might be able to use the template in this description like `@doc("The selected value from {T}, T)` to make the descriptions in `ExtensibleEnumT<OppStatusOptions>` more descriptive

widal001 (commented):
This is looking really good! But I'd love to get a few improvements in before we merge. See my comments below

lib/core/lib/core/fields/extensible-enum.tsp
SnowboardTechie: Yeah, that works. Took your exact shape in abb42f6: the template is now the only definition and the untyped model derives from it, so there's nothing left to keep in sync. Emitted output is unchanged except the ExtensibleEnum property descriptions, which now come from the template.

lib/core/lib/core/fields/extensible-enum.tsp
SnowboardTechie: Reworked in abb42f6: the pattern description now lives on ExtensibleEnum (that's what lands in the emitted JSON schema), and the template docs are maintainer-facing, how and why to use it to define new fields, with the example kept there. "Caller-supplied" is gone too.

lib/core/lib/core/fields/extensible-enum.tsp
SnowboardTechie: Reworded in abb42f6. The emitted description is now "A value from a predefined set of options that can be extended with a custom value when needed", and the doc comment describes the pattern rather than the unconstrained shape.

lib/core/lib/core/fields/extensible-enum.tsp
SnowboardTechie: Tried this, and the compiler rejects it: @doc's formatArgs parameter is typed `{}`, so passing an enum fails with "Argument of type 'CommonGrants.Models.OppStatusOptions' is not assignable to parameter of type '{}'". The `{name}` interpolation only accepts model types. If the per-model wording feels worth it, per-model `@@doc` augments on each instantiation would work, but that reintroduces the duplication this refactor removes, so I kept the shared wording.

widal001: Gotcha, didn't realize it was restricted to models. Shared language works fine!

widal001 (approved):
Thanks for making those updates!

PR #1007, opened 2026-07-15
feat!(core): add non-templated response schemas and rename templates with T suffix

### LINKED ISSUE #960: [Core] Created non-templated response schemas to support versioning
### Summary

**Problem:** When TypeSpec models have templates (e.g. [`Ok<T>`](https://github.com/HHS/simpler-grants-protocol/blob/98e149afd2059cb67d25246175fb53376e4374dc/lib/core/lib/core/responses/success.tsp#L40-L48) or [`Paginated<T>`](https://github.com/HHS/simpler-grants-protocol/blob/98e149afd2059cb67d25246175fb53376e4374dc/lib/core/lib/core/responses/success.tsp#L70-L82)) they don't get emitted as JSON schemas by TypeSpec. This makes it hard to document changes to those fields, like #959, in our schema changelog and generate the correct documentation for them.

**Recommendation:** The CommonBenefits core library demonstrates how to solve this problem using a combination of base, templated, and non-templated schemas to solve this problem by: 
- Keeping the properties of these templated and non-templated schemas stay in sync
- Enabling maintainers to generate new schemas from these templates: [`Responses.PaginatedT<Models.ProgramBase>`](https://github.com/common-benefits/common-benefits/blob/c164f3431521834f1dfa7275af08448bfe742a1e/lib/core/lib/routes/programs.tsp#L33)
- Generating JSON schemas for these base [response schemas](https://commonbenefits.dev/protocol/responses/success/#ok)

### Examples

- [Ok response](https://github.com/common-benefits/common-benefits/blob/c164f3431521834f1dfa7275af08448bfe742a1e/lib/core/lib/responses/success.tsp#L24-L62) doesn't need a base schema
- [Paginated schema](https://github.com/common-benefits/common-benefits/blob/c164f3431521834f1dfa7275af08448bfe742a1e/lib/core/lib/responses/success.tsp#L68-L102) does require a base schema

### Acceptance criteria

- [x] These base response schemas (e.g. `Ok`, `Paginated`, `Filtered`, etc.) now have JSON schemas generated for them
- [x] These base response schemas have been added to the website docs for responses with versioning, starting with the version in which their templated equivalent was added (so v0.1.0 for most)
- [x] The templated versions of these schemas are now suffixed with `T` (e.g. `OkT`) and are still used to generate response bodies for the routes
- [x] The approach to implementing this either closely matches the CommonBenefits protocol, or the PR explains why an easier option exists (e.g. TypeSpec now supports this natively through `tsconfig.json` -- though I don't think they do at the moment)

SnowboardTechie:
### Summary

- Fixes #960
- Time to review: 15 minutes

### Changes proposed

- Adds non-templated `Ok`, `Paginated`, `Sorted`, `Filtered`, and `Created` response models so they are emitted as standalone JSON schemas and picked up by the schema changelog (all versioned from v0.1.0, matching when their templated equivalents were added).
- Renames the templated variants with a `T` suffix (`OkT<T>`, `PaginatedT<T>`, `SortedT<T>`, `FilteredT<ItemsT, FilterT>`, `CreatedT<T>`); routes still use these to generate response bodies.
- Documents the concrete schemas on the website responses pages (including a new `Created` section) with `<SchemaChangelog>` blocks, replacing the previously hand-written JSON schema snippets with the emitted schema files.
- Adds a minor changeset for `@common-grants/core` — specs consuming the old templated names (e.g. `Responses.Ok<T>`) must switch to the `T`-suffixed equivalents.

### Context for reviewers

The approach follows the CommonBenefits core library's base/templated schema split, with one adaptation: the templated variants **spread** a shared `*Base<T>` model instead of extending it. Extending a templated parent makes the openapi3 emitter declare the parent instantiations as named schemas, which fails with `duplicate-type-name` errors in our setup; spreading inlines the shared properties the same way the previous `...Paginated<T>` composition did. The shared base still keeps the concrete and templated variants in sync. TypeSpec still has no native way to emit uninstantiated templates as schemas, so the base-model approach remains necessary.

Notes on the generated-output diffs:

- Route response bodies in the OpenAPI output are unchanged apart from description strings; inherited `status`/`message` now arrive via `allOf: Success` composition instead of inline flattening (semantically equivalent).
- `Success.yaml` loses `unevaluatedProperties: not: {}` — expected: the emitter unseals schemas that are now composed via `allOf` by the new concrete response schemas. The concrete response schemas re-seal at their own level.
- Regenerating on the `HOLD-core-v0.4.0` base also syncs the CLI OpenAPI copies (`lib/cli/lib/openapi/`), including a new `openapi.0.4.0.yaml` — earlier merges to the hold branch (awards, identifiers, the application `name` → `title` rename) had only regenerated the website copies, so the CLI files pick up that drift here alongside the response-schema changes.

Verified with `pnpm run ci` at the repo root (all packages green) and by inspecting the built responses pages, which now render the emitted schemas and v0.1.0 changelog tables for `Ok`, `Created`, `Paginated`, `Sorted`, and `Filtered`.

### Additional information

New emitted schemas: `Ok.yaml`, `Paginated.yaml`, `Sorted.yaml`, `Filtered.yaml`, `Created.yaml` (see `website/public/schemas/yaml/`). Example (`Paginated.yaml`):

```yaml
$schema: https://json-schema.org/draft/2020-12/schema
$id: Paginated.yaml
type: object
properties:
  items:
    type: array
    items: {}
    description: Items from the current page
  paginationInfo:
    $ref: PaginatedResultsInfo.yaml
    description: Details about the paginated results
required:
  - items
  - paginationInfo
allOf:
  - $ref: Success.yaml
unevaluatedProperties:
  not: {}
description: A 200 response with a paginated list of items
```


lib/cli/lib/openapi/openapi.0.1.0.yaml
```diff
@@ -8,6 +8,8 @@ info:
     it must implement all of the routes with the "required" tag in this specification.
   version: 0.1.0
 tags:
+  - name: Awards
+    description: Endpoints related to grant awards
```
widal001: I don't _think_ should have been preserved in the auto-generated OpenAPI specs for this version, since `pnpm typespec:openapi` should run a script that strips out tags without any routes.

Either this PR is missing a recent update from `HOLD-core-v0.4.0` or some change in the core library is causing some awards to leak through into v0.1.0

widal001: Oh wait, I'm realizing now that this is in the CLI. Bumping these OpenAPI specs should happen separately once we've locked down the updates to the core library and will also require a minor changeset for the CLI, since it changes the default spec used by `cg check spec`

widal001 (approved):
Once you remove the CLI openAPI changes (which should be handled in a separate PR) this should be good to merge!

lib/cli/lib/openapi/openapi.0.1.0.yaml
SnowboardTechie: Got it, that makes sense. Reverted the bundled CLI specs to their HOLD-core-v0.4.0 versions and dropped the new 0.4.0 file in 36d9d14, so this PR no longer touches the default spec for `cg check spec`.

PR #1010, opened 2026-07-15
feat(py-sdk)!: derive camelCase wire naming from a shared alias generator

SnowboardTechie:
### Summary

- Fixes the `event_type` pyright errors surfaced by the plugin's pyright switch (common-grants/py-cg-grants-gov#11) with the universal approach from [the review thread](https://github.com/HHS/simpler-grants-protocol/pull/1010#discussion_r3590392259): camelCase wire naming comes from a shared `AliasGenerator` on the model config, and ~60 per-field `alias=` declarations are deleted.
- Time to review: 10 minutes

### Changes proposed

- `CommonGrantsBaseModel` config gains `alias_generator=AliasGenerator(validation_alias=to_camel, serialization_alias=to_camel)` + `populate_by_name=True`; the same config (`CAMEL_WIRE_CONFIG`) is applied to the wire models that don't extend the base (sorting, pagination, filters).
- Delete every per-field alias declaration whose value matches `to_camel(field_name)` (~60 lines across the pydantic models). The one irregular wire name (`CustomField.schema_url` → `"schema"`) keeps its explicit field-level alias, which overrides the generator.
- Internal camelCase constructor calls (pagination aggregation, filter classification, default sorting) move to snake_case field names — pyright now accepts field names and only field names, which is the direction this PR pushes everywhere.
- Harden the internal `field_info.alias` readers (filter classification, transform mapping validation) to consult all three alias fields.
- Breaking for type-checked consumers: constructor kwargs are snake_case field names as of this release (runtime still accepts camelCase via validation aliases; the wire format is unchanged). Ships as a minor bump via the `feat!` title.

### Context for reviewers

Why the generator works where config flags don't: pyright ignores `populate_by_name`/`validate_by_name` and keys the synthesized constructor off a field-level `alias`; with a generator it can't evaluate the callable, so it falls back to field names — which then also validate at runtime via `populate_by_name`.

Wire-safety verification: every deleted declaration's value was mechanically compared against `to_camel(field_name)` (exact match required; mismatches kept), and a model-walk confirmed no multiword field anywhere lacks an alias — so the generated aliases reproduce the previous wire format exactly.

Verified: SDK `make checks` (pyright 0 errors) + 397 tests (including wire serialization/round-trip tests). Re-verified against the grants.gov plugin's current `main` with the branch-built wheel installed: all 34 plugin tests pass (runtime and wire behavior unchanged), and pyright flags the plugin's camelCase constructor kwargs (15 errors), which is the expected consumer-facing effect of this change. The plugin's fix is a mechanical rename of those kwargs to snake_case, done in the same PR that bumps its `common-grants-sdk` pin to the release including this.

### Additional information

Follow-up potential: the TS SDK has no equivalent gap (its zod schemas key off wire names), and generated models inherit this convention from the base for free.

BREAKING CHANGE: SDK model constructors are typed against snake_case field names; rename camelCase keyword arguments when instantiating models. Attribute reads, runtime validation, and the wire format are unchanged.


lib/python-sdk/common_grants_sdk/schemas/pydantic/fields/event.py
```diff
@@ -29,7 +29,8 @@ class EventBase(CommonGrantsBaseModel):
     )
     event_type: EventType = Field(
         ...,
-        alias="eventType",
+        validation_alias="eventType",
+        serialization_alias="eventType",
```
widal001: @SnowboardTechie Have we explored a more universal way to fix this alias issue? Like a config on the `CommonGrantsBase`?

SnowboardTechie: We hadn't fully. Only the populate/validate config flags got tested, which pyright ignores, so the per-field fix looked like the only option.

Looking into it more after your comment, an AliasGenerator on the base does work, and the migration risk checks out. Every multiword field already has an alias so the wire doesn't change, and irregular aliases like `schema` keep their explicit overrides since field-level settings beat the generator.

I'm going to rework this PR to the generator approach, it deletes the per-field alias declarations instead of converting them. Longer term this also sets us up better for generated models, a wire-naming convention on the base is something an emitter inherits for free instead of something it has to emit per field.

SnowboardTechie:
Moving this PR back to draft so we can give the AliasGenerator change broader regression testing before including it in a release. The approach still looks like the right long-term fix, but its reach across the SDK models warrants more stress testing.

I’ll handle the narrower Event alias fix needed for the Python plugin runtime bug in a separate PR against `HOLD-filters`, then return to this change for a subsequent patch release.

Context: [Slack discussion](https://betagrantsgov.slack.com/archives/C05TSB1D1SP/p1784164164270299?thread_ts=1783985878.225589&cid=C05TSB1D1SP)

lib/python-sdk/common_grants_sdk/schemas/pydantic/fields/event.py
SnowboardTechie: @widal001 Closing the loop on the stress testing we held this for. The generator itself checks out: SDK gate green, wire format verified byte-identical, plugin runtime unaffected. But the wider testing did surface one real consequence, so this isn't patch material anymore. Pyright flips from typing constructors against the camelCase aliases to the snake_case field names, which breaks type-checked code that instantiates SDK models, including the `eventType=` / `customValue=` constructor calls we added to the Python plugin as the targeted fix during the filters release. Runtime and wire behavior are unchanged. So it's retitled `feat!` and ships as a minor with a migration note in the changelog (the patch changeset is gone along with changesets generally, release-please handles it now), and the plugin renames those kwargs in the same PR that bumps its SDK floor. Full verification details are in the updated description.

Personally I still think this is the right trade, snake_case constructors are the direction a Python SDK should settle pre-1.0 and the generator makes forgotten aliases impossible. But a minor with a forced rename is a bigger ask than the patch we originally scoped, so if you'd rather sit on it longer or take a different path, push back.

widal001 (approved):
Thanks for re-surfacing this one and for making these changes! Love seeing all of the aliases removed

PR #918, opened 2026-06-16
Custom filters PoC (TypeScript + Python)

SnowboardTechie:
### Summary

- Integration branch rolling up the custom-filters PoC work for both SDKs: #884 (TypeScript, issue #868) and #885 (Python, issue #869).
- Stacked on `HOLD-transforms` (PR #906). Stays in **draft** until #906 merges to `main`, then this PR re-targets `main`.
- Closes #869.
- Time to review: ~5 min. Both commits were already reviewed and merged via #884 and #885; this PR only stacks them onto the transforms bucket.

### Changes proposed

> Two commits, no new code beyond the already-merged sub-PRs.

- **TypeScript custom filters PoC** (#884, issue #868), in `lib/ts-sdk`.
- **Python custom filters PoC** (#885, issue #869), in `lib/python-sdk`.

Diff against the `HOLD-transforms` base: 25 files, +2,959 / -21, confined to `lib/ts-sdk` and `lib/python-sdk`.

### Context for reviewers

- **Why stacked on HOLD-transforms, not main:** HOLD-filters builds directly on HOLD-transforms. Targeting HOLD-transforms keeps this diff scoped to the custom-filters work; the transforms changes are reviewed in #906. Targeting main would re-show every transforms commit (69 files, +9,370) and duplicate that review surface.
- **Merge order:** do not merge until #906 lands on `main`. Once it does, re-target this PR's base to `main`; the three-dot diff narrows to exactly these two commits.
- **Verification:** each sub-PR (#884, #885) carried its own checks and was reviewed before merging into HOLD-filters. No new code is introduced here.
- HOLD-filters currently trails HOLD-transforms by 15 commits. The three-dot diff is unaffected, but I can sync the branch before re-targeting to main so CI runs against the latest base.

### Additional information

The custom-filters API surface and three-bucket classification are exercised by the runnable examples added in #884 (TypeScript) and #885 (Python); see those PRs for demo output.

PR #1011, opened 2026-07-15
fix(sdk): drop integerComparison from both SDK filter catalogs

SnowboardTechie:
### Summary

- Resolves the integer-vs-number filter divergence from [the #988 review thread](https://github.com/HHS/simpler-grants-protocol/pull/988#discussion_r3552572570): `integerComparison` is removed from both SDKs, making `NumberComparisonFilter` the single numeric-comparison type (option 3 from that thread).
- Time to review: 5 minutes

### Changes proposed

- Python SDK: delete `IntegerComparisonFilter` (model, exports, `IntegerComparison` authoring alias, `VALID_FILTER_MODELS` entry) and its three tests.
- TS SDK: remove `integerComparison` from the `CustomFilterType` union, the `FILTER_TYPE_SCHEMAS` map (it was already just an alias for `NumberComparisonFilterSchema`), the type fixtures/tests, and the README catalog (now 10 types).

### Context for reviewers

The divergence: Python's `IntegerComparisonFilter` rejected `3.5` while TS aliased `integerComparison` to the number schema and accepted it — the same registered filter behaved differently per SDK. Dropping the type entirely (rather than keeping an alias key) leaves the two catalogs identical, and matches the spec state of this release: after 886a384's revert moved the core integer filter model to the v0.4.0 train, `integerComparison` had no spec model behind it. `NumberComparisonFilter` already accepts integer values, consistent with JSON Schema's `number` type. If core v0.4.0 later ships a real integer filter family, the type can return in both SDKs together.

Nothing released ships this type and neither grants.gov plugin registers it, so there's no consumer impact; no changeset needed (it removes part of an unreleased addition, and the existing filters changesets don't mention integers).

Verified: Python `make checks` + 396 tests green; TS `pnpm run checks` + 546 tests green.

### Additional information

Targets `HOLD-filters` to land before #918, per the release-train ordering.


widal001 (approved):
Thanks for making this change!

PR #1024, opened 2026-07-21
fix(ts-sdk): preserve plain dates in JSON

### LINKED ISSUE #1016: [TS SDK] Plain dates do not round-trip through JSON serialization
## Summary

In `@common-grants/sdk` 0.5.0, `ISODateSchema` transforms a valid protocol date such as `2026-07-15` into a JavaScript `Date`. Standard JSON serialization changes it to `2026-07-15T00:00:00.000Z`, which the same schema rejects because CommonGrants requires `YYYY-MM-DD`.

This means an SDK-validated model can become invalid CommonGrants JSON when returned from a Node API or otherwise serialized.

## How to reproduce

```ts
import { ISODateSchema } from "@common-grants/sdk/schemas";

const parsed = ISODateSchema.parse("2026-07-15");
const serialized = JSON.parse(JSON.stringify(parsed));

console.log(parsed); // Date(2026-07-15T00:00:00.000Z)
console.log(serialized); // "2026-07-15T00:00:00.000Z"
console.log(ISODateSchema.safeParse(serialized).success); // false
```

The same failure occurs in enclosing models such as `OpportunityBaseSchema` at `keyDates.postDate.date`.

## Root cause and impact

The protocol defines `isoDate` as a date-only `plainDate`, but the TypeScript schema ends with:

```ts
z.string().date().transform(str => new Date(str))
```

The inferred SDK types and client therefore expose `Date` objects for `SingleDateEvent.date` and `DateRangeEvent.startDate`/`endDate`. JavaScript serializes those as timestamps, and the SDK does not currently expose a wire serializer or separate wire type that converts them back to date-only strings.

## Acceptance criteria

- [ ] Add a regression test for `parse → JSON serialization → parse` using a date-only event.
- [ ] Provide a supported serialization path that emits date-only fields as `YYYY-MM-DD`.
- [ ] Keep full timestamps invalid for protocol date-only fields.
- [ ] Document how parsed SDK/client values should be serialized back to CommonGrants JSON.

No matching open or closed issue was found using searches for `ISODateSchema`, date serialization, plain dates, and SDK round trips.

mikec-ai:
### Summary

- Fixes #1016
- Time to review: 10 minutes

### Changes proposed

- Serialize parsed plain-date SDK values as `YYYY-MM-DD` through standard JSON serialization.
- Keep `ISODateSchema` output typed as `Date`, preserving the existing public SDK contract and `instanceof Date` behavior.
- Add scalar and enclosing `SingleDateEvent` JSON round-trip regression tests.
- Add a patch changeset for `@common-grants/sdk`.

### Context for reviewers

The protocol defines `isoDate` as a date-only `YYYY-MM-DD` value. The SDK currently parses that value into a JavaScript `Date`, whose normal `toJSON()` output is a full timestamp. That serialized timestamp is rejected by the same schema.

This patch uses a small internal `Date` subclass that emits the protocol's date-only wire representation from `toJSON()`. The transform is explicitly typed as `Date`, so the internal subclass does not leak into inferred or generated public types. Returning strings directly would be simpler internally but would be a breaking change for consumers using the current `Date` output.

Verification:

- Full TypeScript SDK suite: 548 tests passed
- TypeScript typecheck
- TypeSpec compilation
- ESLint and Prettier checks
- Confirmed generated declarations continue to expose `Date`

One narrow limitation is that copying the parsed value through `structuredClone()` or `new Date(parsedDate)` produces an ordinary `Date` and therefore does not preserve the specialized serialization behavior.

### Additional information

Before: `JSON.stringify(ISODateSchema.parse("2026-07-15"))` emitted `"2026-07-15T00:00:00.000Z"`.

After: it emits `"2026-07-15"`, which reparses successfully.


widal001 (approved):
Thanks for catching and fixing this!

lib/ts-sdk/src/schemas/zod/types.ts
```diff
@@ -56,6 +56,14 @@ const acceptDate =
   (val: unknown): unknown =>
     val instanceof Date && !isNaN(val.getTime()) ? toStr(val) : val;
 
+class PlainDate extends Date {
```
SnowboardTechie, line 65: Worth pinning the copy gotcha from the description on the class so it ships with the code:

```suggestion
/**
 * A `Date` whose `toJSON()` emits the date-only wire format (`YYYY-MM-DD`)
 * so values parsed by `ISODateSchema` round-trip through `JSON.stringify`.
 * Copies (e.g. `structuredClone()` or `new Date(value)`) are ordinary
 * `Date`s and lose this behavior.
 */
class PlainDate extends Date {
```

SnowboardTechie (approved):
Nice work, this is a clean fix. One small suggestion inline, non-blocking. Approving.

Issue #1032, opened 2026-07-22
[Core] Audit and clean up the v0.4.0 hold branch


laurabelinfante:
### Summary

Before v0.4.0 (org identifiers, org profile syncing, and award routes and models) is published, review the hold branch with fresh eyes for inconsistencies.

### Acceptance criteria

- [x] The v0.4.0 hold branch is reviewed for inconsistencies
- [x] Fixes are applied or ticketed before the v0.4.0 release

SnowboardTechie:
Keeping this open for a second cleanup PR against the hold branch, covering the rest of what came out of review on #1040:

- Add a `BooleanComparison` entry to the docs and link to it from the custom filters entry in the Plugins section (from #1045)
- Double-check that the new Award models use `ExtensibleEnumT<>` for extensible enums
- Align ADR 0023's examples with the shipped identifier shape (fully qualified registry codes, `registry` with `code` and `url` only, current catalog slugs), from [this thread](https://github.com/HHS/simpler-grants-protocol/pull/1040#discussion_r3686075669)

The CLI spec update is tracked separately in #1052.

Issue #957, opened 2026-06-25
[Org IDs] Add identifiers models to organization and opportunity


widal001:
### Summary

Add the models for the identifiers collections to the `OrganizationBase` model (and possibly others). The old `ein`, `uei`, etc. models have been removed from `OrganizationBase`.

### Acceptance criteria

- [x] New `Identifier` and `IdentifierCollection` models have been added to the core library
- [x] The website docs have been updated to include these new models and

PR #1000, opened 2026-07-09
feat!(core): add identifier models to core library and website

widal001:
### Summary

Adds `Identifier` and `IdentifierCollection` models; replaces `ein`, `duns`, and `uei` on `OrganizationBase` with new identifier entries; and updates the website docs for these models.

- Fixes #957 
- Time to review: 10 minutes

- Adds `Identifier`, `IdentifierCollection`, and `SystemId` models to `@common-grants/core` library following the pattern defined in [ADR-0023](https://commongrants.org/governance/adr/0023-org-ids/)
- Adds `identifiers` prop to `OrganizationBase` and moves ID-related props into that collection, matching the org-related IDs added in #979 
  - `OrganizationBase.ein` -> `OrganizationBase.identifiers[org:us:ein]`
  - `OrganizationBase.uei` -> `OrganizationBase.identifiers[org:us:uei]`
  - `OrganizationBase.duns` -> `OrganizationBase.identifiers[org:xi:duns]`
- Adds a `/protocol/fields/identifiers/` page for the new identifier fields and updates `/protocol/models/organization` to include the new org ID entries
- Fixes the question bank and Form library mappings to use the new identifiers models.
- Updates `version-generator.ts` to remove props marked with `@Versioning.removed()` in the JSON schemas that are auto-generated from TypeSpec since `@typespec/versioning` doesn't support this natively, and commits these auto-generated schemas.
- Simplified the `Identifier` schema proposed in ADR-0023 by removing:
  - `kind`, `scope`, and `schema` from `Identifer.registry`
  - Removing `uri`, `verifiedAt`, and `verifiedBy` from the root

#### Instructions to review

1. Go to the PR preview page: https://cg-pr-1000.billy-daly.workers.dev/
2. Review the website protocol docs:
    - [`/protocol/fields/`](https://cg-pr-1000.billy-daly.workers.dev/protocol/fields/) should include Identifiers
    - `[/protocol/fields/idenfitiers/`](https://cg-pr-1000.billy-daly.workers.dev/protocol/fields/identifiers/) should include the newly added identifier TypeSpec models
    - [`/protcol/models/organization/`](https://cg-pr-1000.billy-daly.workers.dev/protocol/models/organization/) should have the new org ID models, an updated `OrganizationBase` and a new entry in the changelog
3. Review the updated question bank and form mapping:
    - Check that [the Organization EIN](https://cg-pr-1000.billy-daly.workers.dev/question-bank/org-ein/) form data is mapped to the correct location
[...]

PR #1040, opened 2026-07-24
fix(core): resolve v0.4.0 hold branch audit findings

SnowboardTechie:
### Summary

Audit pass over the v0.4.0 hold branch ahead of the release, fixing three inconsistencies it turned up:

- Adds the missing changeset for the org identifier models, which the 0.4.0 release notes would otherwise have shipped without.
- Uses the fully qualified registry codes (`org:us:ein`, not `us:ein`) in the org route docs and ADR-0026 examples.
- Aligns ADR-0026 and the org-syncing changeset with the shipped 0.4.0 response names and identifier shape.

- Addresses #1032
- Time to review: 5 minutes

**Add `.changeset/org-identifiers.md`** — the identifier work from #957 landed here without a changeset, so `@common-grants/core`'s 0.4.0 notes would have omitted `Identifier`, `SystemId`, `IdentifierCollection`, and `OrgIds`, along with both of its breaking changes: the removal of `OrganizationBase.ein`/`uei`/`duns`, and the tightened `employerTaxId`/`samUEI`/`duns` scalar formats. Minor, matching the other 0.4.0 changesets, so core still lands on 0.4.0.

**Fix registry codes** — codes follow `<schema>:<scope>:<prop>`, so the canonical forms are `org:us:ein`, `org:us:uei`, and `org:grants.gov:system`. The `registry` query param doc in `orgs.tsp` and the ADR-0026 examples still used short forms that contradicted `OrgIds` and `IdentifierT`. Catalog links are corrected alongside them, since slugs derive from the code (`org:us:ein` → `org-us-ein`) and `/registries/us-ein` and `/registries/grants-gov-org` pointed at pages that do not exist. The one-line change to `website/public/openapi/openapi.0.4.0.yaml` is the regenerated param description.

**Align ADR-0026 and the org-syncing changeset with the shipped names** — `.changeset/org-profile-syncing.md` called the 202 envelope `Responses.Accepted<T>`; the shipped names follow #1007's `T`-suffix convention (`AcceptedT<T>` plus the non-templated `Accepted`), and the org routes also return the new `Responses.Forbidden` alias. In ADR-0026, the contract summary used the pre-rename `Responses.Ok<T>`/`Responses.Paginated<T>` names, and the `GET /orgs/{orgId}` example gave `registry` objects `scope` and `kind` fields that don't exist on `IdentifierT.registry` (`code` and `url` only), with relative catalog URLs where the canonical `identifier.tsp` examples use absolute links.

Verified with `pnpm run ci` (exit 0; 40, 119, 548, and 71 tests across changelog-emitter, cli, sdk, and website).

The registry-code correction, against the canonical definitions in `organization.tsp`:

```
model OrgIdEin  is Fields.IdentifierT<Types.employerTaxId, "org:us:ein">;
model OrgIdUei  is Fields.IdentifierT<Types.samUEI,        "org:us:uei">;
model OrgIdDuns is Fields.IdentifierT<Types.duns,          "org:xi:duns">;
```

`pnpm changeset status` confirms the release still resolves to a single minor bump:

```
Packages to be bumped at minor:
- @common-grants/core
```

### DIFF
--- a/lib/core/lib/core/routes/orgs.tsp
+++ b/lib/core/lib/core/routes/orgs.tsp
@@ -56,7 +56,7 @@ interface Organizations {
   list<T extends Models.OrganizationBase = Models.OrganizationBase>(
     ...Pagination.PaginatedQueryParams,
 
-    /** Registry code of an external identifier to filter by (e.g. `us:ein`) */
+    /** Registry code of an external identifier to filter by (e.g. `org:us:ein`) */
     @query registry?: string,
 
     /** Identifier value within the given registry to filter by */
--- a/website/src/content/docs/governance/adr/0026-org-profile-syncing.md
+++ b/website/src/content/docs/governance/adr/0026-org-profile-syncing.md
@@ -79,7 +79,7 @@ Here is a brief summary of the API contract this ADR proposes:
 
 Every endpoint sits under `/common-grants/orgs`, and the path-based `{orgId}` refers to the organization's system-specific UUID (`Organization.id`). A client that only knows an external identifier, like an EIN, UEI, or platform ID, can look up the UUID via a filter query param on `GET /orgs` (see [ADR-0023](/governance/adr/0023-org-ids/)). The org record follows [`OrganizationBase`](/protocol/models/organization). Server-assigned fields like `datasetVersion` appear in response bodies but are ignored in request bodies, and since field-level schemas still need to be finalized in the follow-up spec, the payloads below are just illustrative.
 
-Successful responses use the standard CommonGrants envelope: `Responses.Ok<T>` wraps a single resource as `{ status, message, data }` and `Responses.Paginated<T>` wraps a list as `{ status, message, items, pagination }`, where the envelope's `status` is the HTTP status code. The read and list examples below show only the `data`/`items` payload; the write examples show the full envelope, since a write returns a change whose own lifecycle `status` (`accepted`, `pending`, and so on) sits inside `data`.
+Successful responses use the standard CommonGrants envelope: `Responses.OkT<T>` wraps a single resource as `{ status, message, data }` and `Responses.PaginatedT<T>` wraps a list as `{ status, message, items, pagination }`, where the envelope's `status` is the HTTP status code. The read and list examples below show only the `data`/`items` payload; the write examples show the full envelope, since a write returns a change whose own lifecycle `status` (`accepted`, `pending`, and so on) sits inside `data`.
 
 **Required endpoints**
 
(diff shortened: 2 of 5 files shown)


widal001 (approved):
These changes all looked solid to me! Thanks for catching these inconsistencies.

As we talked about in huddle today. A few other things we want to fix on this branch:

- Adding a `BooleanComparison` entry to the docs and linking to it from the custom filters entry in the Plugins (from #1045 )
- Double checking that we're using `ExtensibleEnumT<>` for any extensible enums in the new Award models
- Final step (should be its own issue) updating the CLI to include the new v0.4.0 spec (as a minor release) though this could technically wait until after we release v0.4.0, we'll just want to track it.

karinamzalez:
I wasn't familiar with [changesets](https://github.com/changesets/changesets) -- looks like it will write to CHANGELOG.mg once the release is cut and your changesets file will be deleted?

lib/core/lib/core/routes/orgs.tsp
```diff
@@ -56,7 +56,7 @@ interface Organizations {
   list<T extends Models.OrganizationBase = Models.OrganizationBase>(
     ...Pagination.PaginatedQueryParams,
 
-    /** Registry code of an external identifier to filter by (e.g. `us:ein`) */
+    /** Registry code of an external identifier to filter by (e.g. `org:us:ein`) */
```
karinamzalez, line 59: just out of curiosity -- why did we initially use this shorthand of the grant ID?

widal001:
> I wasn't familiar with [changesets](https://github.com/changesets/changesets) -- looks like it will write to CHANGELOG.mg once the release is cut and your changesets file will be deleted?

@karinamzalez Yes! You can test it by running `pnpm changeset version` from the repo root. But that won't happen until the `HOLD-core-v0.4.0` branch (the target for this PR) is merged into `main`

website/src/content/docs/governance/adr/0026-org-profile-syncing.md
```diff
@@ -155,19 +155,15 @@ Response:
...
       },
       "id": "01912a8b-7c3d-7890-abcd-ef1234567890"
     },
-    "us:ein": {
+    "org:us:ein": {
       "registry": {
-        "code": "us:ein",
-        "url": "/registries/us-ein",
-        "scope": "US",
-        "kind": "government"
```
karinamzalez, line 170: So, it looks like scope and kind are being removed from 0026's examples because they don't exist on `IdentifierT.registry` -- would we also want to remove them from ADR 0023?

karinamzalez (approved):
Approved with some questions!

lib/core/lib/core/routes/orgs.tsp
SnowboardTechie: ADR 0023 defined the codes as two parts, `us:ein`, and these route docs were written back when that was current. The identifier models that actually shipped in [#1000](https://github.com/HHS/simpler-grants-protocol/pull/1000) added a third segment naming the model the identifier belongs to (`org:us:ein`), and these docs just never got updated to match.

website/src/content/docs/governance/adr/0026-org-profile-syncing.md
SnowboardTechie: I think so, but not in this PR. 0023 is out of date in more ways than `scope`/`kind` (two-part codes throughout, a `schema` field on the registry object, old catalog slugs), so a spot-fix would still leave it stale. We'll fold the 0023 cleanup into the follow-up PR for Billy's feedback.

lib/core/lib/core/routes/orgs.tsp
karinamzalez: Ah, that makes more sense -- thank you!

PR #1044, opened 2026-07-28
docs(sdk): document custom filters and correct the SDK READMEs

### LINKED ISSUE #1031: [SDK] Document custom filter behavior in the extensions READMEs
### Summary

Add documentation of custom filter behavior to the extensions README in both the Python and TypeScript SDK subdirectories, so developers implementing custom filters have a reference beyond the source code.

### Acceptance criteria

- [x] The TypeScript SDK extensions README documents custom filter behavior
- [x] The Python SDK extensions README documents custom filter behavior
- [x] Root level READMEs are up to date

SnowboardTechie:
### Summary

- Fixes #1031
- Time to review: 20 minutes

### Changes proposed

- Document custom filters in the Python SDK extensions README, which had no mention of filters at all — the source was the only reference.
- Fix the TypeScript SDK's package README kitchen sink, which did not compile, and refresh the Python one, which never showed the plugin surface. These are the examples rendered on npm and PyPI.
- Correct four snippets and references that fail when actually run, found by executing every code block rather than reading it.
- Document the custom-filter and transform examples in both `examples/` READMEs.
- Add patch changesets for `@common-grants/sdk` and `common-grants-sdk`.

### Context for reviewers

**The Python and TypeScript filter surfaces are not translations of each other**, so the new Python section is not a port of the existing TypeScript one. TypeScript declares filters as *data* (`filterType: "stringArray"` in a record); Python declares them as *types* (`region: StringArray`, where the annotation is the Pydantic value model, recovered at runtime via `get_type_hints`). That difference cascades into the docs: TypeScript needs its `as const` trap section and Python needs none, but Python needs its own trap note, because `f.*` builders fall through to `DefaultFilter` for `Money`/date values and `in_`/`not_in` carry reserved-word underscores. I verified all ten operator sets directly against the Pydantic models rather than copying the TypeScript table; they match exactly.

**Bugs this fixes.** Each was reproduced, not inferred:

| Where | Problem |
| --- | --- |
| `lib/ts-sdk/README.md` | Kitchen sink did not compile. `definePlugin({ extensions: … })` is not a valid key (`TS2353`); custom fields live under `schemas.<Model>.customFields`. And `plugin.schemas.Opportunity` was passed where a Zod schema was expected (`TS2740`) — that value is a compiled entry object whose schema is at `.commonSchema`. |
| `lib/python-sdk/common_grants_sdk/extensions/README.md` | `Client(base_url=…)` raises `TypeError`; the constructor takes `config=Config(…)`. The reusable-plugin snippet imported `Opportunity` from `schemas.pydantic.models`, which raises `ImportError`, while using `OpportunityBase` in the body. |
| `lib/ts-sdk/src/extensions/README.md` and `types.ts` | `CustomFilterType` described as an 11-value enum; it has held 10 since `integerComparison` was dropped in #1011. The README's own table already listed 10. |
| `lib/ts-sdk/examples/README.md` | Described `plugins.ts` as composing plugins with `mergeExtensions()`, which is no longer in the SDK, and showed a "Composed plugins" output section the script does not print. |
| `lib/python-sdk/examples/README.md` | Carried the TypeScript config table verbatim: documented `CG_BASE_URL` where `Config` reads `CG_API_BASE_URL`, claimed a `localhost:8000` default where `Config` raises `ValueError`, and told readers to run a `pnpm` command. Also pointed at a `lib/py-sdk` directory that does not exist. |
| `lib/python-sdk/examples/typed_custom_filters_failures.py` | The docstring told reviewers to run pyright against the file directly, but the file sits in `pyrightconfig.json`'s `exclude` (deliberately — its errors are intentional) and pyright skips excluded files even when named on the command line, so the command analyzed 0 files. The docstring and the examples README now point to `make check-fixtures`, which checks the file through its own pyright config (`pyrightconfig.fixtures.json`) and passes only when exactly one error per `# EXPECT-ERROR` marker fires; it runs as part of `make checks`. |
| `lib/ts-sdk/src/extensions/README.md` | Four API-table links pointed at a nonexistent `./transformation.ts`; the symbols live in `../utils/transformation.ts`. |

The Python package README's kitchen sink was **not** broken — it runs as written. It was stale: it taught the ad-hoc `with_custom_fields()` path and never showed `define_plugin`, `get_client()`, or filters. It now mirrors the TypeScript example so both landing pages teach the same thing.

**Verification.** Every code block was executed or typechecked, not eyeballed — that is what surfaced the bugs above.

- Both `ts` blocks in `lib/ts-sdk/README.md` through `tsc --noEmit --strict`: 0 errors, down from 2.
- Both `python` blocks in `lib/python-sdk/README.md`, and the new filters section and rewritten client section blocks in the extensions README, executed with the HTTP layer stubbed: all clean. The classifier output matched the documented three-bucket split exactly.
- All 22 `python` blocks in the extensions README: 0 syntax errors.
- 116 internal markdown anchors across the six edited READMEs: 0 broken. All added relative file links resolve.
- `make checks`, `make test` (396 passed), `make build` for the Python SDK.
- TypeScript SDK `check:lint`, `check:types`, `build`, `test`, `test:coverage` (80% threshold), `example:transforms`, and `prettier --check` across all 85 tracked files.

One pre-existing failure is untouched by this branch: `pnpm run ci:sdk` exits 1 only because `prettier --check .` flags `lib/ts-sdk/AGENTS.md`, an untracked, globally gitignored file absent from any CI checkout; every tracked file passes. (The `brace-expansion` audit advisory previously noted here was fixed on `main` by #1041 and is resolved by this branch's merge from `main`.)

### Additional information

The new Python section's classifier example, run against the real SDK, produces exactly the three buckets it documents:

```json
{
  "status": { "operator": "in", "value": ["open", "forecasted"] },
  "closeDateRange": { "operator": "between", "value": { "min": "2025-01-01", "max": "2025-12-31" } },
  "customFilters": {
    "agency": { "operator": "in", "value": ["HHS", "DOE"] },
    "fundingProgram": { "operator": "like", "value": "*Conservation*" },
    "legacyTag": { "operator": "eq", "value": "conservation-2024" }
  }
}
```

`status` and `closeDateRange` land as named top-level fields; the registered (`agency`, `fundingProgram`) and ad-hoc (`legacyTag`) keys land under `customFilters`.


karinamzalez (commented):
I have a few questions! 
- Since the README code snippets will re-drift every API change -- would it be worth it to set it up the blocks so that they're executable in CI? Maybe we could create a follow up ticket?
- A thought on the pyright-exclude step: verifying typed_custom_filters_failures.py means removing it from exclude in pyrightconfig.json first, since pyright skips excluded files even when you name them directly. Totally works! I'm just wondering if a dedicated config or check target for that file might make the negative fixtures easier for the next person to verify? 
- A question on the changesets: both are patch, and since the only non-doc change is a comment fix in types.ts, this reads as docs-only to me. Since these READMEs render on npm/PyPI, is the goal to publish a release so those pages refresh, or would we rather this land without one? Still learning the release cadence here, so genuinely curious which we'd prefer.
- A question on the verification: the description lists a lot of great checks-- all the ts/python blocks executed, 396 Python tests, coverage, etc. Is any of that captured somewhere a future maintainer could re-run, like a script or make target, or was it a one-time pass? Even a short note on how you executed the blocks could be helpful for incoming folk.

.changeset/py-sdk-custom-filter-docs.md
```diff
@@ -0,0 +1,9 @@
+---
+"common-grants-sdk": patch
+---
+
+Document custom filters in the extensions README, and correct two broken snippets plus the package README's kitchen sink example.
```
widal001, line 5: @SnowboardTechie IMO just this snippet is sufficient for our changelog. The rest of the content below feels more appropriate for a PR description or a commit message than a changelog.

widal001: Especially since most of the changes are to nested READMEs, and the only content that is directly displayed on PyPI and npm is the root-level README.

.changeset/ts-sdk-custom-filter-docs.md
```diff
@@ -0,0 +1,9 @@
+---
+"@common-grants/sdk": patch
+---
+
+Correct the package README's kitchen sink example, which did not compile, and fix a stale filter-type count in the extensions docs.
```
widal001, line 5: Same thing for this changeset, I don't know if we need the detailed description below. SDK consumers don't really need in the weeds descriptions on how are docs were out of date in the changelog.

lib/python-sdk/README.md
widal001, line 1: It's not one of the changed lines in here, but something I discovered when I was testing the current [README on PyPI](https://pypi.org/project/common-grants-sdk/) is that the relative links to the nested sub-directories currently 404, because PyPI doesn't seem to resolve relative links back to the root repo the way that npm does.

Can you see if there's a way to preserve the relative links but have them point back to the repo sections or if we need to use absolute refs?

lib/ts-sdk/src/extensions/README.md
```diff
@@ -743,7 +743,7 @@ F.between("2025-01-01", "2025-12-31"); // { operator: "between", value: { min: "
 // Full set: eq, neq, gt, gte, lt, lte, in, notIn, like, notLike, between, outside
 ```
 
-> **Cross-SDK note.** TypeScript uses `F.in` (`"in"` as an object key — valid JS). The Python sibling SDK ([#869](https://github.com/HHS/simpler-grants-protocol/issues/869)) uses `f.in_` (trailing underscore, Python convention for reserved words). This is a documented naming divergence across SDKs.
+> **Cross-SDK note.** TypeScript uses `F.in` (`"in"` as an object key — valid JS). The Python SDK uses `f.in_` (trailing underscore, Python convention for reserved words), and declares filters by _annotating_ a TypedDict subclass rather than by writing a `filterType` string. See the [Python extensions README](../../../python-sdk/common_grants_sdk/extensions/README.md#plugin-custom-filters). These are documented naming and authoring-shape divergences; the wire contract is identical.
```
widal001: I'm not really sure this is a necessary callout, especially on the TypeScript SDK. TS SDK consumers don't really need to know about how the equivalent functionality is implemented in Python.

lib/python-sdk/common_grants_sdk/extensions/README.md
```diff
@@ -462,23 +474,215 @@ if result.errors:
...
+To build a request body without a client, call `classify_filters` directly:
+
+```python
+from common_grants_sdk.extensions import classify_filters
+
+body = classify_filters(
+    grants_gov_plugin.routes, "opportunities", "search", {"agency": f.in_(["HHS"])}
+)
+body.model_dump(by_alias=True, exclude_none=True, mode="json")
+```
```
widal001: This README should be documenting the public interface for custom fields, filters, and plugins. We don't really want to be encouraging SDK consumers to be calling the `classify_filters()` or `validate_routes()` directly so references to these largely internal functions be removed from this README, especially if the goal is to eventually make these private methods.

lib/python-sdk/common_grants_sdk/extensions/README.md
```diff
@@ -462,23 +474,215 @@ if result.errors:
...
+
+`mode="json"` is required — coerced `date` objects are not JSON-serializable in Pydantic's default python mode.
+
+Registered filters are looked up by the **exact** `(resource, method)` strings the plugin declared. A non-matching pair (a pluralization typo, say) yields no registered bucket at all, and every non-standard filter is then validated as if it were ad-hoc.
+
+### Validation — registration time and call time
+
+`validate_routes(...)` runs when the plugin is defined, and again when `plugin.get_client()` binds routes, so an authoring mistake surfaces for the author rather than for a consumer later. It raises `FilterError` if:
```
widal001: Same note as above. While I like the idea of describing what we're validating, we mostly want to keep the descriptions focused on the public interfaces, behaviors, and return types, not the internal functions and implementation details.

widal001 (commented):
@SnowboardTechie Thanks for making all of these updates!

Just a few things I'd like to address before we merge this and kick off the patch release:

- **Must do:** Fix the relative links on the Python README (see my comment below about how PyPI resolves relative links)
- **Should do:** Remove some of the in-the-weeds implementation details from the changesets and docs. We want these docs to mostly be focused on the information that SDK consumers need to know

SnowboardTechie:
@karinamzalez answers in order:

**Executable snippets in CI:** There's real value here: a snippet runner would have caught 4 of the 7 bugs this PR fixes (the other 3 were prose drift). But I piloted it against these READMEs, and only 9 of the 52 blocks run standalone. The examples intentionally build on each other (define the plugin once, use it in the next ten blocks), so making them CI-executable means changing how we author the docs, not just adding a harness; the tooling side is cheap. So not this PR and not a quick ticket, but I think it's worth a team discussion about whether to move the docs in that direction over time.

**The pyright-exclude step:** Great idea! The fixtures file now has its own config (`pyrightconfig.fixtures.json`) and a `make check-fixtures` target that passes only when pyright reports exactly one error per `# EXPECT-ERROR` marker, so a guard that stops firing and a new unintended error both fail it. It runs as part of `make checks` (d814c302).

**Changesets:** Intentional, the goal is a release. The package READMEs render on npm and PyPI, and only publishing a new version refreshes those pages.

**Verification:** The fixtures target above plus the existing package gates (`make checks && make test`, `pnpm run ci:sdk`) are what live in the repo. The block-by-block pass was one-time, so here's the how for whoever does this next: extract each fenced block to a file with imports hoisted, run TypeScript blocks through `tsc --noEmit --strict` with paths mapped to the SDK source, and run Python blocks with the client's HTTP transport monkeypatched to canned responses. Context-dependent blocks need the earlier blocks' setup pasted in front, which is most of the manual labor and the same self-containment issue from the first answer.

.changeset/py-sdk-custom-filter-docs.md
SnowboardTechie: Trimmed both changesets to their summary lines in dc29cc10.

.changeset/ts-sdk-custom-filter-docs.md
SnowboardTechie: Same, trimmed in dc29cc10.

lib/python-sdk/README.md
SnowboardTechie: There isn't a way to keep them relative: npm rewrites relative links against the `repository` field, but PyPI's renderer doesn't resolve them at all, and poetry-core has no build-time rewrite hook (hatch has one, but that's a backend swap for six links). They're absolute GitHub URLs pinned to `main` now (dc29cc10).

lib/ts-sdk/src/extensions/README.md
SnowboardTechie: Removed, and I dropped the mirror-image TypeScript mention from the Python naming note for the same reason.

lib/python-sdk/common_grants_sdk/extensions/README.md
SnowboardTechie: Pulled these out in dc29cc10: the direct-call snippets are gone, classification and validation are described as client behavior (`search()` / `get_client()`), and the glossary rows are removed. The TypeScript extensions README had the same shape (`classifyFilters` / `validateRoutes` snippets and API-table rows), so I gave it the same treatment.

lib/python-sdk/common_grants_sdk/extensions/README.md
SnowboardTechie: Covered by the same pass.

lib/python-sdk/README.md
widal001: Cool thanks for doing some additional research, and good thinking with the fully resolved URLs.

widal001 (approved):
Thanks for the changes you made!

PR #1045, opened 2026-07-28
docs(website): add custom filters section to the per-plugin details page

### LINKED ISSUE #1030: [Website] Add custom filters section to the per-plugin details page
### Summary

The per-plugin details page on [commongrants.org/plugins](https://commongrants.org/plugins/) already has a custom fields section; add an equivalent section for custom filters so users can see which filters each plugin supports.

### Acceptance criteria

- [x] The per-plugin details page includes a custom filters section listing the filters that plugin supports
- [x] The section follows the pattern of the existing custom fields section

SnowboardTechie:
### Summary

- Fixes #1030
- Time to review: 10 minutes

### Changes proposed

- The per-plugin details page now has a **Custom Filters** section listing the filters each plugin accepts, so a reader can see a plugin's filter support without reading its source.
- Filter declarations are hand-maintained in `website/src/content/plugins/index.json` alongside the existing `fields`, shaped resource → method → name. That is a catalog shape and deliberately matches neither SDK: the TS SDK wraps the filter map in a `filters` key, and the Python SDK puts a TypedDict class in the method slot.
- Each card links to the protocol definition for its filter type.
- The section mirrors the existing Custom Fields section — same `CardGrid`, same `CatalogCard`, same badge and pill styling.

### Context for reviewers

**What the section shows.** Both Grants.gov plugins declare the same four filters on `opportunities.search`: `agency`, `applicantType`, `fundingInstrument` (`stringArray`), and `costSharing` (`booleanComparison`). Each card carries the filter name, a description, a filter-type badge, and an `opportunities.search` route pill. Operators aren't shown — that table lives in the SDK, which the website doesn't depend on, so duplicating it here would drift.

**Where the filter data comes from.** Hand-maintained, the same as the existing `fields` catalog. The plugins declare `filterType` only, so the four descriptions are authored here. No change was needed to `fetch-plugin-metadata.ts` — it spreads the whole source entry, so the new key flows into the build cache; that was confirmed by running `pnpm generate:plugin-metadata` rather than assumed.

**How this was verified.**

- `pnpm run ci` in `website/` (build + lint + format + types + astro check + cspell + vitest) — passing.
- Every anchored href is checked against the real headings in the filters docs by a test, not by hand: nine anchors, none dead. `starlight-links-validator` does **not** cover these — it parses markdown, and the plugin pages are `.astro`, so a build with a deliberately dead anchor still reports all internal links valid. The test is the only guard.
- The new tests were mutation-checked to confirm they fail when the mapping breaks — a fabricated boolean anchor fails two cases, and a typo'd page segment fails one.
- Visual parity with the Custom Fields section was checked by comparing the two cards' emitted scoped CSS in the built page: the badge, pill, and pill-list rule bodies are identical.

**Testing instructions.**

```bash
cd website && pnpm dev
# then open /plugins/ts-cg-grants-gov/ and /plugins/cg-grants-gov/
```

**No changeset.** `.changeset/config.json` scopes releases to `lib/python-sdk`, `lib/core`, `lib/cli`, and `lib/ts-sdk`; `website` is not a released package.

### Additional information

No screenshot attached. The rendered section is reproducible with the dev-server steps above; the four cards appear below Custom Fields in the same grid.

### DIFF
--- a/website/__tests__/lib/plugins/filter-docs.spec.ts
+++ b/website/__tests__/lib/plugins/filter-docs.spec.ts
@@ -5,12 +5,6 @@ import { filterDocsHref, isCustomFilterType } from "@/lib/plugins/filter-docs";
 import { Paths } from "@/lib/schema/paths";
 import type { CustomFilterType } from "@/lib/plugins/types";
 
-/**
- * Pages that exist under src/content/docs/protocol/filters/. A filter type may
- * only link to one of these, or to the index.
- */
-const FILTER_DOC_PAGES = ["string", "numeric", "date", "money"];
-
 const ALL_FILTER_TYPES: CustomFilterType[] = [
   "stringComparison",
   "stringArray",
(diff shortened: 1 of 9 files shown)


widal001 (approved):
Looks great! 

A couple future changes we should keep track of:

- **Fix:** On #1040 can we make sure there's an entry for BooleanComparison and link to it from the cost sharing filter?
- **Idea:** As we start to publish more plugins we might want to create a custom filters catalog the same way we do for custom fields, that would show which plugins support the same custom filters.
- **Idea:** As the number of custom fields on a plugin increases, we might want to add a carousel component instead of just listing them all on the page.

Screenshot 2026-07-30 at 1 59 26 PM

Issue #1034, opened 2026-07-22
[Mock API] SPIKE: Mock API playground implementation


karinamzalez:
# Spike: Evaluate mock API playground approach (browsable by version)

> [!IMPORTANT]
> **Updates since this ticket was written (2026-08-04):**
>
> - **Test A is done**: POC [#1049](https://github.com/HHS/simpler-grants-protocol/pull/1049) proved client-side MSW works end-to-end. The review blocked on response fidelity; that's now addressed with a frozen fixture + deterministic hand-authored handlers (repeatable, cross-endpoint-consistent, filters/sorting/errors).
> - **The goal was reframed** by the PR review and onboarding/quickstart discussions: browser,CLI and SDK continuity (same URL, same data everywhere) was promoted from nice-to-have to core, and the mock should double as the SDK quickstart sandbox.
> - **The ADR's recommendation shifted** from client-side MSW to a **small standalone Cloudflare Worker** serving one real mock URL: the docs site stays pure static on GH Pages, untouched. The #1049 fixture/handlers carry forward as the Worker's data layer.
> - **Test B is likely moot** under that direction (no service worker ships to GH Pages), pending the team ratifying the ADR.
 
## Goal
 
Determine the best technical approach for letting users test CommonGrants API endpoints against a mock server directly in the browser, with responses browsable by spec version (v1, v2, v3…). This spike is investigation only-- the output is a recommendation and an ADR.
 
## Questions to answer
 
- **Client-side vs. serverless?** Can we mock endpoints entirely in the browser, or do we need a hosted/serverless component? Prism is the leading candidate-- confirm whether it runs cleanly client-side (JS/WASM) or requires a Node server.
- **Can the playground and sandbox be unified?** If a client-side mock can serve both the "explore the spec" and "test against behavior" use cases, they may be the same feature. Determine whether that's viable or whether they remain distinct.

## Considerations & constraints
 
- **Stay within static assets.** The solution should not veer away from the static-site model (Astro/Starlight). A static, client-side approach is strongly preferred.
- **Budget.** If a separately hosted server is required, factor in hosting cost and ongoing maintenance — this materially changes scope and should be flagged before committing.
- **Versioning.** Whatever approach we pick must support browsing/mocking multiple spec versions (v1, v2, v3…).

## Acceptance Criteria
[...]

PR #1049, opened 2026-07-30
[POC] #1034 Test A: client-side MSW mock playground

karinamzalez:
### Summary

- Refs #1034 (spike **Test A** — proof-of-concept; not intended to merge as-is)
- Addresses @widal001's review on this PR blocking the spike on response fidelity-- all five items are now met; see the review reply below for the item-by-item mapping.
- Time to review: ~20 minutes

> [!IMPORTANT]
> **This PR's role has changed since it was opened.** The review feedback here, plus the onboarding/quickstart discussion, shifted the spike ADR's recommendation away from client-side MSW (this PR) toward a **small standalone Cloudflare Worker** serving the mock: one real URL that answers the browser, `curl`, and the SDK identically, while the docs site stays pure static on GH Pages, untouched. This PR now serves two purposes:
>
> 1. **Evidence:** it proves the client-side path works end-to-end (Test A), and documents why spec-generated responses aren't enough on their own.
> 2. **The portable data layer:** the fixture and handlers built here are what the recommended option would reuse. Of `handlers.ts`'s ~580 lines, MSW touches six (the import, two `HttpResponse.json` wrappers, three route registrations)-- the filter/sort/error logic is pure request->response code that ports into a Worker nearly as-is.
>
> The browser-vs-CLI/SDK trade-off is deliberately **not** resolved in this PR -- that's the ADR's central question, so please put architecture-level thoughts there. Review here is scoped to: are the fidelity must-dos/should-dos met?
>
> **Expected end state:** if the team approves the standalone-Worker direction, this PR closes unmerged with a pointer to the ADR and the successor PR; the fixture/handler modules carry forward. 

- Add MSW + `@mswjs/source` and initialize the `mockServiceWorker.js`; add a throwaway `/protocol/mock-playground` route rendering the shared `OpenApiDocs` island with "Try it out" enabled, generating handlers from the rendered OpenAPI specs via `fromOpenApi()` and swapping the active set across v0.1.0/v0.2.0/v0.3.0 (the shipped `/protocol/api-docs` page still ships with Try-it-out disabled).
- **New** `opportunities/fixtures.ts`: 11 semantic, version-aware opportunity records built from the TypeSpec `@example` decorators instead of faker output. One reproduces the spec's published example under the id Swagger UI pre-fills, so the default Execute returns that record rather than a 404, and it sorts first in the list response.
- **New** `opportunities/handlers.ts`: hand-authored list/detail/search reading from that fixture: deterministic, mutually consistent, with real `status`/date/money-range filters and `OppSortBy` sorting. Bad filter bounds and out-of-range pagination are rejected rather than silently ignored.
- **New** error affordances, listed on the playground page so they're discoverable: 404 for an unknown UUID; 400 for a malformed id, an unknown `sortBy`/operator, a malformed filter bound or JSON body, and `page`/`pageSize` below the spec's `minimum: 1`-- all in the protocol's `{ status, message, errors[] }` envelope.
- `MockPlayground` prepends the opportunity handlers ahead of the generated set (MSW resolves first-match-wins); `spec-handlers.ts` memoizes the rest, so repeat calls are identical across the whole surface for the session.
- Tests across all three protocol versions: determinism, list -> detail consistency, filters, sorting, pagination, error shapes, and handler precedence.

**Root cause of the fidelity gap.** `fromOpenApi()` falls back to `seedSchema()`, which calls `@faker-js/faker` with no `faker.seed()` -- fresh data every call and ignores OpenAPI's singular `example`, the keyword our `@example` decorators compile to. Structural, not configuration: schema-seeding can't resolve a path param, keep two endpoints consistent, apply filters, or emit errors. (This applies to any schema-sampling mock, including Prism's dynamic mode, which is why the fixture layer is needed under every option, not just MSW.)

**Fix: a hybrid.** Fixture-backed handlers for the three opportunity endpoints, layered ahead of the generated set that still covers every other path. Rejected: seeding faker (repeatable, but nothing else); adding response-level examples to the TypeSpec (edits the published protocol to serve a docs feature); hand-authoring everything (loses the breadth that makes the playground worth visiting).

**How to verify** -- open `/protocol/mock-playground`; the intro block lists the error inputs.

- `GET /opportunities` -> Execute twice -> same body. First item is `Small business grant program`.
- `GET /opportunities/{oppId}` -> Execute **without touching the pre-filled id** → 200, matching the "Example Value" pane above it. Any `id` from the list resolves to that same record.
- Search with `status in [open]` -> only `open` records; flip `sortOrder` → exact reverse.
- Version dropdown -> v0.1 omits `competitions` and `acceptedApplicantTypes`.
- Any other endpoint -> still responds; Execute twice → identical.

**Verified locally:** 146 tests pass, `pnpm --filter website run checks` clean, plus the flow above against a dev server.

**Trade-offs.** The fixture is hand-maintained, so it can drift from the TypeSpec models -- tests catch gross shape drift, not full conformance. Two of the 400s aren't declared responses on their operations, so Swagger UI labels them "Undocumented"; either the mock is over-strict or the protocol under-specifies param validation. Long-tail memoization is session-scoped and resets on reload.


karinamzalez:
✅ **Verified on the Cloudflare preview** (`cg-pr-1049.billy-daly.workers.dev`): the MSW service worker registers and Swagger UI **Try it out → Execute** returns `200 OK (from ServiceWorker)` with schema-valid bodies for `GET /common-grants/opportunities` across **all three versions** (v0.1.0 / v0.2.0 / v0.3.0), the version dropdown swapping the active handler set.

**Base-URL finding:** specs have no `servers:` block → `@mswjs/source` emits relative same-origin handlers matching exactly what Swagger UI targets — no base-URL patching needed.

**Known caveat (fix before productionizing):** the root-scoped worker intercepts the page's own navigation requests; a navigation hitting Astro's trailing-slash **307** logs `Failed to fetch` in MSW's `passthrough` (can't re-fetch a `navigate`-mode request). Blast radius is narrow — MSW bypasses when no client has started the worker, so only the active playground page is affected and the canonical trailing-slash URL is clean. Fix: bypass `navigate`-mode requests and/or unregister the worker on leaving the playground.

_CI's website `validate` is red only at the pre-existing repo-wide dependency **audit** (`--level high`); build/checks/tests pass and the preview deploys regardless._

karinamzalez:
Screenshot 2026-07-29 at 11 04 14 PM
Screenshot 2026-07-29 at 11 05 36 PM
Screenshot 2026-07-29 at 11 06 05 PM

widal001:
@karinamzalez Thanks for all of your work on this! It's very cool to see it live.

While this an excellent proof of concept, there are a few key things I'd love to iron out in this test before we can consider MSW a viable path forward:

#### Must do

1. **Deterministic responses:** Repeated calls to the same API endpoint, with the same inputs return the same value. Right now multiple calls to the same API endpoint e.g. `GET /common-grants/opportunities/30a12e5e-5940-4c08-921c-17a8960fcf4b` return different response bodies each time, including an `id` value that doesn't match the path param.
2. **Consistent across API endpoints:** If users call `GET /common-grants/opportunities/` and then `GET /common-grants/opportunities/{oppId}` with the ID from the list response, they should see the same data across those two response calls.
3. **Semantically representative responses:** In addition to being consistent across multiple calls to the same endpoint, the data returned should also be semantically representative of the data returned by an actual endpoint, so that visitors can build some intuition around the expected values per field in addition to their basic shape.

#### Should do

1. **Valid filters and sorting:** If users include sorting and filtering in the request, they should get a difference response than without those inputs. And the responses returned shouldn't contradict the filter/sorting params included (e.g. a status that's not in the list of statuses passed to the request)
2. **Option to test errors:** There should be a way to surface non-2xx responses, like missing opportunity IDs (404) and validation errors, ideally ones that would match the error structure that we outline.

karinamzalez:
@widal001 Thanks for the review! All five items are addressed on this branch-- verifiable on the preview at `/protocol/mock-playground` (the intro block lists the error inputs):

1. **Deterministic** ✅ Opportunity endpoints read from a frozen fixture, so repeat calls are byte-identical and detail echoes the requested id. Your example id (`30a12e5e-…`) now returns the spec's own published example record. Everything else is memoized per session.
2. **Consistent across endpoints** ✅  List, detail, and search project from one fixture; any id from the list resolves through detail to the same record.
3. **Semantically representative** ✅  11 records built from the TypeSpec `@example` decorators (not faker), spanning all four `OppStatus` values, version-aware.
4. **Filters/sorting** ✅ `status`, `closeDateRange`, and funding-range filters apply; sorting is stable and reversible; malformed bounds are a 400 rather than silently dropped.
5. **Errors** ✅ Documented affordances return the protocol `{ status, message, errors[] }` envelope: unknown UUID -> 404; malformed id/sort/operator/filter/body/pagination -> 400.

**Why it took a fixture:** `fromOpenApi()` generates from the schema via unseeded faker and ignores the `example` keyword our `@example` decorators compile to-- so no configuration could meet items 1–3. That's true of any schema-sampling engine (including Prism's dynamic mode), which means this data layer is needed under *every* option-- i.e. it's portable, not MSW-specific patching :) 

**Bigger implication:** your must-dos plus the browser / curl / SDK continuity point shifted the ADR's recommendation to a small standalone Worker serving one real URL for all three consumers (docs site stays static on GH Pages), reusing this PR's fixture/handlers. Are you open to taking the architecture discussion to the ADR? I'm still making small updates but here is the link: https://drive.google.com/file/d/1EqvksgT5zesQelSGvv0XrujDagxDJA4U/view?usp=drive_link

Issue #947, opened 2026-06-24
[Stewardship] Build and publish the CSG participants page on commongrants.org


laurabelinfante:
### Summary

Build and publish a public participants page on the CommonGrants website listing the organizations and individuals in the Community Stewardship Group and their roles. Closes AC #4 on epic #551 and the stewardship group & participants page AC on deliverable HHS/simpler-grants-gov#8757.

### Acceptance criteria

- [ ] A participants page is published on commongrants.org listing participating organizations
- [ ] The page is linked from the site navigation

PR #1051, opened 2026-07-30
ci: adopt release-please for monorepo releases

### LINKED ISSUE #870: [Release] Adopt release-please for monorepo releases
### Summary

Adopt release-please in manifest mode for the four published packages (`lib/core`, `lib/cli`, `lib/ts-sdk`, `lib/python-sdk`) so the conventional commit history drives releases instead of hand-authored changeset files.

### Rationale

The current Changesets setup mostly works, but it has a couple of challenges that release-please would address:

- The team has to remember to create changesets to initiate a version bump
- The team can't easily stack changes before a release goes out, because the current workflow bumps version as soon as one changeset is present
- The team has to remember to manually run the release and deploy after a version bump is created
- The current CI requires disabling branch protections to support directly committing to `main`
- The python codebase requires a bunch of workarounds like `package.json` and a special script to bump the python version

The benefits of switching:

- The conventional commit history is the changeset, so contributors don't have to remember a separate file
- Multiple changes can be documented in a single Release PR and only ship when that PR is merged, so the team controls release timing without disabling branch protections
- Python releases use the real `pyproject.toml` config with `release-type: python`, no vestigial `package.json` or shell-parsing step
- Release notes are generated from PR titles, so they read more like a changelog and less like a list of changeset summaries

There are a few tradeoffs though:

- Requires switching to conventional commits and adopting a different PR title format
- Requires contributors to know which conventional commits map to which kinds of version bumps

### Implementation notes

The CommonBenefits config provides a decent template for what this repo would need to make this change (four packages, separate release PRs per package, conventional commits, component-aware tags):

- [`release-please-config.json`](https://github.com/common-benefits/common-benefits/blob/main/release-please-config.json) — manifest mode with `separate-pull-requests: true`, `include-component-in-tag: true`, `bump-minor-pre-major: true`, `tag-separator: "@"`
- [`.release-please-manifest.json`](https://github.com/common-benefits/common-benefits/blob/main/.release-please-manifest.json) — version tracking, one entry per package
- [`.github/workflows/release-please.yml`](https://github.com/common-benefits/common-benefits/blob/main/.github/workflows/release-please.yml) — uses `googleapis/release-please-action@v4` and fans `released` + `tag` outputs out to the per-package publish workflows
- [`DEVELOPMENT.md`](https://github.com/common-benefits/common-benefits/blob/main/DEVELOPMENT.md) — the contributor-facing description of the flow

Files that would need to be removed or updated:

- The `.changeset/` directory and `.changeset/config.json`
- [`lib/python-sdk/package.json`](https://github.com/HHS/simpler-grants-protocol/blob/main/lib/python-sdk/package.json) (the vestigial one)
- The "Detect changesets", "Bump version for Python package", "Bump version for Node package", and tag-creation steps in [`ci-bump-version.yml`](https://github.com/HHS/simpler-grants-protocol/blob/main/.github/workflows/ci-bump-version.yml), replaced by a single `release-please-action` step
- The per-package publish workflows would key off the release-please `released` outputs instead of changeset-created tags

### Open questions

- Can we make release please work with our existing changelogs rather than creating new changelogs?
- Does release please support PR filtering in the GitHub release notes it creates?

### Acceptance criteria

- [x] A conventional commit on `main` opens a separate Version PR per affected package without any manual changeset step
- [x] Merging a Version PR tags and publishes the right package(s) to npm, and the Python SDK to PyPI
- [x] `lib/python-sdk/package.json` is gone and the Python version lives only in `pyproject.toml`
- [x] ADR 10 (or its successor) is updated to reflect the choice
- [x] Branch protections have been re-enabled

SnowboardTechie:
### Summary

- Fixes #870
- Time to review: 30 minutes

### Changes proposed

Replaces the Changesets release workflow with [release-please](https://github.com/googleapis/release-please) in manifest mode, so the conventional-commit history on `main` drives versioning and releases for the four published packages.

**Why:** a forgotten `.changeset/*.md` ships a change unversioned with nothing to catch it; versions bump the moment any changeset lands, so merges can't be batched into one release; the bump commit is pushed straight to `main` by CI, which is why both `main` protection rulesets are off today; and Changesets is Node-only, so the Python SDK carries a stub `package.json` plus a shell script that greps changeset files to drive the Poetry bump. The options weighed, including keeping Changesets with its official Version-PR action, are recorded in the new ADR.

**What changed:**

- **Added** `release-please-config.json` + `.release-please-manifest.json` (seeded with the current versions) and a `release-please.yml` workflow that maintains one Release PR per package and fans out to the publish workflows when a Release PR merges. The existing `name@version` tag format is preserved exactly.
- **Rewired** the four `cd-deploy-lib-*.yml` publish workflows: the release-creation job is gone (release-please owns tag + GitHub release + notes) and the deploy job is now `workflow_call` from `release-please.yml`, keeping `workflow_dispatch` as a manual fallback for re-running a publish against an existing tag. A prefix guard on `release_tag` protects the manual path against a mistyped ref (replacing the validation the deleted release-notes script used to do). Each publish call receives only the one registry token it uses, declared in the called workflow's `workflow_call.secrets`, rather than inheriting every repo secret; the deploy jobs also drop an unused `id-token: write` permission.
- **Removed** the Changesets machinery: `.changeset/`, `ci-bump-version.yml` (including its direct-to-`main` bump commit that required branch protections off), the release-notes script + `.github/release-config/`, the vestigial `lib/python-sdk/package.json`, the root `@changesets/*` devDependencies, and the `tomlq`/`tomlkit` dev-deps that only served the deleted bump script. The Python version is now edited only in `pyproject.toml` (`release-type: python` updates it natively).
- **Fixed** `common_grants_sdk/__init__.py`, which was stuck at `__version__ = "0.1.0"` (package is at 0.8.0); `release-type: python` updates it natively on each release (it probes both the hyphen and underscore forms of the `pyproject.toml` name, so `common_grants_sdk/__init__.py` is found without an `extra-files` entry or an annotation comment). `lib/python-sdk/tests/test_version.py` now asserts `__version__` equals the `[tool.poetry]` version, so the drift that hid this for several releases fails the Python gate instead of shipping.
- **Added** `ci-pr-title.yml`, which validates conventional-commit PR titles against exactly the types release-please treats as releasable, since squashed PR titles are now release-facing. It is a regex over the event payload rather than a third-party action, so the job needs no token scopes. A title whose type has no `changelog-sections` entry would otherwise merge and then ship unversioned with no error — release-please drops such a commit from the notes and skips the release when the notes come out empty.
- **Docs**: new [ADR 0027](https://github.com/HHS/simpler-grants-protocol/blob/1217e822/website/src/content/docs/governance/adr/0027-release-please.md) records the decision (ADR 0010 gets a superseded banner); `lib/README.md`, the per-package `DEVELOPMENT.md` runbooks, `DEPENDENCY_MANAGEMENT.md`, and `CONTRIBUTING.md` describe the new flow.

### Context for reviewers

**How the new flow works:** merge a PR titled `fix(...)`/`feat(...)` → release-please opens or updates that package's Release PR → merging the Release PR tags, creates the GitHub release, and publishes to npm/PyPI automatically. `chore`/`ci`/`test` are hidden and don't trigger releases; every other accepted type (`fix`, `perf`, `revert`, `docs`, `refactor`, `build`) cuts a patch, `feat` a minor, and a `!` marker a breaking release (minor while pre-1.0). Publish fan-out happens via workflow outputs in the same run because tags created with `GITHUB_TOKEN` don't trigger `on: release` workflows.

**Verification:**

- `release-please release-pr --dry-run` against this branch: config parses, all four packages locate their last release from the existing tags, and it would open Version PRs for `@common-grants/core` 0.3.6, `@common-grants/cli` 0.3.8, and `common-grants-sdk` 0.8.1 (the pending `fix(deps)` advisory work since those tags; `lib/ts-sdk` correctly has nothing releasable since 0.6.1).
- Full workspace gate (`pnpm run ci`: core, changelog-emitter, cli, sdk, website) and the Python SDK gate (`make checks && make test && make build`, 397 tests, wheel builds at 0.8.0) pass locally.

**Repo settings (admin):**

1. **Already applied.** Squash-merge settings now use commit title `PR_TITLE` and message `PR_BODY`. The previous `COMMIT_OR_PR_TITLE` let a single-commit PR take its subject from the commit rather than the PR title, which would have bypassed the title check this PR adds.
2. **After merge:** re-enable the two `main` branch-protection rulesets (acceptance criterion on #870). This PR removes `ci-bump-version.yml`, which was the reason they were turned off, and release-please only opens PRs rather than pushing. One direct-to-`main` push does remain: `deps-mapping-repo.yml`'s scheduled README-diagram commit, which is guarded by a no-op check and has not actually fired since December 2025. It will be blocked once the rulesets are on ("Protect main (pushes)" has no bypass actors), so it is worth converting to a PR-based step separately. "Allow GitHub Actions to create and approve pull requests" is already on and needs to stay on.
3. Optional hardening: a tag-protection ruleset covering `@common-grants/*@*` and `common-grants-sdk@*` would require elevated permission to create release-shaped tags (the manual publish fallback trusts tag contents).

**Known gap:** commits between each package's last tag and this merge that don't follow the conventional format (e.g. #918, #947) are invisible to release-please, so the first Version PRs' changelogs may under-report. Changelog edits pushed to a Release PR branch survive, so the first ones can be hand-amended if it matters.

### Additional information

Dry-run output (abridged; run with `--target-branch=870-adopt-release-please`, so the PR-title scope shows the branch name — on `main` it reads `chore(main): ...`):

```
✔ Building strategies by path
❯ lib/core: node · lib/cli: node · lib/ts-sdk: node · lib/python-sdk: python
❯ Found release for path lib/ts-sdk, @common-grants/sdk@0.6.1
❯ Found release for path lib/python-sdk, common-grants-sdk@0.8.0
❯ Found release for path lib/core, @common-grants/core@0.3.5
❯ found: @common-grants/cli@0.3.7 1f67e30b8153922b67b23d4a0aac54762d6005ff
Would open 3 pull requests
title: chore(870-adopt-release-please): release @common-grants/core 0.3.6
title: chore(870-adopt-release-please): release @common-grants/cli 0.3.8
title: chore(870-adopt-release-please): release common-grants-sdk 0.8.1
```


website/src/content/docs/governance/adr/0027-release-please.md
```diff
@@ -0,0 +1,149 @@
...
+- The Python SDK uses `release-type: python` against the real `pyproject.toml`; the vestigial `package.json` and shell-parsing step are removed.
+- Merging a Release PR publishes automatically — no separate manual deploy step to remember.
+- Release notes are generated from PR titles, reading like a changelog rather than a list of changeset summaries.
+
+### Negative consequences
+
+- Conventional-commit PR titles become mandatory and release-facing; a mistyped title puts a change in the wrong category (or omits it entirely), so a CI title check is required.
+- Contributors must know which commit types map to which version bumps.
+- Commits that do not follow the conventional format are invisible to release-please, so changelog entries for the period before adoption may under-report.
+- A commit's type applies to every package whose files it touches; PRs that span packages need deliberate scoping.
```
karinamzalez: What happens to a PR that spans two packages?
re: "PRs that span packages need deliberate scoping.".. if I merge something that touches both lib/core/ and lib/ts-sdk/, do both packages get a patch Release PR? Is the guidance "split the PR," and is that written down anywhere?

.github/workflows/cd-deploy-lib-ts-sdk.yml
```diff
@@ -1,78 +1,58 @@
...
   deploy:
-    needs: create-release-from-tag
-    if: github.event.inputs.dry_run != 'true'
     runs-on: ubuntu-latest
     environment:
       name: npm-sdk
       url: https://www.npmjs.com/package/@common-grants/sdk
     permissions:
       contents: read
-      id-token: write
```
karinamzalez, line 66: out of curiosity: why is id-token: write being removed from deploy jobs?

CONTRIBUTING.md
```diff
@@ -123,6 +123,20 @@ If you've implemented a new feature, fixed a bug, or made some documentation cle
...
+
+| Type | Effect on the touched package(s) | Example |
+|------|----------------------------------|---------|
+| `fix` | Patch release | `fix(ts-sdk): preserve plain dates in JSON` |
+| `feat` | Minor release | `feat(core): add award schemas` |
+| `feat!` or `fix!` (or a `BREAKING CHANGE:` footer) | Breaking release (minor while pre-1.0) | `feat(cli)!: drop Node 20 support` |
+| `perf`, `revert`, `docs`, `refactor`, `build` | Patch release | `docs(website): clarify filter usage` |
+| `chore`, `ci`, `test` | No release; hidden from changelogs (a `!` marker still cuts a breaking release) | `chore(deps): bump the runtime group` |
+
+The scope is free-form; use the package or area you touched (`core`, `cli`, `ts-sdk`, `py-sdk`, `website`, `deps`). GitHub's **Revert** button titles the new PR `Revert "..."`, which fails this check — retitle it `revert: <description>` before merging. See [lib/README.md](lib/README.md) for how releases are cut from this history.
```
karinamzalez: Reverts seem like general high stakes scenarios -- is the manual rename here an intentional speed bump? Wondering if we'd be open to allowing the regex to just accept `Revert "..."`

karinamzalez (commented):
Hi Bryan! It's really neat to see us adopting release-please! I have a few clarifying questions. Some are file-specific but the more generic one: 

- should docs only changes be hidden alongside chore/ci/test? -- e.g. a `docs(core): fix typo in README` PR would publish a new npm release .. I don't think we'd write a changeset for that?
- is missing-release-check.yml still necessary? looks like it's a nightly job that flags tag without a GH release. I think under release-please, tags and releases are created together so it would never fire -- dead code?

website/src/content/docs/governance/adr/0027-release-please.md
SnowboardTechie: Both, yeah. release-please splits commits by path, so a commit touching `lib/core/` and `lib/ts-sdk/` lands in both packages' lists and the title's type applies to each.

Splitting isn't the default guidance though, usually both packages really did change and both should release. What warrants a split is when they deserve different bumps, like a `feat` in core alongside a mechanical regen in the sdk that would otherwise take a minor it didn't earn. Expanded this bullet to say that and pointed it at CONTRIBUTING, which covers it now too.

.github/workflows/cd-deploy-lib-ts-sdk.yml
SnowboardTechie: It wasn't doing anything. No `--provenance` on any of the publishes and no attestations on the published versions, and PyPI goes through `POETRY_PYPI_TOKEN_PYPI` rather than trusted publishing. So it was a permission with no consumer. It comes back if we want provenance later.

CONTRIBUTING.md
SnowboardTechie: Intentional, and it's doing more than slowing you down. release-please can't parse `Revert "feat(core): add award schemas"`, the header grammar is fixed and there's no knob for it. An unparseable header gets dropped at debug level, so a revert merged under GitHub's default title ships with no version bump and no error. Loosening the regex would let through exactly what the check exists to catch.

Added that reason to the line so it doesn't read as arbitrary.

SnowboardTechie:
@karinamzalez thank you for the review! On docs, releases are scoped by path, not by the title's scope, so `docs(website): fix typo` releases nothing. Inside a package it does release, and I'd keep it: npm ships `README.md` in the tarball regardless of the `files` allowlist, so it's a real change to the published artifact. Nothing publishes on the merge either, it just opens or updates a Release PR that sits until we merge it. `chore(core):` is the lever for internal notes that shouldn't ship. Pushed that to CONTRIBUTING, it wasn't written down.

Good catch on `missing-release-check.yml`, that's dead and I've deleted it along with `check-for-unreleased-tags.sh`. It covered the old gap where `ci-bump-version.yml` pushed the tag on merge and release creation was a separate manual dispatch. release-please does both in one step.

karinamzalez (approved):

PR #1053, opened 2026-07-30
fix(core): emit RFC 7396-compliant merge-patch schemas for OrgPatchData

### LINKED ISSUE #1026: [Org IDs] Fix how `OrgPatchData` is generated
### Summary

The generated merge-patch schemas for the org profile (`OrgPatchData` and its `*OrCreate` children) don't faithfully represent RFC 7396 in two systemic ways. Both come from TypeSpec's `MergePatchUpdate`, not from hand-rolled models, but consumers validate against these published schemas.

### Root causes

**1. `Record` members have no `null` branch.** For a `Record<T>` field, the transform applies only the create-or-update shape to the value and never adds a `null` branch (record members are treated as always create-or-update). So per-key deletion, which RFC 7396 expresses as `null`, fails validation. Clearing a whole field works (`{"customFields": null}`); deleting one key does not (`{"customFields": {"foo": null}}`).

This affects the `other*` record on nearly every sub-model:

- `customFields` (`Record<Fields.CustomField>`)
- `identifiers.otherIds` (`Record<Identifier>`)
- `addresses.otherAddresses` (`Record<Address>`) and `addresses[].geography` (`Record<unknown>`)
- `emails.otherEmails` (`Record<Types.email>`)
- `phones.otherPhones` (`Record<Fields.Phone>`)
- `socials.otherSocials` (`Record<url>`)

**2. `extends` base fields are inherited as-is.** When a patched model extends another (`OrgIds extends IdentifierCollection`), the generated patch schema references the base as `allOf: [IdentifierCollection.yaml]`, which is the un-patched read schema. All of the fields that are inherited through `extends` skip the merge-patch transformation, which means they're not made optional or nullable, and that nested records like `otherIds` require a full object rather than a patch record. You can see this pattern in `OrgPatchOrgIdsOrCreate` (`identifiers`) and `OrgPatchPCSOrgTypeOrCreate` (`orgType`).

### Options

While the long-term fix needs to occur upstream in `@typespec/http`, so that the transformation adds a `null` option to `Record` keys and also applies to models referenced by `extends`, we have a few options for an interim fix:

1. **Custom post-processor.** A step in the emit pipeline (alongside `prune-openapi-tags.ts`) that rewrites the generated `OrgPatch*` schemas: add a `null` branch to record members and replace `allOf: [PlainBase]` with a nullable/patched inline. Fixes both issues, but needs to be replicated everywhere we're generating the OpenAPI spec/JSON schemas (core, CLI, website, TS SDK).

2. **Custom mutator.** A small TypeSpec JS library exposing our own `OrgMergePatchUpdate<T>` built on the compiler's experimental `mutateSubgraph`. Fixes both issues in a re-usable way across the repo, but with the extra overhead of maintaining a custom emitter.

3. **Compose sub-models manually.** Run `MergePatchUpdate` on each sub-model and hand-assemble `OrgPatchData` from the patched pieces, referencing patched sub-models instead of relying on recursion. Provides the most control but also requires the most maintenance and could lead to drift between `OrgPatchData` and `OrganizationBase`

### Acceptance criteria

- [x] A bug with a reproducible example is filed upstream against `@typespec/http` for both root causes, and linked here
- [x] We've recorded a decision on which interim option (1, 2, or 3) to implement, or we've documented our decision to wait for an upstream fix
- [x] The chosen approach makes `{"customFields": {"foo": null}}` (and the equivalent for other fields) pass validation, while still rejected invalid merge patch inputs.
- [x] The fields inherited from `IdentifierCollectionT` (`systemId`, `otherIds`) also support the correct merge patch behavior (optional, clearable)
- [ ] ~~`OrgPatchData` doc notes any remaining issues or conflicts until a more reliable upstream fix is available~~

SnowboardTechie:
### Summary

- Fixes #1026
- Time to review: 20 minutes

### Changes proposed

`OrgPatchData` and its generated `OrgPatch*OrCreate` schemas now accept RFC 7396 merge patches that the published 0.4.0 schemas rejected:

- Record members (`customFields`, `otherIds`, `otherAddresses`, `otherPhones`, `otherEmails`, `otherSocials`) accept `null`, so a patch can delete individual keys (`{"customFields": {"foo": null}}`).
- Fields inherited through `extends` (`identifiers.systemId`, `identifiers.otherIds`, and the `PCSTerm` fields of `orgType`) are optional, clearable, and patch-shaped, instead of being referenced as the un-patched base schema through `allOf`.

How: a non-emitted `CommonGrants.Patch` namespace in `organization-sync.tsp` mirrors each org profile sub-model with `extends` flattened via spread and record members accepting `null`; `OrgProfileWritable` references the mirrors. The single `MergePatchUpdate` derivation and all emitted `OrgPatch*` schema names are unchanged.

Also:

- Regenerated the 0.4.0 artifacts. The collection-level `OrCreate` schemas now emit inline as `$defs` of `OrgPatchData.yaml` (they derive from the non-emitted namespace), so six standalone yaml files fold into it, and two new leaf schemas appear (`OrgPatchIdentifierOrCreate`, `OrgPatchSystemIdOrCreate`).
- Added `website/src/lib/schema/org-patch-data.test.ts`, which validates the published schemas verbatim: per-key deletion, inherited-field clearing, rejection of malformed patches, and a drift guard comparing `OrganizationBase` properties against `OrgPatchData`.
- The `OrgPatchData` docs example now demonstrates record-key deletion (`customFields: { legacyCode: null }`).

### Context for reviewers

- Implements option 3 per the [decision recorded on #1026](https://github.com/HHS/simpler-grants-protocol/issues/1026#issuecomment-5137057345). Both root causes are upstream in `@typespec/http`; the upstream bug reports are tracked by #1026's first acceptance criterion. The deliberate OrCreate strictness (nested patch values must be valid standalone creates) is unchanged and stays the contract.
- Verified with the full root `pnpm run ci` (all packages green). The new test discriminates: run against the pre-fix schemas it fails on exactly the four fixed behaviors (three record-key deletions and the inherited `systemId` clear) and nothing else.
- The test uses its own Ajv instance because the website's shared loader strips top-level `unevaluatedProperties`; external consumers validate the published files verbatim, so sealing is part of the contract under test.
- OpenAPI component naming: the inlined sub-models emit as `CommonGrants.Patch.OrgPatch*OrCreate` (previously `CommonGrants.Fields.OrgPatch*OrCreate`). 0.4.0 is unreleased, so no published name changes.
- The CLI's `lib/openapi` snapshots are stale relative to this branch (they predate the org routes) and get refreshed at release prep, so they are deliberately untouched here.

### Additional information

`identifiers.otherIds` in the published `OrgPatchData.yaml`, before and after:

```yaml
# before: inherited as-is from the read schema
allOf:
  - $ref: IdentifierCollection.yaml

# after: optional, clearable, per-key deletable
otherIds:
  anyOf:
    - type: object
      properties: {}
      unevaluatedProperties:
        anyOf:
          - $ref: OrgPatchIdentifierOrCreate.yaml
          - type: "null"
    - type: "null"
```


lib/core/lib/core/models/organization-sync.tsp
```diff
@@ -4,22 +4,135 @@ import "@typespec/http";
...
+ * name of the model it mirrors so the generated `OrgPatch{name}` schemas keep
+ * stable names. If a mirrored model gains or loses a field, the change flows
+ * through spread automatically; only the record-typed fields are redeclared
+ * here and must be kept in sync by hand. The same applies one level down: a
+ * leaf type reached from these mirrors (`Address`, `Phone`, `Identifier`,
+ * `CustomField`) needs a mirror of its own if it ever gains `extends`.
+ * `website/src/lib/schema/org-patch-data.test.ts` fails if either slips.
+ */
+namespace CommonGrants.Patch {
+  /** A collection of identifiers associated with an organization.
```
widal001, line 26: Small suggestion: Since we're adding this new namespace whose members also aren't emitted as JSON schemas, is it worth moving `OrgProfileWritable` up into this namespace to get it out of the root-level namespace?

widal001 (approved):
Great work! I actually like that this has the secondary benefit of defining the sub-models in the `$defs` section rather than as a bunch of separate models.

Image
Image

lib/core/lib/core/models/organization-sync.tsp
SnowboardTechie: I think so, that's cleaner. Moved it into `CommonGrants.Patch`.

PR #1054, opened 2026-07-31
871 simplify openapi doc

### LINKED ISSUE #871: [Website] Simplify the OpenAPI docs component
### Summary

Replace the `swagger-ui-react`-based `OpenApiDocs.tsx` component with a simpler Astro page that mounts `swagger-ui-dist` directly and uses Swagger's built-in dark theme instead of the current CSS-in-JS overrides.

### Rationale

The current `OpenApiDocs.tsx` component works, but has a few complexities that switching to `swagger-ui-dist` would remove:

- The team maintains a full React component with state, `useEffect`, and a re-render key for what's essentially a panel with a version dropdown
- The component requires CSS overrides to keep `swagger-ui-react` from clashing with Starlight's theme
- The site pulls in `swagger-ui-react` as a dep even though `swagger-ui-dist` is already in `website/package.json`
- `swagger-ui-react` has been a source of vulnerabilities in the past

The benefits of switching:

- The same panel renders as a plain Astro page mounting `swagger-ui-dist`, no React state and no re-render key
- We can switch to a much simpler dark mode toggle, and remove the custom CSS overrides
- We can drop the `swagger-ui-react` dep entirely

There are a few tradeoffs though:

- The page can't be embedded as a React component inside MDX content the way the current component can (not that we're actively doing this now)
- We need to change some of the code around for updating the URL with the API version being selected.

### Implementation notes

The CommonBenefits page provides a pretty decent reference, about 130 lines including styles and script:

- [`website/src/pages/protocol/api-docs.astro`](https://github.com/common-benefits/common-benefits/blob/main/website/src/pages/protocol/api-docs.astro) — the whole page
- The dark-mode bit worth copying: a four-line `MutationObserver` on `<html>` that mirrors Starlight's `data-theme` into the `dark-mode` class
- `SwaggerUIBundle` from `swagger-ui-dist` mounts on a `<div id="swagger-ui">`, and the `<select>` change handler re-renders the panel with the new spec URL

Files that would need to be removed or updated:

- [`website/src/components/OpenApiDocs.tsx`](https://github.com/HHS/simpler-grants-protocol/blob/main/website/src/components/OpenApiDocs.tsx) replaced by an Astro page (likely at `website/src/pages/protocol/api-docs.astro`)
- `swagger-ui-react` removed from `website/package.json`
- Any MDX page that currently embeds `<OpenApiDocs />` needs to switch to a link to the Astro page

### Open questions

- Does anyone embed `<OpenApiDocs />` inside MDX content today? If so, what's the migration path for those pages?
- Are there features in `swagger-ui-react` that we'd lose by dropping to `swagger-ui-dist`?

### Acceptance criteria

- [ ] The OpenAPI docs page renders using `swagger-ui-dist` and is no longer a React component
- [ ] Dark mode follows Starlight's theme without CSS overrides
- [ ] Version switching still works, including via the `?version=` URL param as a sharable deep link
- [ ] `swagger-ui-react` is gone from `website/package.json`

karinamzalez:
### Summary

- Fixes #871
- Time to review: 10 minutes

- Replaced the `swagger-ui-react` React component (`OpenApiDocs.tsx`) with a plain Astro page that mounts `swagger-ui-dist` directly — no React state, `useEffect`, or re-render key.
- Deleted the ~300-line CSS override block. Dark mode now uses Swagger UI's built-in dark theme, switched on by a 4-line `MutationObserver` that mirrors Starlight's `data-theme` to a `.dark-mode` class.
- Dropped `swagger-ui-react` (+ its types); added `swagger-ui-dist` (+ its types) as a direct dependency.
- Moved the `?version=` logic into a helper (`api-docs-version.ts`) with unit tests, and fixed a bug where deep links using the `v`-prefixed form (e.g. `?version=v0.2.0`, used by RFC 0002) fell back to the latest version instead of the one requested.

- `swagger-ui-react` has a history of vulnerabilities and needed a lot of CSS to fit Starlight's theme. `swagger-ui-dist` was already installed under the hood, so this swap drops the wrapper, the overrides, and the React island at once.
- The version dropdown, `?version=` deep links, and the disabled "Try it out" buttons all work exactly as before. The OpenAPI YAML files are unchanged — this only touches how they're rendered.
- Trade-off: the page can no longer be embedded in MDX as a React component. Nothing did that, so nothing needs migrating.

**How it was verified:**
- Build and checks pass: `pnpm --filter website run build` (153 pages) and `pnpm --filter website run checks` (eslint, prettier, `tsc`, `astro check`, cspell).
- Tests pass: `pnpm --filter website run test` — 80 tests, including 4 new ones for the version helper (bare version, `v`-prefix, missing, and unknown values).
- End-to-end run in headless Chrome, all passing:
  - Panel renders via `swagger-ui-dist` with no React island; defaults to `v0.3.0` and loads the 0.3.0 spec.
  - `?version=v0.2.0` loads v0.2.0 (not 0.3.0); `?version=bogus` falls back to 0.3.0.
  - Changing the dropdown updates the URL without a reload and loads the right spec.
  - Dark mode follows the theme toggle, and the panel background changes with it.

**To review locally:** run `pnpm --filter website run dev`, open `/protocol/api-docs`, then toggle the theme, switch versions, and try `/protocol/api-docs?version=v0.2.0`.

**Dark mode (built-in theme):
01-dark-v0 3 0
**

**Light mode:**
03-light-v0 3 0

**Deep link `?version=v0.2.0`:**
04-dark-v0 2 0-deeplink


website/src/pages/protocol/api-docs.astro
```diff
@@ -1,323 +1,102 @@
...
+  import { SwaggerUIBundle } from "swagger-ui-dist";
+  import { resolveVersion } from "@/lib/api-docs-version";
+
+  const render = (version: string): void => {
+    SwaggerUIBundle({
+      url: `/openapi/openapi.${version}.yaml`,
+      dom_id: "#swagger-ui",
+      supportedSubmitMethods: [],
+    });
+  };
```
widal001, line 90:
```suggestion
  const render = (version: string): void => {
    SwaggerUIBundle({
      url: `/openapi/openapi.${version}.yaml`,
      dom_id: "#swagger-ui",
      supportedSubmitMethods: [],
      deepLinking: true,
    });
  };
```

widal001: Adds `deepLinking: true` so that you can pass [links to specific operations](https://commonbenefits.dev/protocol/api-docs/?version=0.1.0#/Applications/Applications_submit)

website/src/pages/protocol/api-docs.astro
```diff
@@ -1,323 +1,102 @@
...
-  :global(.swagger-ui) {
-    color: var(--sl-color-text) !important;
-    font-family: sans-serif !important;
-    background-color: var(--sl-color-background) !important;
+<style>
+  /* Layout for the version selector only — Swagger's own styles (incl. its
+     built-in dark theme) handle the panel itself. */
+  .version-selector {
+    max-width: 480px;
   }
```
widal001, line 35: There are a few more style overrides it would be good to have:

### Version pill borders

This fixes the version pill borders that show up by default.

```css
:global(.swagger-ui .version) { border: none !important }
```
Image

Although there might be a deeper root cause for this that is worth finding and fixing too.

widal001: ### Container width

```css
:global(.sl-container) {
  max-width: 100% !important;
  padding: 0 1.5rem !important;
}
```

Without the override, things look constrained:

Image

With the override, they take up more of the width

Image

widal001: This is less important but I think this will also fix the extra background shading in dark mode:

```css
:global(html.dark-mode) {
  background: var(--sl-color-bg) !important;
}
```

Image

widal001 (commented):
@karinamzalez Great work!! Loved seeing the diff

I left a few small suggestions, the deep linking one is the most useful to include, the others are mostly styling.

website/src/pages/protocol/api-docs.astro
karinamzalez: makes sense-- thank you!

website/src/pages/protocol/api-docs.astro
karinamzalez: > There are a few more style overrides it would be good to have:
> 
> ### Version pill borders
> This fixes the version pill borders that show up by default.
>...
> Although there might be a deeper root cause for this that is worth finding and fixing too.


Good catch! Looks like Swagger renders version pills as `<pre class="version">` and Starlight styles _every_ `<pre>` with a solid border for code blocks and that bleeds into the pill. Not touching the pre rule bc code blocks rely on it but I added a scoped override based on what you suggested. I feel like this is a good way to avoid an `important!` here -- does this look good to you?
```
  :global(.swagger-ui .version) {
    border: none;
  }
```
Screenshot 2026-08-03 at 1 22 59 PM

website/package.json
```diff
@@ -50,7 +50,7 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "sharp": "^0.35.3",
-    "swagger-ui-react": "^5.32.11",
+    "swagger-ui-dist": "^5.32.1",
```
SnowboardTechie: Heads up that this swap quietly downgrades Swagger UI: the react wrapper was shipping 5.32.11, but `^5.32.1` resolves to exactly 5.32.1 here because pnpm reused the resolution already in the lockfile (it was a transitive dep of swagger-ui-express). Bumping the specifier to ^5.32.12 and refreshing the lockfile fixes it. Dark mode is still keyed off `html.dark-mode` in 5.32.12, so your MutationObserver approach carries over unchanged, just worth re-running the visual pass after the bump.

website/src/pages/protocol/api-docs.astro
karinamzalez: >Container width
>:global(.sl-container) {
  max-width: 100% !important;
  padding: 0 1.5rem !important;
}
>Without the override, things look constrained:
>...

I agree that it looks better with the wider width! Looks like Starlight caps the content width of `sl-container` at 45rem .. The following update lets it span the full column without an !important: 
```
  :global(.sl-container) {
    max-width: 100%;
  }
```
Screenshot 2026-08-03 at 1 48 23 PM

website/package.json
karinamzalez: 🫢 -- confirmed and fixed!! Thank you Bryan

widal001 (approved):
If you can resolve the vulnerabilities surfaced by the latest CI check run, then this looks good to merge in from my perspective. Thanks for the hard work on this!

PR #1061, opened 2026-08-03
fix(ci): upgrade to pnpm 11 and restore native dependency audits

### LINKED ISSUE #731: [Deps] Remove audit-deps.js workaround after pnpm v11 upgrade
### Summary

PR #729 added `.github/scripts/audit-deps.js` as a workaround for npm retiring their legacy audit endpoint, which broke `pnpm audit` with HTTP 410 (tracked upstream at pnpm/pnpm#11265). Once this repo upgrades to a pnpm version with native bulk advisory support (expected v11+), this workaround should be removed.

### Acceptance criteria

- [ ] `.github/scripts/audit-deps.js` has been deleted
- [ ] All workflow audit steps have been reverted to use `pnpm audit` — each workflow file has a comment documenting the original command to restore
- [ ] Express-js template GHSA suppressions are handled via `pnpm audit` (either through `auditConfig.ignoreGhsas` or `--ignore` flags)
- [ ] CI passes with native `pnpm audit` on all workflows

### LINKED ISSUE #1060: [Dependency Mgmt] Template CI audit steps pass vacuously — audit-deps.js audits 0 packages from a template directory
### Summary

`.github/scripts/audit-deps.js` discovers packages by shelling out to `pnpm list --json --depth=Infinity`, whose output depends on the directory it runs in. The template CI jobs run it from a template directory, and because the templates are standalone packages outside the pnpm workspace, `pnpm list` there returns the package with no dependencies. The audit step reports `Auditing 0 packages… No vulnerabilities found` and exits 0 regardless of what is installed.

Verified in `templates/express-js`:

```
# from templates/express-js, as ci-template-express-js.yml invokes it
Auditing 0 packages…
No vulnerabilities found.

# same tree, actually audited
$ pnpm audit --ignore-workspace --audit-level high
30 vulnerabilities found
Severity: 1 low | 11 moderate | 16 high | 2 critical
```

`pnpm list --json --depth=Infinity --ignore-workspace` from that directory reports 7 dependencies and 25 devDependencies. The hidden advisories sit in tar, vitest, brace-expansion, fast-uri, form-data, js-yaml, postcss, vite, body-parser, qs and uuid — all pre-existing, none from the templates' own code.

`templates/quickstart` has the same wiring and also audits 0 packages. `ci-template-fast-api.yml` is unaffected; it runs `poetry audit`.

Workspace-scoped invocations are fine: `--filter @common-grants/cli` audits 183 packages, `--filter @common-grants/sdk` 99, and the repo-wide `--level high` run from `./website` 696.

The step's own comment says to "revert to `pnpm audit --ignore-workspace` once pnpm ships native support", so the dropped flag looks like an oversight when the script replaced the direct call.

### Acceptance criteria

- `audit-deps.js` audits a standalone template's real dependency tree when invoked from that template's directory.
- A known-vulnerable dependency added to a template fails that template's CI job, confirming the gate is live.
- The advisories currently hidden in `templates/express-js` are triaged: upgraded, or suppressed with a rationale comment matching the existing four.
- The `--ignore-ghsa` list in `ci-template-express-js.yml` is re-checked once the audit runs for real.

SnowboardTechie:
### Summary

- Fixes #1060
- Fixes #731
- Time to review: 15 minutes. The three lockfile diffs are regeneration noise; the reviewable surface is the eight workflow call sites, the two new template `pnpm-workspace.yaml` files, and the override floors.

### Changes proposed

Upgrades the workspace and both standalone templates from pnpm 10.33.0 to 11.20.0, whose `pnpm audit` uses npm's current bulk advisories endpoint natively. That removes the reason the workaround existed.

- **Deletes `.github/scripts/audit-deps.js`** (208 lines) and restores native `pnpm audit` at all eight workflow call sites, each keeping the severity threshold the script ran at: `--audit-level=moderate` for catalog validation, `audit:high` for the website, and low (pnpm's default) for the two templates and the four `--filter` package audits.
- **Makes the two template audit gates real.** Every advisory they were hiding is now fixed through in-range dependency updates and raised security floors. No suppressions remain anywhere: the stale `auditConfig.ignoreGhsas` list in the express-js template is deleted outright, and no `--ignore` flags replace it. All four previously suppressed GHSAs are resolved rather than silenced, since the tree now installs minimatch 3.1.5, tar 7.5.21, ajv 6.15.0/8.18.0 and qs 6.15.2.
- **Moves template pnpm settings out of the `package.json` `"pnpm"` field**, which pnpm 11 no longer reads, into a per-template `pnpm-workspace.yaml`. This also makes each template its own workspace root, so the template CI jobs drop `--ignore-workspace`: under pnpm 11 that flag would skip the template's own config file, security overrides included. `templates/README.md`'s dependency-fix instructions are updated to match, since following the old steps would have regenerated a template lockfile with those floors stripped.
- **Pins the new pnpm 11 `allowBuilds` entries to `false`**, preserving exactly the build-script behavior CI already ran with under pnpm 10. The quickstart template needs no such block; nothing in its tree runs a build script.
- **Raises the documented Node prerequisite** in `lib/ts-sdk/DEVELOPMENT.md` to 22.13, which is pnpm 11.20.0's own `engines.node` requirement.

Separately, and not driven by any advisory: the express-js template's dependencies get a deliberate version refresh (`prettier`, `tsx`, `typescript-eslint`, `vitest`, `eslint`, `js-yaml`, `uuid`, `@types/node`, and `@common-grants/core` 0.3.2 to 0.3.5). This is the maintainer-initiated refresh that `DEPENDENCY_MANAGEMENT.md`'s tiered policy calls for on templates, and it accounts for most of that manifest's diff. It is independent of the security-floor work above.

### Context for reviewers

#1060 documents the underlying bug: `audit-deps.js` discovered packages by shelling out to `pnpm list`, whose output depends on the directory it runs in. Run from a standalone template directory, that returns the project with no dependency tree, so both template audit steps reported `Auditing 0 packages…` and exited 0 regardless of what was installed. #1060 recorded 30 real advisories sitting behind that green check in `templates/express-js` alone (2 critical, 16 high, 11 moderate, 1 low).

Rather than patch a script whose own header says to delete it once pnpm ships native support, this takes the documented exit in #731. The pnpm 11 line has been stable since late April, and its audit works against the current registry endpoints.

Verified locally, running each command exactly as the workflows invoke it:

- Root workspace: `pnpm run ci` green across core, changelog-emitter, cli, sdk and website. `pnpm audit` exits 0 at both `moderate` and the stricter `low`.
- Both templates: `pnpm install --frozen-lockfile`, `pnpm run checks`, the build/typespec step, and `pnpm audit` all green. The templates gate at low without passing the flag, since low is pnpm's own default; `pnpm audit --json` reports their real trees (447 and 112 packages) with zero advisories at every severity.
- The security floors reach CI, not just local runs: each template's lockfile records the `overrides` block from its own `pnpm-workspace.yaml`, and the CI jobs install that lockfile with `--frozen-lockfile`.
- Every override floor is load-bearing. Four entries carried over from the old `package.json` blocks matched nothing in their own lockfiles and are dropped; regenerating both template lockfiles afterwards changed only the recorded overrides block, leaving every resolved version identical.

### Additional information

The gate is live, not just non-vacuous. Adding `lodash@4.17.20` to the express-js template makes its audit step fail, and the [template CI run](https://github.com/HHS/simpler-grants-protocol/actions/runs/30853928686) recorded that failure on this branch before the test dependency was removed:

```
5 vulnerabilities found
Severity: 3 moderate | 2 high
```

The same step on the current branch state, auditing the template's real 447-package tree:

```
$ pnpm audit
No known vulnerabilities found
```

For contrast, this is what the step printed before this PR, from the same directory with the same dependencies installed:

```
Auditing 0 packages…
No vulnerabilities found.
```

No changeset: nothing here changes published-package behavior. The pnpm requirement is a dev-toolchain constraint, documented in `lib/ts-sdk/DEVELOPMENT.md`.


.github/workflows/ci-template-express-js.yml
```diff
@@ -42,19 +43,4 @@ jobs:
...
-        #   GHSA-83g3-92jg-28cx  high     node-tar symlink escape (transitive of npm/pacote)
-        #   GHSA-2g4f-4pwh-qvx6  moderate ajv ReDoS with $data (transitive of swagger-ui-express)
-        #   GHSA-w7fw-mjwx-w883  low      qs arrayLimit DoS (transitive of express body-parser)
-        run: >
-          node ${GITHUB_WORKSPACE}/.github/scripts/audit-deps.js
-          --ignore-ghsa GHSA-3ppc-4f35-3m26
-          --ignore-ghsa GHSA-83g3-92jg-28cx
-          --ignore-ghsa GHSA-2g4f-4pwh-qvx6
-          --ignore-ghsa GHSA-w7fw-mjwx-w883
+        run: pnpm audit --audit-level=low
```
widal001: Do we want these to flag at low?

widal001 (approved):
Woohoo happy to get onto pnpm 11 and glad that it helps us switch back to the native `pnpm audit`

.github/workflows/ci-template-express-js.yml
SnowboardTechie: Yeah, low is what I think we should keep. It's pnpm's default, so the four `lib/*` audits are already running at low today via the bare `pnpm audit` in their package scripts. This was just saying it out loud, so I went ahead and dropped the flag.

The reason I'd keep it there specifically for templates: they're the manual maintenance tier, no Dependabot, so this audit is the only thing watching their trees.

PR #953, opened 2026-06-25
feat(core): add boolean and integer comparison filter models

SnowboardTechie:
### Summary

- Closes #895
- Time to review: ~5 minutes

Adds the two protocol-level filter models the custom-filter work needs — `BooleanComparisonFilter` and `IntegerComparisonFilter` — so the SDK filter catalogs have a real spec model to validate against instead of borrowing `NumberComparisonFilter`. Spec-only (`lib/core` TypeSpec); no SDK changes in this PR.

### Changes proposed

- New `lib/core/lib/core/filters/boolean.tsp` — `BooleanComparisonFilter` (`operator: eq | neq`, `value: boolean`), `@Versioning.added(v0_3)`.
- `lib/core/lib/core/filters/numeric.tsp` — add `IntegerComparisonFilter` (`operator: eq | neq | gt | gte | lt | lte`, `value: integer`), `@Versioning.added(v0_3)`.
- `lib/core/lib/core/filters/index.tsp` — export `boolean.tsp`.
- Changeset: `@common-grants/core` minor.

### Context for reviewers

**Why:** the #884 / #885 custom-filter PoCs added `booleanComparison` and `integerComparison` to the SDK filter-type catalogs, but `lib/core` had no matching TypeSpec models — so `integerComparison` validated against `NumberComparisonFilter` and the integer constraint was enforced nowhere.

**Decisions:**

- Integer value type = TypeSpec `integer` → JSON Schema `type: integer`. At the spec level this treats `100.0` as a valid integer (zero fractional part), matching Zod's `z.number().int()`. Strict-vs-lenient enforcement per SDK (e.g. Pydantic `StrictInt`) is the SDK's call — tracked by #907 (TS) and #896 (Py) under "align SDK schemas with the TypeSpec filter definitions (operator surface and value strictness)".
- Versioned at protocol v0.3.0, the open contract version. `IntegerComparisonFilter` is net-new at v0.3.0 (full operator union from the start; no `typeChangedFrom`).

**Verification:** `pnpm ci:core` green — eslint, `prettier --check` + `tsp format --check`, `tsc --noEmit`, build, `tsp compile`. Both models emit to `lib/core/tsp-output/@typespec/json-schema/`. Like every filter model, they don't appear in the openapi3 components because no core route references them.

**Base branch:** targets `HOLD-filters` (filters-epic integration branch), consistent with sibling PRs #941 / #934 / #928 — not `main`.

### Additional information

Emitted JSON Schema:

```yaml
[...]

PR #1066, opened 2026-08-04
fix(core): apply remaining v0.4.0 hold branch audit findings

### LINKED ISSUE #1032: [Core] Audit and clean up the v0.4.0 hold branch

SnowboardTechie:
### Summary

Second cleanup pass over the v0.4.0 hold branch, covering the three items left open on #1032:

- `BooleanComparisonFilter` ships in 0.4.0 but had no docs page, so every plugin's `booleanComparison` filter linked to the filters overview instead of a model.
- The two status models added in 0.4.0 hand-rolled the extensible enum shape instead of instantiating `Fields.ExtensibleEnumT<T>` like every other extensible enum field.
- ADR-0023's recommended examples described a richer `registry` object than the one that shipped.

- Closes #1032
- Time to review: 10 minutes

### Changes proposed

> What was added, updated, or removed in this PR.

**Document `BooleanComparisonFilter`** — adds `protocol/filters/boolean.mdx` (table, format tabs, changelog) and a Boolean card on the filters overview, then points the plugin filter map at `/protocol/filters/boolean#booleancomparisonfilter`. The map previously sent `booleanComparison` to the overview because no boolean model existed when the custom filters section was built in #1045; the model landed in #953 on this branch. Both tests that pinned the old fallback are dropped rather than repointed: removing the `if (!anchor) continue;` guard raises the anchor-exists check from the nine types with anchors to all ten, which covers `booleanComparison` on its own, and the loader resolves every filter through one `filterDocsHref` call, so the `agency` case already covers filterType to docsHref. The py-SDK docstring saying no spec model backs the filter is removed for the same reason.

**Define the 0.4.0 status models via `ExtensibleEnumT`** — `AwdStatus` and `RevisionStatus` both spelled out `value` / `customValue` / `description` by hand. `OppStatus`, `AppStatus`, `ApplicantType`, `FormResponseStatus`, and `CompetitionStatus` all instantiate the template, so these two were the only extensible enum fields with a second source of truth for the shape. `RevisionStatus` is included alongside `AwdStatus` because it has the same defect and ships in the same release. Wire shape is unchanged; the regenerated `AwdStatus.yaml`, `RevisionStatus.yaml`, and `openapi.0.4.0.yaml` diffs are property descriptions moving to the template's wording, which is what `.changeset/extensible-enum-field.md` already documents for the other five fields. That changeset's list is extended rather than adding a new one.

**Align ADR-0023 with the shipped identifier** — the recommended nested-hybrid example carried `scope`, `kind`, `schema`, and a top-level `uri`, and the recommended parent-reference example carried `scope` and `kind`. `IdentifierT.registry` defines `code` and `url` only, and both are optional, so the prose describing `code`/`url` as required is corrected too. The pros, cons, and three comparison rows that rested on the dropped fields move with the example: inline scope/kind filtering and per-registry schema validation are now ❌ for the nested hybrid, payload brevity ✅, and self-contained-without-a-catalog ❌. The remaining `scope`/`kind` examples are in the rejected flat and fully-decomposed options, where the richer shape is the point, and the `US-EIN`/`US-SAM` codes are in the rejected adopt-org-id.guide-codes option. This is the follow-up to karinamzalez's question on #1040 about whether ADR-0023 should drop the fields ADR-0026's examples dropped.

### Context for reviewers

> Testing instructions, background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers. Explain how the changes were verified.

Verified with `pnpm run ci` from the root (exit 0), plus `make checks && make test` in `lib/python-sdk` (0 pyright errors, 396 passed). The website build reports all internal links valid, which covers the new overview card and the anchors on the new page.

The anchor check was confirmed to discriminate: pointing `booleanComparison` back at `/protocol/filters/` fails it.

### Additional information

> Screenshots, GIF demos, code examples or output to help show the changes working as expected.

The shipped shape the ADR examples now match, from `lib/core/lib/core/fields/identifier.tsp`:

```typespec
model IdentifierT<Id extends string = string, Code extends string = string> {
  registry?: {
    code?: Code;
    url?: url;
  };
  id?: Id;
  allIds?: Array<{ id: Id; status: IdentifierStatus }>;
}
```

The regenerated schema diff for both status models is descriptions only:

```diff
   value:
     $ref: AwdStatusOptions.yaml
-    description: The status of the award, from a predefined set of options
+    description: The selected value, from a predefined set of options
   customValue:
     type: string
-    description: A custom value for the status
+    description: A custom value, used when the selected value is the `custom` option
```

### DIFF
--- a/website/__tests__/lib/plugins/loader.spec.ts
+++ b/website/__tests__/lib/plugins/loader.spec.ts
@@ -30,16 +30,18 @@ describe("plugins loader", () => {
       expect(agency?.description).not.toBe("");
     });
 
-    // The one filter type with no protocol model of its own, so it is the one
-    // whose href comes from the fallback rather than the anchor map.
-    it("sends a booleanComparison filter to the filters index", () => {
+    // The newest filter model, and the one a plugin declares against a boolean
+    // custom field, so it is the likeliest to lose its docs link in a rewire.
+    it("links a booleanComparison filter to the boolean filter model", () => {
       const plugin = loadAllPlugins().find((p) => p.id === "cg-grants-gov");
       const costSharing = plugin?.resolvedFilters.find(
         (f) => f.name === "costSharing",
       );
 
       expect(costSharing?.filterType).toBe("booleanComparison");
-      expect(costSharing?.docsHref).toBe("/protocol/filters/");
+      expect(costSharing?.docsHref).toBe(
+        "/protocol/filters/boolean#booleancomparisonfilter",
+      );
     });
   });
 });
--- a/website/__tests__/lib/plugins/loader.spec.ts
+++ b/website/__tests__/lib/plugins/loader.spec.ts
@@ -29,19 +29,5 @@ describe("plugins loader", () => {
       // Authored per filter in index.json; "" means the wiring dropped it.
       expect(agency?.description).not.toBe("");
     });
-
-    // The newest filter model, and the one a plugin declares against a boolean
-    // custom field, so it is the likeliest to lose its docs link in a rewire.
-    it("links a booleanComparison filter to the boolean filter model", () => {
-      const plugin = loadAllPlugins().find((p) => p.id === "cg-grants-gov");
-      const costSharing = plugin?.resolvedFilters.find(
-        (f) => f.name === "costSharing",
-      );
-
-      expect(costSharing?.filterType).toBe("booleanComparison");
-      expect(costSharing?.docsHref).toBe(
-        "/protocol/filters/boolean#booleancomparisonfilter",
-      );
-    });
   });
 });
--- a/website/src/content/docs/governance/adr/0023-org-ids.mdx
+++ b/website/src/content/docs/governance/adr/0023-org-ids.mdx
@@ -711,11 +711,11 @@ All three options below share the same principle: registry-level facts (the same
 | ---------------------------------------------- | :------------------: | :--: | :--------------: |
 | Explicit pointer to the catalog entry          |          ✅          |  ✅  |        ✅        |
 | Registry vs. instance fields visually grouped  |          ✅          |  ❌  |        ✅        |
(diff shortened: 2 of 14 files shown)


website/__tests__/lib/plugins/loader.spec.ts
```diff
@@ -30,16 +30,18 @@ describe("plugins loader", () => {
...
-    // The one filter type with no protocol model of its own, so it is the one
-    // whose href comes from the fallback rather than the anchor map.
-    it("sends a booleanComparison filter to the filters index", () => {
+    // The newest filter model, and the one a plugin declares against a boolean
+    // custom field, so it is the likeliest to lose its docs link in a rewire.
+    it("links a booleanComparison filter to the boolean filter model", () => {
       const plugin = loadAllPlugins().find((p) => p.id === "cg-grants-gov");
       const costSharing = plugin?.resolvedFilters.find(
         (f) => f.name === "costSharing",
       );
```
widal001: I feel like this could probably be dropped altogether if the only reason we had this test in the past was because it went to the index page

website/src/content/docs/governance/adr/0023-org-ids.mdx
widal001, line 1: Thanks for updating this!

widal001 (approved):
Looks good! I left one small comment, but don't let it block this merge.

website/__tests__/lib/plugins/loader.spec.ts
SnowboardTechie: good call, I went ahead and dropped

PR #1084, opened 2026-08-07
chore(deps): bump js-yaml to 4.3.1 for GHSA-5p4m-2wfm-xmqj

### LINKED ISSUE #1083: [Deps] Fix `js-yaml` vulnerability
### Summary

Fix the `js-yaml` vulnerability that is failing our CI dependency audit checks

### Acceptance criteria

- [ ] `js-yaml` issue is resolved
- [ ] A new patch version is published of each library that was affected by this vulnerability

widal001:
### Summary

Resolves a high-severity advisory for `js-yaml` which is failing other PRs.

- Fixes #1083 
- Time to review: 2 minutes

- Bumps `js-yaml` in:
  - Website
  - TS SDK `@common-grants/sdk`
  - CLI `@common-grants/cli`
- Adds patch release for:
  - TS SDK `@common-grants/sdk`
  - CLI `@common-grants/cli`

####  Notes

- This vulnerability is failing CI checks on a few existing PRs, like #1082 
- If we want to fix this in CI checks, but not yet publish a patch release, we could remove the changesets

See [this check for current failure](https://github.com/HHS/simpler-grants-protocol/actions/runs/31183528270/job/92882346762?pr=1082) and see the CI checks below for the new state.


.changeset/bump-js-yaml-4-3-1.md
```diff
@@ -0,0 +1,6 @@
+---
+"@common-grants/cli": patch
+"@common-grants/sdk": patch
```
karinamzalez: Still developing my intuition re: changesets -- do we need the SDK changeset here? Looks like in the SDK, js-yaml is only a dev dependency and one test helper uses it. So nothing actually changes for people who install @common-grants/sdk. Just curious!

pnpm-workspace.yaml
```diff
@@ -51,7 +51,7 @@ overrides:
   fast-uri@>=3.0.0 <3.1.5: '>=3.1.5 <4'
   form-data@>=4.0.0 <4.0.6: '>=4.0.6'
   immutable@<4.3.9: '>=4.3.9 <5'
-  js-yaml@>=4.0.0 <4.3.0: '>=4.3.0'
```
karinamzalez, line 54: Should this override have an upper bound? Asking bc dependencies right above provide a cap (-- e.g. fast-uri: >=3.1.5 <4). Is the difference intentional, or should this be >=4.3.1 <5 so a breaking v5 can't sneak in later?

karinamzalez (approved):
I added some non-blocking comments!

.changeset/bump-js-yaml-4-3-1.md
widal001: Thank you for catching this!! That's what I get for trying to open a PR while I'm in a meeting 🤦 

You're totally right that we don't need a patch release for the SDK, I removed it in this commit: [chore(core): Remove TypeScript SDK from changeset](https://github.com/HHS/simpler-grants-protocol/pull/1084/commits/fdc775f677748bc4b345d9d57bfcf69597ef071a)

pnpm-workspace.yaml
widal001: This is a good question, and based on the initial research that I did, here's what I found:

- By default running `pnpm audit --fix` adds the `>=X.X.X` without the limit
- Whether the limit makes a difference in terms of dependency resolution isn't totally clear. There is a v5.2.3 of `js-yaml` published, but transitive dependencies that had `^4.1.1` didn't get bumped to that version.
- It seems like there _may_ be a risk of that happening if we were to completely delete and recreate the `pnpm-lock.yml` but I was having trouble re-creating it myself though.

More importantly though, all of my testing made me realize we actually don't need the override at all though, so I removed it here: [build: removes unnecessary override](https://github.com/HHS/simpler-grants-protocol/pull/1084/commits/82e3877ec2782853a3da1de1aa343058c6630044)

I created [this ticket](https://github.com/HHS/simpler-grants-protocol/issues/1085) for follow-up investigation and cleanup though

PR #1087, opened 2026-08-10
ci(website): update PR preview workflow

### LINKED ISSUE #1079: [Mock API] Update website preview workflow
### Summary

Update the website preview workflow to:

- Use the Agile Six Cloudflare account, instead of Billy's personal account
- Use `wrangler versions upload` instead of a deploying a new worker, so PR previews are persistent (up to 100 previous versions)

### Acceptance criteria

- [ ] PR previews are published to Agile Six's Cloudflare instance
- [ ] We can remove the cleanup stage without having lingering workers in Cloudflare
- [ ] PR comment displays the most recent commit the preview was built from

widal001:
### Summary

Updates `ci-website-preview.yml` to use `versions upload` instead of creating a new worker for each PR.

- Fixes #1079 
- Time to review: 2 minutes

- Changes `ci-website-preview.yml` to use `wrangler versions upload` instead of `wrangler deploy`, which:
  - uploads new version to the same worker with a `--preview-alias` instead of creating a separate worker
  - preserves live previews after the PR closes (until 100 subsequent versions are uploaded)
  - removes the need for a separate cleanup step
- Deletes the unnecessary cleanup workflow `ci-website-cleanup.yml`
- Changes `name` attribute in `wrangler.jsonc` to `common-grants` since previews are just versions on an existing worker
- Fixes transitive dependency advisory `nanoid`

- PR previews will only be live for the previous 1000 aliases uploaded, so roughly 1000 PRs.
- If the url of the Agile Six worker (`brian-derfer`) gets updated (e.g. to `agile-six` or `a6`) the comment will automatically get updated when CI runs again, but comments on previously closed PRs that have `brian-derfer` will 404 unless they are updated to the new URL)

Here's the new [GitHub PR preview comment](https://github.com/HHS/simpler-grants-protocol/pull/1087#issuecomment-5240899962)

Screenshot 2026-08-10 at 9 48 48 AM

You can also see that the [previous PR is closed](https://github.com/HHS/simpler-grants-protocol/pull/1082#issuecomment-5217783315), but the [preview link](https://cg-pr-1084.brian-derfer.workers.dev/protocol/api-docs/) still works.

And we only have a single `common-grants` worker:
Screenshot 2026-08-10 at 9 54 14 AM

Instead of a separate worker for each PR:
Screenshot 2026-08-10 at 9 54 44 AM


pnpm-workspace.yaml
```diff
@@ -52,6 +52,7 @@ overrides:
   form-data@>=4.0.0 <4.0.6: '>=4.0.6'
   immutable@<4.3.9: '>=4.3.9 <5'
   lodash@>=4.0.0 <=4.17.23: '>=4.18.0'
+  nanoid@<3.3.17: '>=3.3.17 <4'
```
widal001, line 55: I intentionally added an upper range to this override

website/wrangler.jsonc
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "common-grants-preview",
+  "name": "common-grants",
```
widal001, line 2: Renamed this since previews will be uploaded as versions on the same worker vs. separate workers.

This also allows us to use the same worker for the main site if we decide to move off of GitHub pages

.github/workflows/ci-website-preview.yml
```diff
@@ -114,35 +111,62 @@ jobs:
...
+            dlx wrangler@4 versions upload \
+            --preview-alias "pr-${PR_NUMBER}" \
+            --message "PR #${PR_NUMBER} @ ${HEAD_SHA}" 2>&1 | tee upload.log
+          url=$(grep -E 'Version Preview Alias URL:' upload.log \
+            | sed -E 's/.*URL:[[:space:]]*//' \
+            | head -n1 || true)
+          if [ -z "$url" ]; then
+            echo "::warning::Could not extract preview URL from wrangler output"
+          fi
+          echo "url=${url}" >> "$GITHUB_OUTPUT"
```
widal001, line 149: This looks more complicated than the previous version, but the main reason for this complexity is that we're dynamically parsing the URL vs. using a fixed URL.

I've used this pattern on a few other repo with the same preview workflow.

karinamzalez: A couple of follow-ups here:

- If wrangler ever reworded that log line (`Version Preview Alias URL`) -- this grep wouldn't find anything and the check still goes green. Would it be possible to fail this loudly to account for that?
- Since we're using this parsing pattern in a few other repos.. is it worth pulling into something shareable (a composite action, or a script under .github/scripts/ alongside audit-deps.js?). Not urgen, but it'd mean one place to fix if wrangler's output changes.

.github/workflows/ci-website-preview.yml
```diff
@@ -114,35 +111,62 @@ jobs:
...
 
       - name: Find existing comment
         id: find-comment
+        if: steps.upload.outputs.url != ''
```
karinamzalez: similar silent error concern here if URL can't be parsed.

karinamzalez (approved):
Approved with a few non-blocking questions

.github/workflows/ci-website-preview.yml
widal001: Good questions! I made a few changes in this commit [ci(website): update preview script](https://github.com/HHS/simpler-grants-protocol/pull/1087/commits/55b6276ac1914f1e54782d7879b7d61152557b89)

- Moved this `run` block to `.github/scripts/upload-worker-preview.sh`
- Updated the script to have a clearly defined `LABEL` so we can see what's being passed into `grep` more explicitly
- Set it to `exit 1` if a match isn't found, with an error message

On the composite action suggestion specifically, I'm not sure the juice would be worth the squeeze. Maybe at some point in the future, but keeping this a 50ish line bash script feels like a good compromise for now.

.github/workflows/ci-website-preview.yml
widal001: I removed this after updating the script to exit with code 1 if a URL isn't parsed, so it will fail on the upload step if a URL isn't parsed.

.github/workflows/ci-website-preview.yml
karinamzalez: I've never heard that juice and squeeze metaphor in this context and I love it lol -- but I agree! 

two small things:
- The --allow-build=workerd --allow-build=esbuild --allow-build=sharp flags went away along with the comment explaining them. Was that intentional?
- do we want to git ignore upload.log?

widal001: - Yes, dropping those was intentional. When I was moving this over to the separate script I realized we don't need to run the post-install script for those dependencies to use `versions upload` and adding them to `--allow-build` slightly increases our surface area for supply chain attacks (the post-install scripts were part of the attack vector for shai hulud) 
- Any files created during this step are disposed after CI runs, but doesn't hurt to add it, so I ignored it here: [chore(website): add upload.sh to .gitignore](https://github.com/HHS/simpler-grants-protocol/pull/1087/commits/816c789429c38f0418ce5f22fa60522e3b313862)

PR #1115, opened 2026-08-21
ci: publish npm packages through trusted publishing

### LINKED ISSUE #765: [Dependency Mgmt] Set up trusted publishing for CommonGrants npm packages
### Summary

Configure trusted publishing for CommonGrants npm packages to eliminate the need to rotate npm tokens stored in GitHub Actions. Trusted publishing lets GitHub Actions authenticate to npm without long-lived tokens. May require GitHub org admin approval. PyPI trusted publishing for common-grants-sdk is out of scope here and will be tracked separately (see #650).

### Acceptance criteria

- [x] Trusted publishing configured for CommonGrants npm packages
- [ ] Token-based npm auth removed from GitHub Actions workflows
- [ ] Required GitHub org admin approvals secured if needed
- [ ] Publishing of npm packages is verified for all packages switched to trusted publishing
- [ ] Setup team Lunch and Learn to share Release Please learnings

### References

- Related umbrella: #567 [Release Mgmt] Security improvements (Q7)

SnowboardTechie:
### Summary

Refs #765

Replaces long-lived npm token authentication with GitHub Actions OIDC trusted publishing for:

- `@common-grants/core`
- `@common-grants/cli`
- `@common-grants/sdk`

Automated releases and manual publish retries for every package (npm and PyPI) run from one workflow, `cd-release.yml`. npm binds each trusted publisher to the filename of the workflow that publishes, so a single entry point lets both paths share one registration.

PyPI authentication is unchanged (`PYPI_TOKEN`); only the workflow that runs it moved.

### Changes

- Removes `NPM_TOKEN` and `NODE_AUTH_TOKEN` from the npm publishing and dry-run workflows.
- Publishes with native `pnpm publish --provenance --access public --no-git-checks`. pnpm 11 performs the OIDC exchange itself (fixed in pnpm 11.1.3 and `pnpm/action-setup` 6.0.6; the repo pins 11.20.0), so no separate npm client or Node 24 runtime is needed, and pnpm resolves `catalog:` specs as it always did.
- Renames `release-please.yml` to `cd-release.yml`. The release-please step runs only on `push`; a `plan` step then emits the packages to publish (on push: what release-please just released; on dispatch: the selected package and tag).
- Replaces the three per-package publish jobs with one `publish-npm` matrix job driven by that plan, each run in the package's own environment (`npm-core`, `npm-cli`, `npm-sdk`) with `id-token: write`. Tag validation and the `refs/tags/` checkout are preserved.
- Moves the Python SDK publish inline into the same workflow and adds `common-grants-sdk` to the dispatch choices, so one retry procedure covers all four packages.
- Deletes `cd-deploy-lib-core.yml`, `cd-deploy-lib-cli.yml`, `cd-deploy-lib-ts-sdk.yml`, and `cd-deploy-lib-pysdk.yml`.
- Returns the PR dry runs to `pnpm publish --dry-run` on the repository's `.nvmrc` Node baseline.
- Updates `lib/README.md` and the three `DEVELOPMENT.md` release checklists for the new filename and retry procedure.

### npm configuration

The trusted-publisher settings for all three packages:

- Workflow: `cd-release.yml`
- Existing package-specific environment: `npm-core`, `npm-cli`, or `npm-sdk`
- Allowed action: `npm publish`

### Verification

- `actionlint` passes on the changed workflows.
- The `plan` step was checked against: no release, one npm release, npm + Python releases, Python only, dispatch of each package, and dispatch of an unknown package name.
- All hosted checks pass, including the three `pnpm publish --dry-run` workflows.

### Remaining verification

This PR does not publish a package, so it cannot prove the live OIDC exchange with pnpm 11's native publish. Issue #765 should remain open after merge until core, CLI, and SDK have each published successfully through trusted publishing and their npm provenance has been verified.


.github/workflows/cd-deploy-npm.yml
widal001, line 1: Is there a reason we're using this instead of just running `pnpm publish` directly in `release-please.yml`?

If this was an attempt to address the previous issues with OIDC trusted publishing and pnpm 11, it looks like those issues have been addressed as long as we're using the latest pnpm 11 version and the right `pnpm/action-setup`

- https://github.com/pnpm/pnpm/issues/9812
- https://github.com/pnpm/pnpm/issues/11566#issuecomment-4437507118
- https://github.com/pnpm/pnpm/issues/11513#issuecomment-4404990486

.github/workflows/ci-lib-cli-publish.yml
widal001, line 1: I think these changes and the equivalent changes to the other dry run publish steps could go away if we switch back to `pnpm publish`

.github/workflows/release-please.yml
widal001, line 1: Can we also change this to `cd-release.yml` or something similar since it's triggering the npm deployment?

.github/workflows/release-please.yml
```diff
@@ -30,35 +48,72 @@ jobs:
       py_sdk_released: ${{ steps.release.outputs['lib/python-sdk--release_created'] }}
       py_sdk_tag: ${{ steps.release.outputs['lib/python-sdk--tag_name'] }}
 
-  # Publishes fan out from the outputs above. Tags and releases created with
-  # GITHUB_TOKEN do not trigger other workflows, so the deploys must be called
-  # from this run rather than via `on: release`.
+  # GITHUB_TOKEN tags do not trigger workflows, so deploys are called here.
+  # !cancelled() passes the skipped release job on retry; result blocks a failed
+  # push. Keep dispatch choices aligned with predicates. Callers grant OIDC
+  # because reusable workflows cannot elevate permissions.
```
widal001: Can we try to rephrase these docs more plainly? I'm not really clear on what this means.

.github/workflows/release-please.yml
```diff
@@ -30,35 +48,72 @@ jobs:
...
-    uses: ./.github/workflows/cd-deploy-lib-core.yml
+    if: |
+      !cancelled()
+      && (
+        (github.event_name == 'push'
+          && needs.release-please.result == 'success'
+          && needs.release-please.outputs.core_released == 'true')
+        || (github.event_name == 'workflow_dispatch'
+          && inputs.package == '@common-grants/core')
+      )
```
widal001: It seems like this chunk is repeated across each step with a different `inputs.package == '*'`, and has a lot of nested `&&` and `||` is there any way to centralize it or make it easier to reason about?

Like maybe there's just a step that outputs which package (if any) needs to be published and then this logic is replaced by a check to see whether that output includes `@common-grants/core` (and the other package names below)

lib/README.md
```diff
@@ -75,7 +75,9 @@ Merging a package's Release PR:
 
 ### Re-running a failed publish
 
-The per-package deploy workflows (`cd-deploy-lib-*.yml`) keep a `workflow_dispatch` trigger. To retry a publish for an existing tag, run the matching workflow from the **Actions** tab and enter the release tag.
+- **npm retry:** Dispatch `release-please.yml` with the package and tag. This publishes only; npm authenticates its caller.
+- **Release Please retry:** Re-run the failed Action or merge another commit to `main`.
+- **Python SDK retry:** Dispatch `cd-deploy-lib-pysdk.yml`; PyPI uses a repository secret.
```
widal001: I know the main reason we're consolidating the npm packages into `release-please.yml` is for OIDC, which isn't necessary for PyPI, but now it seems a tad confusing to have a different deploy workflow for the one python package.

Is it worth consolidating `cd-deploy-lib-python.yml` (or whatever it's called) into the release please workflow for consistency?

widal001 (commented):
See my comments below for the following key questions/suggestions:

- Removing `cd-deploy-npm.yml` and reverting `ci-lib-*-publish.yml` after switching pack to `pnpm publish` it should work with OIDC if we're using the right pnpm version and `pnpm/action-setup`. We can also test this on [`common-grants/ts-cg-grants-gov`](https://github.com/common-grants/ts-cg-grants-gov) if we want to be extra safe
- Consider consolidating the Python SDK publishing step into the release please action for consistency, so we don't have two different publishing workflows.
- Consider renaming the workflow `cd-release.yml` since it's kicking off deploys, and combining the repetitive condition checks.

.github/workflows/cd-deploy-npm.yml
SnowboardTechie: Yeah, that was the reason. When I set this up I was thinking pnpm 11's native publish couldn't do the OIDC exchange, so I packed with pnpm and published with npm, and the reusable workflow was just to avoid copying those steps three times. Not sure how I missed that fix in 11.1.3 (we're on 11.20.0) it's now plain pnpm publish --provenance in cd-release.yml and cd-deploy-npm.yml is gone.

.github/workflows/ci-lib-cli-publish.yml
SnowboardTechie: Yeah, these are back to `pnpm publish --access public --dry-run --no-git-checks`

.github/workflows/release-please.yml
SnowboardTechie: Yeah, there is. I added a step in the release-please job that outputs the list of packages to publish (on push, whatever release-please just released; on dispatch, the one package selected). The three publish jobs collapsed into a single job that reads that list, so the nested `&&`/`||` checks are gone entirely. The `!cancelled()` guard went away too, since the release-please step is now skipped on dispatch instead of the whole job.

lib/README.md
SnowboardTechie: Yeah, good idea. The Python publish job is inline in `cd-release.yml` now and `cd-deploy-lib-pysdk.yml` is gone, so all four packages release and retry from the same workflow.

.github/workflows/cd-release.yml
```diff
@@ -0,0 +1,180 @@
...
+      - name: Install Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: .nvmrc
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run checks
```
karinamzalez, line 125: I'm wondering if we want to run checks towards the beginning of this publish-npm build step? So the job doesn't fail after the tag exists?

SnowboardTechie: Good question. Those checks do already run before the tag exists, just not in this file. The release PR that release-please opens goes through normal PR CI, and the version bump hits the `lib/<pkg>/**` path filter, so ci-lib-* runs the same `checks` on the exact content that gets tagged when it merges (#1136 is a recent example). This step is a second pass right before publish, which is the one step we can't take back. If something transient trips it here, we can manually retry the publish for that same tag from the Actions tab and it reruns from the top.

karinamzalez (approved):

PR #1117, opened 2026-08-24
ci: audit Release Please version PRs

### LINKED ISSUE #1151: Scheduled dependency audit is failing
The daily workspace-wide dependency audit found advisories against deps on `main`. See the run log for details: https://github.com/HHS/simpler-grants-protocol/actions/runs/33647223758

SnowboardTechie:
### Summary

- Fixes: N/A — no linked issue
- Time to review: 5 minutes

### Changes proposed

- Block the release run — tag creation, GitHub release, and the npm/PyPI publish fan-out — on a passing workspace-wide `pnpm audit`.
- Run that same audit on every Release Please version PR, so an advisory surfaces while a reviewer is looking at the PR rather than at publish time.
- Document both in `DEPENDENCY_MANAGEMENT.md` as the production dependency-audit checkpoint.

### Context for reviewers

Two gates, at two different moments.

`cd-release.yml` gains a `dependency-audit` job that the `release-please` job now `needs`. Release Please creates tags, GitHub releases, and the publish fan-out within a single run, so the audit has to pass before that run starts — there is no later point that gates the publish without the tag and release already existing.

`deps-audit.yml` gains `.release-please-manifest.json` in its `pull_request` path filter. Release Please runs in manifest mode with separate PRs for the four published packages, and every version PR updates that file, so it is a stable package-independent trigger. This is the review-time signal; the `cd-release.yml` job is what actually blocks the publish.

The audit command and the low-severity threshold are unchanged from the daily `main` sweep.

### One open question

**Should the gate sit on `release-please` or on the publish jobs?** In `cd-release.yml` the `release-please` job also runs on `workflow_dispatch` — only the release-please *step* is gated on `push`; the job still runs to compute the publish plan. So `needs: dependency-audit` now also blocks a manual republish, which is the break-glass retry path, and it audits the default branch while a republish checks out `refs/tags/<tag>`. I kept the original placement rather than redesign it during the restack; moving `needs` to `publish-npm` / `publish-py-sdk` would gate only the publish and leave dispatch usable.

### Additional information

The first merged version PR, #1092, confirms the trigger premise: it updated `.release-please-manifest.json`, and its head commit carries a `lint-pr-title` check run from `ci-pr-title.yml`, which is `pull_request`-only. Version PRs do start `pull_request` workflows, so the new path filter will fire.

`main`'s ruleset requires one approving review but sets no required status checks, so the `deps-audit.yml` run on a version PR is an advisory signal only — `needs: dependency-audit` is the sole enforcement in this PR.

This was originally stacked on #1115, which has since merged. `main` has been merged into this branch and the diff is only its own three files.

The second open question this PR carried is resolved. `pnpm audit` was red on `main` when it was opened — high-severity `fast-uri` advisories reaching in through `@typespec/compiler > ajv`, tracked in #1151. #1161 raised the override floors and closed #1151, so the gate no longer risks stalling Release Please on merge.

Verified on this head: YAML parses and the job graph resolves as `dependency-audit → release-please → publish-npm / publish-py-sdk`; `git diff --check` clean; `pnpm audit` reports no known vulnerabilities.


karinamzalez (changes requested):
Lgtm! Looks like it worked and is blocking in CI? 😅

SnowboardTechie:
@karinamzalez ya, that was dependency audits failing but now that #1161 is merged and this has the latest main CI is all green ✅

karinamzalez (approved):

PR #1118, opened 2026-08-25
fix(sdk): stop duplicating baseUrl path prefix in Client.get query params

### LINKED ISSUE #1112: [TS SDK] Client.get() re-applies baseUrl to requests with query params, breaking list()
### Summary

`client.opportunities.list()` returns 404 when `baseUrl` carries a path prefix.
In `Client.get()`, the query-param branch builds `fullPath = url.pathname +
url.search` (`lib/ts-sdk/src/client/client.ts:158`), which already includes the
base path, and then `fetch()` applies `baseUrl` again. A request to
`https://host/api/v0.4.0/common-grants/opportunities?page=1` lands on
`/api/v0.4.0/api/v0.4.0/common-grants/opportunities?page=1`.

This is invisible when `baseUrl` is a bare origin, which is why existing tests
miss it. It breaks anything sending query params; `get()` and `search()` are
unaffected. Found during experiment 3B (#1078) and written up in
`website/MOCK-FINDINGS.md` on branch `karina/1078-astro-endpoint-mock`.

### Acceptance criteria

- [ ] `Client.get()` sends query-param requests to the same path as requests without them
- [ ] `opportunities.list()` works against a `baseUrl` that includes a path prefix
- [ ] Regression test covers a `baseUrl` with a path prefix

karinamzalez:
### Summary

- Fixes a 404-causing URL bug in the TS SDK's `Client.get()`
- Time to review: 5 minutes

### Changes proposed

- lib/ts-sdk/src/client/client.ts: Client.get() was assembling the request address the wrong way-- folders in path were getting duplicated. With a base URL like `http://host/api/v0.4.0` (the documented way to point the SDK at the CommonGrants mock API), any GET with query params ended up asking for `/api/v0.4.0/api/v0.4.0/common-grants/opportunities` -- an address that doesn't exist, so it 404'd.

- `lib/ts-sdk/__tests__/client/client.spec.ts`: two regression tests: one asserting a path-carrying baseUrl produces a single (non-duplicated) path prefix when params are passed, and one covering the merge of params into a path that already has a query string.

### Context for reviewers

- Repro before the fix: `CG_BASE_URL=http://localhost:4322/api/v0.4.0 pnpm --filter @common-grants/sdk example:list` against a running mock 404s. `example:get` and `example:search` were unaffected because they don't pass query params through `GetOptions.params`.
- Verified the path-prefix regression test fails against the pre-fix code (request hits the doubled path and 404s) and passes with the fix.
- `pnpm --filter @common-grants/sdk run ci` is fully green: 27 test files, 550 tests passed, plus compile and examples steps.

### Additional information

Before: `GET https://host/api/v0.4.0/api/v0.4.0/common-grants/opportunities?page=1` → 404
After: `GET https://host/api/v0.4.0/common-grants/opportunities?page=1` → 200

### DIFF
--- a/lib/ts-sdk/src/client/client.ts
+++ b/lib/ts-sdk/src/client/client.ts
@@ -149,13 +149,17 @@ export class Client {
   async get(path: string, options?: GetOptions): Promise<Response> {
     let fullPath = path;
 
-    // Append query params if provided
+    // Append query params if provided. Build the query string from the path
+    // alone — extracting it from the full URL would fold the baseUrl's path
+    // prefix (e.g. "/api/v0.4.0") into fullPath, duplicating it in fetch().
     if (options?.params && Object.keys(options.params).length > 0) {
-      const url = new URL(this.url(path));
+      const queryIndex = path.indexOf("?");
+      const pathname = queryIndex === -1 ? path : path.slice(0, queryIndex);
+      const searchParams = new URLSearchParams(queryIndex === -1 ? "" : path.slice(queryIndex + 1));
       for (const [key, value] of Object.entries(options.params)) {
-        url.searchParams.set(key, String(value));
+        searchParams.set(key, String(value));
       }
-      fullPath = url.pathname + url.search;
+      fullPath = `${pathname}?${searchParams.toString()}`;
     }
 
     return this.fetch(fullPath, {
--- a/lib/ts-sdk/src/client/client.ts
+++ b/lib/ts-sdk/src/client/client.ts
@@ -150,8 +150,7 @@ export class Client {
     let fullPath = path;
 
     // Append query params if provided. Build the query string from the path
-    // alone — extracting it from the full URL would fold the baseUrl's path
-    // prefix (e.g. "/api/v0.4.0") into fullPath, duplicating it in fetch().
+    // alone: fetch() prepends the baseUrl, so it must not leak into fullPath.
     if (options?.params && Object.keys(options.params).length > 0) {
       const queryIndex = path.indexOf("?");
       const pathname = queryIndex === -1 ? path : path.slice(0, queryIndex);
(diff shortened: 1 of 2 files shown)


lib/ts-sdk/src/client/client.ts
```diff
@@ -149,13 +149,16 @@ export class Client {
...
-      const url = new URL(this.url(path));
+      const queryIndex = path.indexOf("?");
+      const pathname = queryIndex === -1 ? path : path.slice(0, queryIndex);
+      const searchParams = new URLSearchParams(queryIndex === -1 ? "" : path.slice(queryIndex + 1));
       for (const [key, value] of Object.entries(options.params)) {
-        url.searchParams.set(key, String(value));
+        searchParams.set(key, String(value));
       }
-      fullPath = url.pathname + url.search;
+      fullPath = `${pathname}?${searchParams.toString()}`;
```
widal001: Thanks for catching this bug!

I see the value of the slice-based approach especially to avoid accidentally pre-fixing the `path` with `baseUrl` but it seems a tad harder to reason about at a glance and I get a little nervous with all of the slicing and the `+ 1` and `-1` which feels ripe for off-by-one errors with future edits.

Did you consider simply changing?

```ts
const url = new URL(this.url(path));
```

To something like this:
```ts
// Uses a placeholder to parse path into a pathname and
// existing query params without pre-pending baseUrl twice
const url = new URL(path, "http://placeholder");
```

I think that would have the same effect without introducing manual string slicing and would effectively reduce this to a 3-line fix with comments

karinamzalez: Thanks @widal001 -- this is way simpler haha. Done!

widal001 (approved):
Thanks for finding and fixing this bug!

SnowboardTechie (approved):

Issue #1078, opened 2026-08-06
[Mock API] Experiment: mock endpoint on the website via Cloudflare (ADR Option 3B)


karinamzalez:
### Summary

Build a throwaway, timeboxed experiment of the **integrated endpoint** shape from the
mock playground ADR (Option 3B): the mock API served by the docs website itself,
same origin as the docs. Companion to the Option 3A experiment (#1077), judged
against the same rubric.

**Status of 3A (baseline to beat):** the standalone Worker works end-to-end — deployed
at a `*.workers.dev` URL, serving the three opportunity endpoints for spec versions
0.1.0–0.4.0, with CI conformance tests against the generated per-version schemas.
Its measured surface: **21 package files + 1 deploy workflow + 3 root-config edits**.
The hypothesis to test: the website's existing harness (test setup, lint/format,
preview deploy) absorbs most of that, so the integrated shape may cost only ~5–6
ported files.

**Updates from the 2026-08-11 sync** (supersede the original description):

- **Data layer = port the proven kernel from 3A**, not the website's
  `example-generator.ts` / `openapi-sampler` path. The ADR's follow-up section
  already established that the schema-sampling layer has no records and can't meet
  the determinism / cross-endpoint-consistency criteria. Port as-is:
  `fixtures.ts` (11 deterministic records), the opportunities handlers
  (filter/sort/pagination/error logic), and the envelope/CORS helpers.
- **Do not flip the site to SSR.** Keep `output: "static"` in `astro.config.mjs`;
  add the `@astrojs/cloudflare` adapter and set `export const prerender = false`
  on the API route(s) only. Everything else on the site stays static.
- **The 3A PR stays open as-is.** It is the fallback if this experiment hits a
  wall; the deliverable of this issue is the comparison, not merged code.

Scope: opportunity endpoints only (list / detail / search), all four versions via
the same path-prefix routing the Worker uses. Non-goals: other resources,
productionizing, and the actual GH Pages → Cloudflare production migration (this
experiment runs on PR previews; the findings must *size* the migration, not do it).

### Acceptance criteria

- [ ] `@astrojs/cloudflare` adapter added with `output: "static"` retained; only the
      API route is dynamic (`prerender = false`); existing docs pages still render
      correctly on a Cloudflare PR preview (including the #1087 preview flow:
      `create-worker-preview.sh` + `wrangler versions upload`)
- [ ] The three opportunity endpoints are served same-origin from the website
[...]

PR #1135, opened 2026-08-27
ci(website): add parallel Cloudflare production deploy on merge to main

### LINKED ISSUE #1110: [Mock API] Deploy the beta CommonGrants site to Cloudflare
### Summary

Stand up a parallel production deployment of commongrants.org on the beta
production Cloudflare Worker. The beta runs alongside the current production
site until the playground works on it, then it becomes the official production
site.

### Acceptance criteria

- [ ] Beta site deploys to Cloudflare on its own URL, separate from production
- [ ] Mock API playground works on the beta site
- [ ] Cutover path from beta to production is written down

karinamzalez:
### Summary

- Part of #1110 (deliberately not `Fixes` — this is the parallel-deploy setup only;
  verification, the #1078 merge-safety work, and the cutover doc follow)
- Time to review: 10 minutes

### Changes proposed

One file changed: `.github/workflows/cd-deploy-website.yml`.

- **Added a `deploy-cloudflare` job** that runs on every push to `main` touching
  `website/**` (and on `workflow_dispatch`), alongside the existing GitHub Pages
  deploy. It downloads the shared build artifact and runs `pnpm dlx wrangler@4 deploy`
  from `website/`, which publishes a new version of the `common-grants` Worker and
  switches live traffic to it-- i.e. beta.commongrants.org starts tracking `main`.
- **Added an `Upload build output` step** to the existing `build` job, mirroring the
  artifact pattern from `ci-website-preview.yml`, so both deploy targets consume one shared build.
- **Renamed** the `deploy` job to `deploy-github-pages` and the workflow to
  "CD - Deploy Website" (behavior unchanged; nothing in the repo references the old names).
- **Added the workflow file itself to the push paths filter**, so workflow edits
  (including this PR when it merges) trigger a deploy run.
- Guardrails on the new job: `permissions: contents: read`, a `concurrency` group with
  `cancel-in-progress: false` so deploys never race or get killed mid-flight, and the
  same `main`-ref guard as the Pages deploy.

### Context for reviewers

**Why:**  we don't want to swap the production host and cut DNS in one move. This PR is the first step of the migration plan: every merge to `main` deploys to *both* GitHub Pages (still production, DNS untouched) and the Cloudflare Worker (beta.commongrants.org). Once a few deploys flow through cleanly and beta matches production, the DNS cutover is just a record change. It also deliberately lands *before* the #1078 SSR/mock-API work, so we're only changing one thing (the hosting pipeline) at a time.

**Design choices worth a look:**
- The two deploy jobs both `needs: build` but have no edge between each other-- a
  Cloudflare failure can't block the Pages deploy, and vice versa.
- One shared build artifact rather than a second build: parity between the two targets
  is the point of this phase. `MOCK_API_ENABLED` is deliberately not set. 
- Reused the exact artifact upload/download pattern from `ci-website-preview.yml`-- same `website-build` name won't collide since the two workflows never run on the same event.
- `wrangler deploy` (not `versions upload`) -- PR previews stay as aliased, undeployed
  versions of the same Worker; this job is the only thing that moves live traffic.
- The existing `upload-worker-preview.sh` wasn't reusable: it's hardwired to `versions upload --preview-alias` and greps the preview-URL line that `wrangler deploy` never prints, so the deploy is a plain inline step.

**How it was verified:** the workflow YAML parses clean; `wrangler deploy --dry-run`
validates the config/credential path locally; the deploy command uses the same
checked-in `wrangler.jsonc` and repo secrets (`CLOUDFLARE_API_TOKEN` /
`CLOUDFLARE_ACCOUNT_ID`) the PR-preview flow already exercises daily. The real proof is the first post-merge run-- I've created a follow up ticket with  a short parity checklist (homepage, docs, redirects, OpenAPI/schema assets) between beta.commongrants.org and commongrants.org over the next 2–3 merges.

**Heads-up on the first run:** the first `deploy-cloudflare` run replaces whatever Worker version is currently live on beta.commongrants.org.

### Additional information

Job graph after this PR:

```
build ──┬──> deploy-github-pages   (production, commongrants.org — unchanged)
        └──> deploy-cloudflare     (beta, beta.commongrants.org — new)
```

Migration sequence this PR belongs to (#1110): parallel deploys (this PR) →
2–3 verified deploy cycles → #1078 merge made safe for both targets →
documented DNS cutover.

### DIFF
--- a/.github/workflows/cd-deploy-website.yml
+++ b/.github/workflows/cd-deploy-website.yml
@@ -105,9 +105,7 @@ jobs:
 
       - name: Deploy to Cloudflare Worker
         # Uploads a new version of the Worker named in website/wrangler.jsonc
-        # (common-grants) and switches live traffic to it. PR previews are
-        # versions of this same Worker, uploaded (not deployed) with a
-        # preview alias by .github/scripts/upload-worker-preview.sh.
+        # (common-grants) and switches live traffic to it. PR previews are versions of this same Worker
         working-directory: ./website
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
--- a/.github/workflows/cd-deploy-website.yml
+++ b/.github/workflows/cd-deploy-website.yml
@@ -11,6 +11,13 @@ on:
   # Allows us to run this workflow manually from the Actions tab on GitHub.
   workflow_dispatch:
 
+# Serialize whole runs (build + deploys) so a slow older build can't deploy
+# after a newer one. The default single pending slot skips stale intermediate
+# runs; only the newest queued run deploys.
+concurrency:
+  group: deploy-website-${{ github.ref }}
+  cancel-in-progress: false
+
 # Allow this job to clone the repo and create a page deployment
 permissions:
   contents: read
@@ -40,22 +47,22 @@ jobs:
         run: pnpm --filter website build
 
       - name: Upload Pages artifact
+        # Nonblocking so an upload failure for one deploy target
+        # doesn't take down the other one.
+        continue-on-error: true
         uses: actions/upload-pages-artifact@v5
         with:
           path: ./website/dist
 
       - name: Upload build output
         # Shared with deploy-cloudflare so both targets ship the same build.
+        # Nonblocking so an upload failure for one deploy target
+        # doesn't take down the other one.
+        continue-on-error: true
         uses: actions/upload-artifact@v7
         with:
           name: website-build
-          path: |
-            website/dist
-            website/tsp-output
-            website/.extension-schemas
-            website/cache
-            website/public/schemas/yaml
-            website/public/openapi
+          path: website/dist
           retention-days: 1
           if-no-files-found: error
 
@@ -81,9 +88,9 @@ jobs:
       contents: read
     # Only deploy from main branch
     if: github.ref == 'refs/heads/main'
-    concurrency:
-      group: deploy-cloudflare
-      cancel-in-progress: false
+    environment:
+      name: cloudflare
+      url: https://beta.commongrants.org
     steps:
       - name: Checkout Code
         uses: actions/checkout@v6
@@ -101,7 +108,7 @@ jobs:
         uses: actions/download-artifact@v8
         with:
           name: website-build
-          path: website
+          path: website/dist
 
       - name: Deploy to Cloudflare Worker
         # Uploads a new version of the Worker named in website/wrangler.jsonc
--- a/.github/workflows/cd-deploy-website.yml
+++ b/.github/workflows/cd-deploy-website.yml
@@ -1,4 +1,4 @@
-name: "CD - Deploy Website to GH Pages"
+name: "CD - Deploy Website"
 
 on:
   # Trigger the workflow every time we push to the `main` branch
@@ -7,6 +7,7 @@ on:
     branches: [main]
     paths:
       - "website/**"
+      - ".github/workflows/cd-deploy-website.yml"
   # Allows us to run this workflow manually from the Actions tab on GitHub.
   workflow_dispatch:
 
@@ -43,7 +44,27 @@ jobs:
         with:
           path: ./website/dist
 
-  deploy:
+      - name: Upload build output
+        # Shared with deploy-cloudflare so both targets ship the same build.
+        # Path-structure invariant (same as ci-website-preview.yml): every
+        # entry must stay under `website/` so upload-artifact strips that
+        # common prefix and the download (path: website) restores them at
+        # website/dist, etc. Adding a path outside website/ shrinks the
+        # prefix and silently shifts download locations.
+        uses: actions/upload-artifact@v7
+        with:
+          name: website-build
+          path: |
+            website/dist
+            website/tsp-output
+            website/.extension-schemas
+            website/cache
+            website/public/schemas/yaml
+            website/public/openapi
+          retention-days: 1
+          if-no-files-found: error
+
+  deploy-github-pages:
     needs: build
     runs-on: ubuntu-latest
     # Only deploy from main branch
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-deploy-website.yml
```diff
@@ -43,7 +44,22 @@ jobs:
         with:
           path: ./website/dist
 
-  deploy:
+      - name: Upload build output
```
SnowboardTechie, line 57: Would you consider allowing this step to fail without failing the whole `build` job? Since `deploy-github-pages` needs `build`, a transient artifact-upload failure—or `if-no-files-found: error`—would currently prevent the existing Pages deployment as well. Making the step nonblocking would preserve the intended isolation between the two deployment targets.

```suggestion
      - name: Upload build output
        continue-on-error: true
```

.github/workflows/cd-deploy-website.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches: [main]
     paths:
       - "website/**"
+      - ".github/workflows/cd-deploy-website.yml"
   # Allows us to run this workflow manually from the Actions tab on GitHub.
   workflow_dispatch:
```
SnowboardTechie, line 12: Would it make sense to establish concurrency at the workflow level instead? With job-level concurrency, a newer commit could finish building and deploy before an older, slower build, after which the older version could overwrite it. Adding `queue: max` would also preserve pending deployments rather than replacing them.

This would still allow the Pages and Cloudflare jobs to run in parallel within each workflow run.

```suggestion
  workflow_dispatch:

concurrency:
  group: deploy-website-${{ github.ref }}
  cancel-in-progress: false
  queue: max
```

.github/workflows/cd-deploy-website.yml
```diff
@@ -55,3 +71,43 @@ jobs:
...
+  deploy-cloudflare:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    # Only deploy from main branch
+    if: github.ref == 'refs/heads/main'
+    concurrency:
+      group: deploy-cloudflare
+      cancel-in-progress: false
```
SnowboardTechie: If concurrency is moved to the workflow level above, could we remove this job-level block?

```suggestion
```

.github/workflows/cd-deploy-website.yml
```diff
@@ -55,3 +71,43 @@ jobs:
...
+          path: website
+
+      - name: Deploy to Cloudflare Worker
+        # Uploads a new version of the Worker named in website/wrangler.jsonc
+        # (common-grants) and switches live traffic to it. PR previews are versions of this same Worker
+        working-directory: ./website
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+        run: pnpm dlx wrangler@4 deploy
```
SnowboardTechie, line 120: Nonblocking: would you consider pinning the Wrangler version used for the production deployment? The lockfile currently contains Wrangler `4.123.0`, while `pnpm dlx wrangler@4` resolves the latest v4 release at runtime. An exact version would keep deployment-tool changes visible in the repository history.

```suggestion
        run: pnpm dlx wrangler@4.123.0 deploy
```

SnowboardTechie (changes requested):
The parallel-deployment architecture looks sound: one build, independent GitHub Pages and Cloudflare deploy jobs, and a direct `wrangler deploy` rather than adding unnecessary abstraction to the preview script.

I found two deployment-safety issues that I think are worth addressing before merging:

1. Because the new Cloudflare artifact upload remains part of the `build` job, an upload failure could also prevent the existing GitHub Pages production deployment.
2. Concurrency begins only after each build finishes, which could allow deployments to occur out of commit order. The default single pending slot could also discard intermediate deployments.

I also left one nonblocking suggestion about pinning the Wrangler executable used for production deployments.

.github/workflows/cd-deploy-website.yml
widal001: I think keeping it to 4 is fine, we do that in the PR preview build already and I think it's more likely that we'd forget to change this when we bump the lockfile, which would defeat the purpose of pinning it to match exactly.

.github/workflows/cd-deploy-website.yml
widal001: We might also want to do the same thing with `continue-on-error` for the Upload pages artifact step to keep the two downstream deploy steps independent of failures in the other one.

.github/workflows/cd-deploy-website.yml
```diff
@@ -43,7 +44,22 @@ jobs:
...
+        uses: actions/upload-artifact@v7
+        with:
+          name: website-build
+          path: |
+            website/dist
+            website/tsp-output
+            website/.extension-schemas
+            website/cache
+            website/public/schemas/yaml
+            website/public/openapi
```
widal001: I think we can actually skip most of these and just preserve the `website/dist` since the others are just local inputs to the build or intermediate build outputs.

You'll want to make the corresponding change in the download.

.github/workflows/cd-deploy-website.yml
```diff
@@ -55,3 +71,43 @@ jobs:
...
+
+  # Parallel production deploy to the shared `common-grants` Cloudflare Worker
+  # (beta.commongrants.org) during the GH Pages deploy
+  deploy-cloudflare:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    # Only deploy from main branch
+    if: github.ref == 'refs/heads/main'
```
widal001, line 90: Can we add an `environment` here? The same way we have one for GitHub pages.

widal001 (commented):
Excited to get this in! I left a few comments in addition to @SnowboardTechie's

While the parallel deployments optimizations are good to have, this is designed to be a pretty short-lived dual deployment workflow -- just until we can safely cut over to Cloudflare completely.

The main change I'd like to see is adding a new `environment` for the Cloudflare deploys.

.github/workflows/cd-deploy-website.yml
karinamzalez: Added continue-on-error: true to both upload steps (Pages artifact and build output)

.github/workflows/cd-deploy-website.yml
karinamzalez: I moved concurrency to the workflow level! But re: `queue: max`, if several commits come in while a deploy is running, we only need to deploy the newest one.. I think the default behavior does that already (skips the in-between commits). This is also how GH Pages' workflow works. With queue: max, we'd deploy every in-between commit, and each would be replaced seconds later. I feel like that's extra CI time for no benefit. Unless there's a benefit there that I'm not seeing? 👀

.github/workflows/cd-deploy-website.yml
karinamzalez: removed!

.github/workflows/cd-deploy-website.yml
karinamzalez: Done!

.github/workflows/cd-deploy-website.yml
SnowboardTechie: You’re right. Moving concurrency to the workflow level addresses the stale-deployment risk I was concerned about, and the latest commit covers that ty. The default single pending run is the better fit, so I’m good with this as implemented.

widal001 (approved):
LGTM! Thanks for making these changes!

SnowboardTechie (approved):

PR #1114, opened 2026-08-21
feat(website): add same-origin mock API playground 

karinamzalez:
### Summary

- Fixes #1078 and #1109
- Time to review: 30 minutes

~~**Scope is frozen and it is ready for review, but it must not merge until the GitHub Pages → Cloudflare migration is complete.**~~ As of 2026-09-01 the migration gate is cleared: [#1145](https://github.com/HHS/simpler-grants-protocol/pull/1145) retired the `deploy-github-pages` 

The findings write-up this experiment produced was removed from the branch head in [f8ae2aa](https://github.com/HHS/simpler-grants-protocol/pull/1114/commits/f8ae2aab1f76dde9dc0f5222e3d94dc5ddb3bd71) (the deliverable of #1078 is the comparison, not merged code) — read it at its last commit: [MOCK-FINDINGS.md @ 642a804](https://github.com/HHS/simpler-grants-protocol/blob/642a80463f7962ccff489aad5bef4f6113325b7a/website/MOCK-FINDINGS.md).

**Live preview:** <https://pr-1114-common-grants.brian-derfer.workers.dev> — the mock, the docs, and Swagger UI's Execute button are all active there.

### Changes proposed

Each bullet links the commit that carries it, so you can review one diff at a time.

- **`website/src/lib/mock/`** — the mock kernel ported from the #1077 Worker: router, opportunity handlers, envelope + CORS helpers, and the 11-record fixture set. `MOCK_API_BASE_PATH` is the single source of the `/api` prefix. Includes **`website/src/pages/api/[...path].ts`**, the site's only dynamic route (`prerender = false`) — a catch-all so the router matches on `request.url` the way the Worker did. Also carries the kernel test suites and the **golden-envelope corpus** that pins byte-identity against the 3A Worker's responses across a 31-case request matrix. → [0239878](https://github.com/HHS/simpler-grants-protocol/pull/1114/commits/02398782802e482ae835913127245f6ba07160f1)
[...]

PR #1137, opened 2026-08-27
feat(website): extend the mock API playground fixtures

### LINKED ISSUE #1101: [Mock API] Grow the hand-written opportunity fixture set
### Summary

The mock API's opportunity records live in `mock-api/src/data/fixtures.ts` — 11
hand-written records built from the TypeSpec `@example` values under
`lib/core/lib/core/models/opportunity/`. We considered switching to a seeded
`@faker-js/faker` generator, but hand-written records are simpler, already
deterministic (it's just a static file), and easier to review. Sticking with that.

This ticket adds more hand-written records so the mock has enough variety for filter,
sort, and pagination behavior to be visible in Swagger UI — and sets the pattern
`#3C-2-T1` will follow for the other resources.

New records should read like real grant data: reuse the TypeSpec `@example` values
where they exist, and keep titles/agencies/amounts plausible. Use the existing `usd()`,
`closeOn()`, `postOn()` helpers so dates and money stay coherent.

This targets `mock-api/` (branch `karina/1077-cloudflareworkermock`), since `#1078-T3`
picked Option 3A. Opportunities only — other resources are `#3C-2-T1`.

### Acceptance criteria

- [ ] New records are hand-written in the same style as the existing 11: fullest
      (v0.3+, detail) shape, semantic values, assembled with the existing helpers
- [ ] Enough records that list pagination, sorting, and the search filters all have
      visible effect with default page sizes
- [ ] These stay exactly as they are — Swagger UI pre-fill and tests depend on them:
  - [ ] `CANONICAL_OPPORTUNITY_ID` (`30a12e5e-…`) is still the spec's published
        example, still first in the default list order
  - [ ] `RESERVED_MISSING_OPPORTUNITY_ID` (`00000000-…`) is still absent from the set
  - [ ] Records 2–4 keep the id + title they share with
        `lib/ts-sdk/examples/mock-api-server.ts`
- [ ] All four status values are still represented, and every `custom` status record
      carries a `customValue`
- [ ] The `#1077-T4` conformance suite passes for every version and variant
      (0.1.0–0.4.0 × list/detail)
- [ ] `mock-api/__tests__/data/fixtures.spec.ts` is updated where it pins the record
      count, and passes
- [ ] `pnpm --filter @common-grants/mock-api run ci` is green

karinamzalez:
### Summary

- Fixes #1101
- Stacked on #1114 — base branch is `karina/1078-astro-endpoint-mock`, so this diff shows only the fixture/endpoint work
- Time to review: 30–45 minutes (roughly half the diff is test suites and hand-written fixture data)

### Changes proposed

**1. Grew the opportunity fixture set from 11 to 25 records** ([`2bd7d75`](https://github.com/HHS/simpler-grants-protocol/commit/2bd7d75))

- With 25 records, paging, sorting, and filtering visibly change what the list returns (at the docs' example page size of 20, there is now a real second page).
- Every existing guarantee is kept: the record the docs use as their example still comes back first, and the id reserved for demonstrating a 404 is the same

**2. The mock API now serves every endpoint the spec documents — not just opportunities** ([`accb24e`](https://github.com/HHS/simpler-grants-protocol/commit/accb24e))

Before this change, clicking "Execute" on any non-opportunity endpoint in the API docs returned a 404. Now all 21 documented endpoints respond with realistic data:

| Resource | Endpoints | Fixture records |
|---|---|---|
| Awards | list, detail, search | 11 |
| Organizations | list, detail, update, submit change, list changes, view change | 8 orgs + 5 changes |
| Applications | start, detail, submit, search, read/write form responses | 7 (each with filled-in form responses) |
| Forms | list, detail | 6 |
| Competitions | detail | 6 |

How it behaves:

- **Old protocol versions.** Awards and organizations only exist from v0.4.0, so asking `/api/v0.2.0/...` for them returns a 404 that says exactly that (see example below). Application search does the same below v0.3.0.

- **Writes check your input and echo the result back, but save nothing.** Submitting an org change returns a 202 with a `Location` header; a PATCH applies a real JSON Merge Patch to a copy of the record and shows you the result. Every call returns the same answer every time, which keeps the playground deterministic.

- **The pieces every endpoint shares-- paging, sorting, filtering, request parsing -- now live in one place** (`src/lib/mock/http/query.ts`) instead of being copied per resource. 

- **New test suites pin all of it:** per-endpoint tests for each resource, a cross-resource suite proving ids that point across resources always point at records that exist, and a schema-conformance suite validating every record against the generated JSON Schemas for every protocol version. 

### Context for reviewers

**How to test:**

- The Cloudflare PR preview builds with the mock enabled -- open `/protocol/api-docs` on the preview URL, pick any endpoint, and click "Try it out" → "Execute". Every id box is pre-filled with a value that resolves.
- Locally: `pnpm --filter website run test` (482 tests) and `pnpm --filter website run checks`.

**Design decisions worth knowing before reading the code:**

1. **Why every resource's first record has the same id** (`30a12e5e-…`): the spec gives all path parameters one shared example value, and Swagger UI pre-fills the id boxes with it. So each resource carries one record with that id-- otherwise the very first "Execute" a visitor runs would 404. A test suite protects this so a future fixture edit can't quietly break it.

2. **Writes don't persist on purpose.** A stateful mock would give different answers depending on call order, and the Cloudflare Worker runtime doesn't guarantee memory survives between requests anyway -- so state would work in tests and silently vanish in production. Trade-off: `POST /applications/start` returns a draft whose id then 404s on GET (mock didn't actually create an application)

3. **Auth is documented but not enforced.** The org endpoints declare OAuth scopes in the spec; the mock ignores tokens so the public playground stays one-click. This means the 401/403 branches can't be demonstrated.

4. **One known rough edge:** `PUT /applications/{appId}/submit` has no request body, so a bare `curl` (no `Origin`, no `Content-Type` header) hits Astro's CSRF guard and gets a plain 403 instead of the mock's response. Adding `-H 'Content-Type: application/json'` clears it. Documented in the handler; every other write sends a JSON body and is unaffected.

5. **Competitions only have a read endpoint** because the spec's API surface (`api.tsp`) only publishes `read` -- the `apply` operation exists as a reusable template but isn't part of the documented API, so the mock deliberately doesn't invent it.

### Additional information

**Screenshots*

every endpoint group live in the API docs at v0.4.0, with the /api/v0.4.0 server wired
image

"Try it out" on GET /awards/{awdId}: the pre-filled id answers 200 with the documented example award
image

**The gaps the new fixture data fills**

Until now the mock's dataset was 25 opportunities and nothing else. This PR fills four kinds of gaps:

1. **Whole-resource gaps** -- awards, organizations, applications, forms, and competitions had no data at all. Each now has a hand-written set, built from the spec's own `@example` values so the docs pane and the live response tell the same story.
2. **"First click" gaps** -- every detail endpoint (including the two-parameter ones like `/orgs/{orgId}/changes/{changeId}`) resolves with its id boxes left untouched.
3. **Model-coverage gaps** -- every field a model declares now appears on at least one record, including the easy-to-miss ones:
   - an award paid to a person rather than an organization
   - form instructions attached as files
   - an organization identifier with an archived historical value
   - a DUNS entry
   - custom fields on a form response
   - Every status enum value (award, application, competition, form-response, revision, opportunity) appears at least once, so filtering by any status returns something. A "representativeness" test suite pins all of this.
4. **Error-path gaps** -- each resource keeps one well-formed id that matches nothing (for demonstrating 404s), one application is deliberately blocked with validation errors (so the submit endpoint's 400 branch is reachable), and version-gated resources explain which version added them:

```console
$ curl https://<preview>/api/v0.2.0/common-grants/awards
{"status":404,"message":"Not found","errors":[{"field":"path",
 "message":"The awards endpoints were added in v0.4.0 and are not served by v0.2.0"}]}
```

```console
$ curl -X POST https://<preview>/api/v0.4.0/common-grants/orgs/30a12e5e-5940-4c08-921c-17a8960fcf4b/changes \
    -H 'Content-Type: application/json' -d '{"mission":"A proposed new mission"}'
HTTP/2 202
location: /api/v0.4.0/common-grants/orgs/30a12e5e-…/changes/4e5f6071-…
{"status":202,"message":"Success","data":{"status":{"value":"pending",…},"patch":{"mission":"A proposed new mission"}}}
```


widal001 (approved):
This looks great! As we discussed yesterday, I'm mostly focusing on testing the behavior of the mock API because we can optimize behind the mock's interface in future PRs.

And the behavior is exactly what I was hoping we could achieve!! I checked the following things:

- Deterministic results from the same endpoint with the same input
- Consistency between data returned by list and get endpoints
- 404 errors when trying to get an ID that doesn't exist
- Correctly handling variation in the response body based on version
- Consistency between the browser and the the same query using `curl`

PR #1149, opened 2026-09-02
feat(website): always serve the mock API, dropping the build-time gate

karinamzalez:
Removes the `MOCK_API_ENABLED` build-time gate so the mock API playground is always on.

## Why it existed

The site used to be hosted on GitHub Pages, which can only serve fixed files. It genuinely couldn't run code to answer `/api/...` requests. So the switch was left off in production to avoid advertising a button that didn't work.

## Why it's gone

The site moved to Cloudflare, which can run code and answer those requests-- i.e. it is now a switch that is never off. So it made sense to me to remove it altogether (dead code).

## What was broken

Only `ci-website-preview.yml` ever set the flag, so previews had the playground and commongrants.org didn't:

| Surface | Production |
|---|---|
| `/api/**` | `404`, empty body |
| Execute button | hidden |
| Injected `servers:` block | absent |

## Changes

- `docs-wiring.ts` — dropped `isGateValueEnabled`, `isMockApiEnabled`, `specDirFor`; last becomes a `SPEC_DIR_URL` constant
- `astro.config.mjs` — dropped the `vite.define` block that inlined the flag
- `api/[...path].ts` — delegates straight to `handleMockRequest`
- `inject-mock-server.ts` — dropped the gate branch and `skipped`
- `api-docs.astro` — dropped `data-try-it-out`; `supportedSubmitMethods` unconditional
- `ci-website-preview.yml` — dropped the env var
- tests — gate tests out; `MockServerInjector`'s no-op test replaced with two asserting it writes a `servers:`-carrying copy of every emitted spec

Net **−148 lines**. `cd-deploy-website.yml` is untouched.

## Second commit: the method allowlist

`supportedSubmitMethods` had drifted — it listed GET and POST, but #1137 taught the router PUT and PATCH. Three operations the mock serves rendered in the docs with Execute silently off:

| Operation | Handler |
|---|---|
| `PUT /applications/{appId}/forms/{formId}` | `writeFormResponse` |
| `PUT /applications/{appId}/submit` | `submitApplication` |
| `PATCH /orgs/{orgId}` | `updateOrganization` |

`cors.ts` was updated for that change; the docs list was the copy nobody touched. The method set lived in three places, with the invariant written down in one (`cors.ts`: *"the method list must track the router's"*) and enforced nowhere.

So: new `lib/mock/http/methods.ts` exports `SUPPORTED_METHODS`, and both consumers derive from it — `cors.ts` builds its `Access-Control-Allow-Methods` header, `api-docs.astro` passes the list to the client via `data-submit-methods`. A leaf module because `router.ts` already imports `cors.ts`; a data attribute because the client script can't import it without dragging the router into the browser bundle.

The test pairs each supported verb with a route that serves it, then asserts every constructible verb outside the constant answers 404 on all of them. Verified it fails in both directions:

- drop PUT/PATCH from the constant while the router still serves them → the 404 sweeps fail (the exact historical drift)
- add DELETE to the constant with no handler → the serves-it check fails

`TRACE`/`CONNECT`/`TRACK` are excluded — the Fetch spec forbids them in the `Request` constructor, so the router can never see one.

## Verification

serving the built Worker via `wrangler dev`:

- all four versions of `/api/v{version}/common-grants/opportunities` → `200` with fixture JSON
- `/openapi-mock/openapi.0.4.0.yaml` → `servers: - url: /api/v0.4.0`
- `/protocol/api-docs/` → `data-spec-dir="/openapi-mock"`
- no gate string left in `dist/`
- all four verbs answer through the Worker: `GET /opportunities`, `POST /opportunities/search`, `PUT /applications/{id}/forms/{formId}`, `PATCH /orgs/{id}` → 200; `PUT /applications/{id}/submit` → 400 validation (route reached); `DELETE` → 404, so the explicit method list still earns its keep

`checks` exit 0 (astro check + cspell clean), 542 tests pass.

## Note

commongrants.org now serves fixture data at `/api/**`.

### DIFF
--- a/website/src/pages/protocol/api-docs.astro
+++ b/website/src/pages/protocol/api-docs.astro
@@ -2,6 +2,7 @@
 import StarlightPage from "@astrojs/starlight/components/StarlightPage.astro";
 import "swagger-ui-dist/swagger-ui.css";
 import { SPEC_DIR_URL } from "@/lib/mock/docs-wiring";
+import { SUPPORTED_METHODS } from "@/lib/mock/http/methods";
 
 // Available OpenAPI versions (mirror of the emitted specs in public/openapi/)
 const availableVersions = [
@@ -26,7 +27,12 @@ const availableVersions = [
     </label>
   </div>
 
-  <div id="swagger-ui" data-spec-dir={SPEC_DIR_URL}></div>
+  <div
+    id="swagger-ui"
+    data-spec-dir={SPEC_DIR_URL}
+    data-submit-methods={SUPPORTED_METHODS.join(",").toLowerCase()}
+  >
+  </div>
 </StarlightPage>
 
 <style>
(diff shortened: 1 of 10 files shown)


website/src/pages/protocol/api-docs.astro
```diff
@@ -96,18 +82,17 @@ const specDir = specDirFor(tryItOutEnabled);
...
+  const specDir = mount?.dataset.specDir ?? "/openapi-mock";
 
   const render = (version: string): void => {
     SwaggerUIBundle({
       url: `${specDir}/openapi.${version}.yaml`,
       dom_id: "#swagger-ui",
-      // Empty disables Execute entirely; the mock serves GET and POST only.
-      supportedSubmitMethods: tryItOut ? ["get", "post"] : [],
+      // The mock serves GET and POST only.
+      supportedSubmitMethods: ["get", "post"],
```
widal001: Out of scope for this PR, but why don't we support mocking patch and put requests?

widal001 (approved):
LGTM! Left one question re: mocking PATCH and PUT methods that would love to capture in a follow-up ticket.

Issue #1192, opened 2026-09-10
[Core] Align opportunity API nullability with ADR 0024


SnowboardTechie:
### Summary

Implement [ADR 0024](https://commongrants.org/governance/adr/0024-optional-field-nullability/) for the opportunity API and its shared request/response schemas. The decision is already accepted; this issue is not reopening whether optional fields should allow null.

During verification of the [Hono template](https://github.com/common-grants/cg-api-template-hono/pull/1), SDK 0.7.2 accepted an opportunity with `funding.details: null` and the API returned it with HTTP 200. Independent validation against the protocol 0.4.0 schema bundled with CLI 0.4.0 rejected that response. `source: null` reproduced the same disagreement. Consumers following the protocol can therefore reject responses produced using our SDK.

[Decision issue #736](https://github.com/HHS/simpler-grants-protocol/issues/736) was closed by the ADR PR, not by an implementation change. This is a bounded opportunity-API implementation slice, not a claim to complete the ADR rollout across every resource.

### Acceptance criteria

- [ ] Apply ADR 0024 to optional fields reachable from the opportunity list, detail and search contracts, including their shared models. Review effects on other consumers of shared definitions; do not make currently required properties optional.
- [ ] Regenerate the affected OpenAPI and JSON Schema artifacts from TypeSpec. Record the intended protocol-version and release impact rather than silently changing historical-version behavior.
- [ ] Add independent regression tests covering omission, explicit null and valid values, including `funding.details` and `source`. Wrong non-null types and missing required fields still fail. Validate serialized wire data against the generated protocol schemas and verify corresponding TypeScript/Python SDK behavior.
- [ ] Document the affected schemas and package-release requirements so downstream consumers can adopt the aligned versions. Do not change SDK semantics merely to make the checker pass.

### Evidence and scope

- [Opportunity funding definition](https://github.com/HHS/simpler-grants-protocol/blob/66476ec54eaf6b81c7af10284597b40d6107bf80/lib/core/lib/core/models/opportunity/funding.tsp) still declares `details?: string`.
- [Shared sorting definitions](https://github.com/HHS/simpler-grants-protocol/blob/66476ec54eaf6b81c7af10284597b40d6107bf80/lib/core/lib/core/sorting.tsp) also distinguish optional request settings from required response metadata; preserve that requiredness distinction.
[...]

Issue #1160, opened 2026-09-02
[TS SDK] Publish the v0.8.0 release


laurabelinfante:
### Summary

Publish the TypeScript SDK v0.8.0 release. Sprint 7.2.

v0.8.0 is deliberately minimal: a parity bump to bring TS (0.7.2) in line with Python (0.8.1). It can't be skipped, but nothing substantive rides on it. The real SDK work is sequenced into the v0.9.0 and v0.10.0 pairs.

**No hold branch** (decided 2026-09-14). Release-please's pending release PR already provides the staging a hold branch would. `cd-release.yml` runs release-please only on `push: branches: [main]` with no `target-branch`, so a hold branch would never reach the release tooling in any case.

### Tickets

- #1193 Catch permissive schema drift and empty parity-test runs — #1195, merging straight to `main`.

### Acceptance criteria

- [x] #1193 is merged to `main` and the squashed commit carries the `Release-As: 0.8.0` footer. The repo squash-merges with `squash_merge_commit_message=PR_BODY`, so the footer has to live in the PR body rather than only in a branch commit. Without it, `bump-minor-pre-major` (set) and `bump-patch-for-minor-pre-major` (unset) make a plain `fix(ts-sdk):` cut 0.7.3 and miss parity with Python 0.8.1.
- [x] release-please opens a release PR for `@common-grants/sdk` at 0.8.0
- [ ] `@common-grants/sdk@0.8.0` is published to npm

SnowboardTechie:
https://github.com/HHS/simpler-grants-protocol/pull/1212 will need to merge ahead of this release to address CI audit failures.

SnowboardTechie:
https://www.npmjs.com/package/@common-grants/sdk/v/0.8.0

Issue #1130, opened 2026-08-26
[TS SDK] Consolidate the duplicate sort and pagination param schemas


laurabelinfante:
### Summary

`SortQueryParamsSchema` and `SortBodyParamsSchema` (`sorting.ts:23` and `:39`) have identical shapes. Same for `PaginatedQueryParamsSchema` and `PaginatedBodyParamsSchema`. The split adds no type safety.

One caveat to resolve first: if we auto-generate schemas from TypeSpec later, a single shared schema may be hard to maintain. Zod does not distinguish query from body params, but TypeSpec does. Consolidating in Zod without a matching approach in the core library could create drift.

### Acceptance criteria

- [ ] The TypeSpec query-vs-body distinction is evaluated, and the approach either consolidates in the core library too or documents why the SDK can diverge
- [ ] Each query/body pair shares one internal schema definition, or a concrete validation difference is implemented and tested to justify separate definitions
- [ ] The existing query and body schema exports remain available and continue to pass parity tests against their corresponding generated TypeSpec schemas
- [ ] Any Core or TypeScript SDK changes resulting from the decision are included in the `[Core] v0.6.0 release` and `[TS SDK] v0.10.0 release` candidates, respectively

### Release coordination

This ticket may affect Core and the TypeScript SDK. GitHub allows only one milestone per issue, so `[Core] v0.6.0 release` is the primary planning milestone and `[TS SDK] v0.10.0 release` is the coordinated secondary release. If the decision does not require Core consolidation or a public SDK behavior change, one or both releases may not be necessary; update the milestone and release implications to match the recorded decision before closing the ticket.

SnowboardTechie:
Worth folding one more pagination-schema fix in here, since it's the same file and the same breaking release.

[...]

PR #1195, opened 2026-09-15
fix(ts-sdk): catch permissive schema drift and empty parity-test runs

### LINKED ISSUE #1193: [TS SDK] Catch permissive schema drift and empty parity-test runs
### Summary

Strengthen the TypeScript SDK's existing schema-parity tests so they detect values the SDK accepts but the protocol rejects, and cannot pass without exercising their promised sample set.

The current harness already uses an independent TypeSpec-generated schema and AJV. However, it generates examples from the protocol schema, then checks them with Zod. That does not systematically exercise values excluded by the protocol but accepted by a more permissive SDK, such as `paginationInfo.pageSize: 0` or nullable opportunity fields. Those disagreements surfaced during [Hono template verification](https://github.com/common-grants/cg-api-template-hono/pull/1) despite passing tests.

The harness also skips generation failures and sets `passed` from the absence of mismatches. Source inspection shows that it can report success when no samples were generated. This is a reachable harness defect, not a claim that an observed CI run executed zero samples.

### Acceptance criteria

- [x] Retain the independent TypeSpec/AJV reference and protocol-valid sampling. Add deterministic adversarial or boundary cases that exercise the reverse direction: SDK-accepted values outside the protocol's accepted domain. Initially cover opportunity fields and pagination results.
- [x] Add harness regression tests using deliberately mismatched schema pairs. They must detect both an overly permissive SDK schema and an overly restrictive one, including nullability, numeric boundaries and requiredness, without depending on a live production mismatch staying unresolved.
- [x] Report attempted, generated and validated sample counts accurately. Fail when the required sample budget is not met, including all-generation-failed and partial-generation-failed cases; test those failure paths explicitly.
- [x] Ensure missing or invalid reference schemas and unsupported validation setup fail visibly. Compare the appropriate JSON wire representation with format validation, rather than confusing runtime `Date` values with serialized strings.
- [x] Run the added checks through the existing required SDK test path. Link real model disagreements to their owning alignment issues instead of silently skipping them or changing production schemas in this harness issue. Known unresolved cases must be visibly tracked; do not claim full parity while they remain unresolved.

### Evidence and scope

- [Parity harness](https://github.com/HHS/simpler-grants-protocol/blob/66476ec54eaf6b81c7af10284597b40d6107bf80/lib/ts-sdk/__tests__/utils/fuzz-test.ts#L155-L201): protocol-derived samples, skipped generation failures and success calculation.
- [AJV reference loader](https://github.com/HHS/simpler-grants-protocol/blob/66476ec54eaf6b81c7af10284597b40d6107bf80/lib/ts-sdk/__tests__/utils/ajv-validator.ts): independent generated schemas and format validation already exist; extend this path rather than create a second framework.
- [Pagination parity test](https://github.com/HHS/simpler-grants-protocol/blob/66476ec54eaf6b81c7af10284597b40d6107bf80/lib/ts-sdk/__tests__/schemas/zod/pagination.spec.ts#L154-L171): a parity assertion exists, but its sample source excludes the zero boundary accepted by the SDK.
- Production nullability changes belong to the [companion alignment issue](https://github.com/HHS/simpler-grants-protocol/issues/1192); the TypeScript SDK's `pageSize >= 0` response allowance is folded into [#1130](https://github.com/HHS/simpler-grants-protocol/issues/1130). Harness work can begin independently; use their settled contracts for real-model regression expectations.
- Hono HTTP-response conformance tests remain in [#1152](https://github.com/HHS/simpler-grants-protocol/issues/1152). This issue does not implement CLI checking ([#735](https://github.com/HHS/simpler-grants-protocol/issues/735), [#524](https://github.com/HHS/simpler-grants-protocol/issues/524)), introduce schema code generation ([#873](https://github.com/HHS/simpler-grants-protocol/issues/873)), or deliver the public v1.0 conformance suite ([#581](https://github.com/HHS/simpler-grants-protocol/issues/581)).

SnowboardTechie:
### Summary

- Fixes #1193
- Time to review: 20 minutes

This is the only change in `[TS SDK] v0.8.0`, so it merges straight to `main` with no hold branch and carries the version bump itself. #1160 stays open until the release publishes.

### Changes proposed

Test-only — seven files under `lib/ts-sdk/__tests__/`, and no production schema changes. The `README.md` change is the dependency-diagram refresh the mapping workflow pushed onto this branch; it is the same update it makes on `main`.

- **Honest sample accounting.** Attempted / generated / validated counts are reported separately and any generation failure fails the run. Previously a failed generation was `continue`d and `passed` came from `mismatches.length === 0`, so a run that generated nothing reported success.
- **No vacuous passes.** A generated sample that its own source schema rejects now throws. Both validators would reject it, which reads as agreement and scores as a success — the same shape as the bug above, one level down.
- **The reverse direction.** `expectZodMatchesJsonSchema` now takes deterministic boundary cases. Generated samples are protocol-valid by construction, so they can only ever catch a Zod schema *stricter* than the protocol — never one accepting a superset. Wired up on `PaginatedResultsInfo` and `OpportunityBase`.
- **Tracked divergences.** A boundary case can be marked `expect: "divergent"` naming the issue that owns the fix. The harness fails once such a case stops diverging, so an entry cannot outlive what it tracks; one with no issue is rejected outright.
- **Wire-form comparison.** AJV validates the serialized form, so a `Date` the SDK accepts is not read as a mismatch against `format: date-time`.
- **Visible setup failures.** Construction asserts format validation actually bites — drop `ajv-formats` and every `uuid` / `uri` / `date-time` constraint silently passes while the harness still reports parity. A missing reference bundle now names the file and `pnpm build`.

### Context for reviewers

**What it surfaces.** Nine real disagreements, all pre-existing and all left to their owning issues:

| Schema | Values the SDK accepts and the protocol rejects | Owner |
|---|---|---|
| `PaginatedResultsInfo` | `pageSize: 0` | #1130 |
| `PaginatedResultsInfo` | `totalItems: null`, `totalPages: null` | #1192 |
| `PaginatedResultsInfo` | `page: 2147483648` | #1196 |
| `OpportunityBase` | `funding`, `keyDates`, `source`, `customFields`, `acceptedApplicantTypes` each `null` | #1192 |

The `page` row is a Core inconsistency rather than an SDK bug: `page` is the only `int32` in `pagination.tsp` (its siblings are `integer`), and the SDK's `.int()` is already the safe-integer encoding of `integer`. #1196 widens Core; that entry then fails-on-fix and gets removed there.

**Verification.** `check:lint`, `check:types`, `build`, `test:coverage` (27 files, 573 tests, up from 550; 98% statements against an 80% threshold) and `example:transforms` all pass.

Three claims were mutation-tested rather than assumed:

- Tightening `PaginatedResultsInfoSchema.pageSize` to `min(1)` makes `pagination.spec.ts` fail demanding the now-stale divergence entry be removed.
- The boundary cases were first run unmarked, confirming the harness reports all nine divergences as hard failures, before being marked with their owners.
- Inverting the generated-sample guard fails 72 tests across the suite, confirming it is armed at every parity call site rather than only in the harness's own tests.

The harness regression tests run against deliberately mismatched **synthetic** schema pairs rather than production ones, so they do not depend on any live disagreement staying unresolved.

Release-As: 0.8.0


lib/ts-sdk/__tests__/utils/fuzz-test.ts
```diff
@@ -144,60 +284,138 @@ export async function checkZodMatchesJsonSchema(
...
+      } else {
+        successCount++;
+      }
+      continue;
+    }
+
+    const tracked = { label: boundaryCase.label, issue: boundaryCase.issue as string };
+    if (disagreement) {
+      knownDivergences.push(tracked);
       successCount++;
```
karinamzalez: Nit: this counts a tolerated divergence as a success, but successCount's doc says "agreed as expected". Worth a separate divergentCount so each number means one thing?

lib/ts-sdk/__tests__/utils/ajv-validator.ts
```diff
@@ -12,24 +12,62 @@ import * as yaml from "js-yaml";
...
+ * that the SDK matches the protocol. Nothing else in the suite would notice, so
+ * the setup asserts it up front.
+ *
+ * @param ajv - The instance to check
+ * @param source - Where its schemas came from, used in the error message
+ */
+export function assertFormatValidationActive(ajv: Ajv2020, source: string): void {
+  // No $id, so nothing is registered and there is nothing to clean up. A throw
+  // from compile() is already loud, so it is left to propagate as-is.
+  const probe = ajv.compile({ type: "string", format: "uuid" });
```
karinamzalez: Nit: the comment lists three formats at risk but only uuid is probed. Worth adding uri and date-time so the check covers what it says it does?

karinamzalez (approved):
LGTM! A few questions: 
- Is there a follow-up to add boundary cases beyond `PaginatedResultsInfo` and `OpportunityBase`? Issue 1193 scoped this PR to those two, but the other parity specs still have the one-direction blind spot.
- Should the helper enforce that a caller with tracked divergences asserts on them? Today a future spec could mark a case divergent and never assert, and it reads fully green. An `expectedDivergences` count on the helper would make the omission a failure.
- Why 0.8.0 instead of 0.7.3? A fix yields a patch under our release-please config, and the published code is unchanged. Apologies if this was already discussed in a call at some point and I'm forgetting!

lib/ts-sdk/__tests__/utils/fuzz-test.ts
SnowboardTechie: Yeah, that increment was wrong. Dropped it, so `successCount` is agreement only now. `knownDivergences.length` already carries the tolerated count, so I didn't add a second number for it.

lib/ts-sdk/__tests__/utils/ajv-validator.ts
SnowboardTechie: The probe covers uuid, uri and date-time now and names whichever one let a bad value through. There's a test registering only `uuid` via `addFormats(ajv, ["uuid"])`, which is the case the old probe waved through.

SnowboardTechie:
Branch is back at what you approved plus one commit for the two inline nits and the declared count. On the three:

**Follow-up for the other specs.** There is now: #1218. #1193 scoped this PR to opportunity fields and pagination, so fields, types, filters, sorting, models and responses get their reverse-direction cases there.

**Declared divergence count.** I think so, that closes a real gap. The helper takes `expectedDivergences` (default 0) and fails when the tolerated count differs, so a spec can't mark a case divergent without saying so at the call site. `PaginatedResultsInfo` declares 4 and `OpportunityBase` declares 5.

**0.8.0.** It's the version-parity bump from the SDK release sequencing: TS is on 0.7.2 and Python on 0.8.1, and from here the SDK minors move as pairs (0.9.0, 0.10.0), so TS needs to land on 0.8.0 first. Nothing substantive rides on it, which is why it's this one test-only change. `Release-As: 0.8.0` in the PR body is what gets release-please to cut a minor instead of the 0.7.3 a `fix` would produce. Written up in #1160.

